PR #1, opened 2023-03-12
DRAFT: Add run command and VM layer

gerred:


jeff-mccoy:
lol *fail* I updated v8 -> vm to see your changes cleaner in this PR and created merge conflicts....sad times

gerred:
@jeff-mccoy I'll fix, unless you're doing also doing something - feel free to push to this branch. I prefer vm just because it's more indicative of how we're using it with v8 as an impl detail

jeff-mccoy:
yeah I agree vm is better, I'll let you fix it, so I don't bork the direction you were going, may need to leave the tsconfig stuff as it was though for the stuff I was working on

gerred:
I'll uncomment that part, fixture won't work so we'll need to update that but not sure how - we can revisit that on Tuesday so we have a `pepr run` example that works.

jeff-mccoy:
also agree we can get fixtures cleaned up to show an actual e2e example

jeff-mccoy:
sounds like a plan

gerred:
Some context:

The v8go context and isolate were separated because we're defining some core objects that rely on being defined against the isolate (and then re-defined against the context later). This includes stuff like console, fetch, etc. While it's inline now, this needs to get moved into it's own package once we start working on fetch, and this package will get passed the isolate into it rather than the context because that's the level where that gets defined (before then we re-define it in the global scope of each context).

So this PR just proves out the method and then I'll move it into a new package in another PR.

gerred:
Also, note on the GlobalName change - since it's compiling to IIFE based on the settings currently, GlobalName allows default exports to be available to future scripts that are being run. We'll just need to be aware of this in future changes if we ever change from iife to adjust the way the user script actually gets called correctly. We can change that global as well - it doesn't matter and doesn't actually refer to global in the `window` or `document` sense if that makes sense, it's just a well known name between what variable the iife puts into global scope and what we call when running the transform function against the payload.

pkg/vm/common.go
```diff
@@ -17,19 +19,45 @@ func New(b *builder.Build) *Script {
 	}
 }
 
-func (s *Script) Run() (*v8go.Value, func(), error) {
-	ctx := v8go.NewContext()
+func (s *Script) Run(payload []byte) (*v8.Value, func(), error) {
+	iso := v8.NewIsolate()
```
jeff-mccoy, line 23: this is the format I originally had, but it was unclear to me if iso.Dispose would also release the ctx so I combined them, can you clarify how that works?

pkg/vm/common.go
```diff
@@ -1,12 +1,14 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: 2023-Present The Pepr Authors
 
-// vm package contains the v8go wrapper for running JavaScript code in Pepr.
```
jeff-mccoy, line 4: We should preserve these as lint will throw an error when I add it later this week

pkg/vm/common.go
```diff
@@ -17,19 +19,45 @@ func New(b *builder.Build) *Script {
...
-	out, err := script.Run(ctx)
+	_, err = script.Run(ctx)
+	if err != nil {
+		return nil, iso.Dispose, err
+	}
+
+	out, err := ctx.RunScript("global.transform(__pepr_payload)", "main.js")
 	if err != nil {
-		return nil, ctx.Close, err
+		fmt.Println(err)
```
jeff-mccoy, line 60: assume this was for debugging purposes, but we should return the error/close handler vs just log it

pkg/vm/common.go
gerred: actually not sure how that got removed, will re-add.

pkg/vm/common.go
gerred: Isolate is a parent of the context - so when we're defining globals it's at the isolate level, then applied to each context individually. Idea is you can re-use the isolate w/ different contexts.

Since we're now adding in externals like console/fetch/etc., that's where the separation in the API is important.

pkg/vm/common.go
gerred: oh yes, and also this should do the iso.Dispose as well.

gerred: if we get to reusable isolates per-script (1:N script to instantiations on isolate vs. context) top level failures should be iso.Dispose vs. ctx.Close

pkg/vm/common.go
gerred: Opposing that, we'd have to check to be sure, but ctx.Close when separated COULD cause a memory leak on the isolate.

jeff-mccoy:
yeah I left it on IIFE, but for no particular reason other than I figured we'd tweak it to be an optimal comm later on. Also super important--we shouldn't expose those details at all to users; we should handle all the comms between the js / go for them.

jeff-mccoy:
closing this since we've pivoted to node for now

PR #28, opened 2023-04-14
added example of configuring keycloak from scratch, WIP

bdw617:
Related to #42


examples/keycloak-pepr/capabilities/util.ts
```diff
@@ -0,0 +1,17 @@
+import fetch from 'node-fetch';
```
jeff-mccoy: Probably better to use the crypto library for this, something like

```ts
import * as crypto from "crypto";

/**
 * Generate a random password of the given length
 *
 * @param length
 * @returns
 */
export function generatePassword(length = 16) {
  const chars =
    "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";
  const buf = crypto.randomBytes(length);
  const password = Array.from(buf, byte => chars[byte % chars.length]).join("");

  return password;
}
```

examples/keycloak-pepr/capabilities/keycloak-api.ts
```diff
@@ -0,0 +1,264 @@
...
+  getHeaders() {
+    return {
+      'Content-Type': 'application/json',
+      Authorization: `Bearer ${this.accessToken}`,
+    }
+  }
+
+  // 
+  async createUser(realmName: string, username: string, email: string, firstName: string, lastName: string): Promise<string> {
+    await this.connect()
```
jeff-mccoy, line 65: calling `this` in the manner is considered dangerous. In almost all cases the use of `this` only makes sense when used with a class and typically an arrow function to guarantee `this` context safety (which can move in javascript)

examples/keycloak-pepr/capabilities/keycloak-api.ts
```diff
@@ -0,0 +1,264 @@
...
+      const createUserUrl = `${this.baseURL}/admin/realms/${realmName}/users`
+
+      const requestData = {
+        username: username,
+        email: email,
+        firstName: firstName,
+        lastName: lastName,
+        enabled: true,
+      }
+      var { data, ok, status, statusText } = await fetch(
```
jeff-mccoy: `var` should be avoided in modern js/ts--should be `let` or `const`

examples/keycloak-pepr/capabilities/keycloak-api.ts
```diff
@@ -0,0 +1,264 @@
...
+    // likely we want to only do this for a new user....
+    const password = generateRandomPassword(12);
+
+    const setPasswordUrl = `${this.baseURL}/admin/realms/${realmName}/users/${userId}/reset-password`
+
+    var { data, ok, status, statusText } = await fetch(
+      setPasswordUrl,
+      {
+        method: 'PUT',
+        headers: this.getHeaders(),
```
jeff-mccoy, line 110: not really user what `this.getHeaders()` is supposed to do here

examples/keycloak-pepr/capabilities/keycloak-api.ts
```diff
@@ -0,0 +1,264 @@
...
+
+  private async getUser(realmName: string, username: string): Promise<string | undefined> {
+    interface usersResponse {
+      id: string
+    }
+    const getUserUrl = `${this.baseURL}/admin/realms/${realmName}/users?username=${encodeURIComponent(username)}`;
+    const { data, ok, status, statusText } = await fetch<usersResponse[]>(
+      getUserUrl,
+      {
+        method: 'GET',
```
jeff-mccoy: the default method is `GET`, this field isn't required

examples/keycloak-pepr/capabilities/keycloak-api.ts
```diff
@@ -0,0 +1,264 @@
...
+  async getRealm(realmName: string): Promise<boolean> {
+    const { data, ok, status, statusText } = await fetch(
+      `${this.baseURL}/admin/realms/${realmName}`, {
+      method: 'GET',
+      headers: this.getHeaders(),
+    });
+    if (ok) {
+      return true
+    }
+    if (status !== 404) {
```
jeff-mccoy: would probably be better to use `fetchStatue.NOT_FOUND`

examples/keycloak-pepr/capabilities/keycloak-api.ts
```diff
@@ -0,0 +1,264 @@
...
+      throw new Error(`Error checking for realm existence: ${statusText}`);
+    }
+    return false
+  }
+
+
+  async createOrGetKeycloakRealm(realmName: string) {
+    await this.connect()
```
jeff-mccoy, line 172: not sure what `this.connect()` is doing

jeff-mccoy:
Closing, moved to the new repo: https://github.com/defenseunicorns/pepr-capabilities/pull/8

PR #81, opened 2023-05-02
Ask Sage automated code improvements

nicolaschaillan:


bdw617 (dismissed):
looks good, lots of formatting changes. woah.

nicolaschaillan:
Yea don't approve it though, this was a test with Jeff McCoy of Ask Sage automated code review. Lots of issues so we're improving the prompt with Jeff & working this! ;-)

The beauty here is that Ask Sage can scan the Git repo every day or week and propose changes/improvements/security review everytime!

jeff-mccoy:
we'll keep playing with it, but going to close this PR for now as there are too many things not ready to bring in:

## Where it is good:
- spelling errors and general code comments
- JSDoc addition / correction

## Things it breaks:
- does not respect lint/formatting rules setup in the repo
- breaks openssf scorecard with changes from hash -> tags
- all workflow changes are invalid
- doesn't understand the template portion of pepr that has to be consumed during `pepr init`
- doesn't understand the implicit fail state (no try-catch) in capabilities and how that is managed by Pepr 
- doesn't understand the concept of exports intended to be used by another package (in the `upstream.ts` file)
- doesn't understand how IDEs interpret complex types + JSDoc (they require the JSDoc on the type def vs the actual implementation) 
- doesn't understand the complex generic path-through type pattern with a fluent interface (just deletes all the types)

## Non-breaking, but subjective:
- several areas are arbitrary code rearrangement with no obvious reason why
- strong preference for `Promise.all` vs `async/await`

PR #84, opened 2023-05-03
CLI/SDK packaging optimizations + enhance SDK transparency

jeff-mccoy:
Several changes to improve the number of dependencies in each use case (CLI/SDK/K8s Controller) as well as provide real Typescript again for better DevX:

- Resolve issue with combinations of ESModule and CommonJS imports when creating capabilities
- Leverage `peerDependencies` to separately track dependencies to exclude from the controller, reducing image size and size for SBOM
- Readme fixes + use SVG vs mermaid for module diagram (several NPM sites don't support mermaid)
- Run unit tests against transpiled code vs pure typescript (closer to the real behavior)
- Stop emitting typescript defs and instead include actual typescript for use in the SDK
- Remove TSNode with it's bad opinions on typescript in NPM packages
- Add `strict:true` for tsconfig in the core project for tighter control of the codebase
- Remove `commander` use by the controller `run.js` entrypoint
- Gracefully handle controller termination
- Fix hotreload behavior when using `pepr dev`
- Switch to `fork` vs `spawn` and `require` where able
- Only watch `pepr.ts` and `capabilities/` for typescript file changes on `pepr dev`


src/cli/update.ts
```diff
@@ -52,7 +52,9 @@ export default function (program: RootCmd) {
         console.log(`Module updated!`);
       } catch (e) {
         Log.debug(e);
-        Log.error(e.message);
+        if (e instanceof Error) {
+          Log.error(e.message);
+        }
```
bdw617, line 57: why is this necessary? can catch ever have something not an Error? it seems like just too much.

bdw617 (approved):
one comment but overall looks fine fine, just fine.

src/cli/update.ts
jeff-mccoy: Yeah it's a result of adding `strict` to tsconfig, just one of those fun little js things.... * sigh *

PR #94, opened 2023-05-09
Fix `pepr dev` regression

jeff-mccoy:
Re-adds `ts-node` as a peer dependency due to rollupjs source maps not working properly during `pepr dev` debugging in Typescript files (only worked properly in the transpiled Javascript).


jeff-mccoy:
Screenshot 2023-05-09 at 10 22 11 AM

jeff-mccoy:
Have to manually test as we don't have a good way today to do a GUI-based E2E of this experience using VSCode in CI

bdw617 (approved):

PR #121, opened 2023-05-31
Breaking Change: Automatic base64 encoding/decoding for secrets

### LINKED ISSUE #95: Add better handling/wrapping for Secrets
Currently the Secret `data` property is carried as Base64 encoded (unlike the GO SDK that does this for you), Pepr should have a wrapper for these functions, also should include an update to `HelloPepr` showing this:

Read from:
```ts
 if (data && data[p]) {
    return Buffer.from(data[p], "base64").toString("utf-8");
  }
```

Write back:
`Buffer.from(secret).toString("base64");`

jeff-mccoy:
This PR closes #95 by transparently handling Base64 encoding for the `data` field of Secrets. This behavior mirrors the Go K8s client behavior and eliminates the need for users to understand/worry about dealing with the encoding. 

Other options explored included extending new classes for ConfigMaps and Secrets, and using getters/setters, but these failed due to the nature of JS getter/setter support for modifying object properties.

### DIFF
--- a/hack/e2e.test.mjs
+++ b/hack/e2e.test.mjs
@@ -111,8 +111,9 @@ test.serial("E2E: `pepr deploy`", async t => {
     const cm4 = await waitForConfigMap("pepr-demo", "example-4");
     const cm4a = await waitForConfigMap("pepr-demo-2", "example-4a");
     const cm5 = await waitForConfigMap("pepr-demo", "example-5");
+    const s1 = await waitForSecret("pepr-demo", "secret-1");
 
-    t.log("ConfigMaps created");
+    t.log("ConfigMaps and secret created");
 
     // Validate the example-1 CM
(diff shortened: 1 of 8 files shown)


hack/e2e.test.mjs
```diff
@@ -247,6 +254,16 @@ async function waitForConfigMap(namespace, name) {
   }
 }
 
+async function waitForSecret(namespace, name) {
```
bdw617, line 257: this looks like could be an endless loop?

bdw617 (commented):
see comment

hack/e2e.test.mjs
jeff-mccoy: Yes it is, but the ava test config holds the timeout and will kill it.

bdw617: ok

bdw617 (approved):
ok, if the ava stuff will kill it, prob good to document that though

Issue #31, opened 2023-04-17
Make ClusterRole, Webhook bindings restricted to current module


jeff-mccoy:
Currently, Pepr doesn't limit the ClusterRole or Webhook bindings to the current module. These should be limited if possible to not over-privilege the controller service account. We also need to evaluate how the webhook bindings work.

<img width="461" alt="Screenshot 2023-04-17 at 1 44 50 AM" src="https://user-images.githubusercontent.com/882485/232405228-0fbbd835-b947-4dcf-8081-f598dd3b9563.png">
<img width="1136" alt="Screenshot 2023-04-17 at 1 44 37 AM" src="https://user-images.githubusercontent.com/882485/232405229-3db46fe3-df0a-4df2-81fe-1ef803e9a951.png">

PR #133, opened 2023-06-05
Generate Dynamic MutatingWebhookConfiguration Rules

jeff-mccoy:
Related to #31


bdw617 (dismissed):
ok, but we need to do more downstream testing too.

bdw617 (approved):

PR #159, opened 2023-06-16
experiment to start adding tracing

bdw617:
you'll need to deploy k3d, and tempo. run a pepr capability in the debugger (with this version) and have the port-forward:

to deploy tempo:
1. https://github.com/grafana/helm-charts/tree/main/charts/tempo#chart-repo
2. once it's deployed you'll need to forward port 4318 so that a pepr capability in the debugger can contact it
3. kubectl port-forward svc/tempo 4318:4318

then you'll need to have a capability running, best to use one that makes external async calls like in pepr-istio: https://github.com/defenseunicorns/pepr-istio/tree/dd-play-3 it needs the CRDs for istio deployed (but not the rest of it).


jeff-mccoy:
going to close this PR now as stale, we can revisit in the future if we have a user demand for tracing.

PR #170, opened 2023-06-26
Added version to the zarf.yaml

### LINKED ISSUE #169: Add version to zarf.yaml generated by pepr build
Given package.json has a version key set, I would like that version to be added to the zarf.yaml that `pepr build` generates.

Current behavior: 
Version in package.json is ignored

Desired behavior:
Version is added as a metadata key in the generated zarf.yaml

Michael-Kruggel:
closes #169

### DIFF
--- a/src/cli/build.ts
+++ b/src/cli/build.ts
@@ -37,6 +37,7 @@ export default function (program: RootCmd) {
       // Generate a secret for the module
       const webhook = new Webhook({
         ...cfg.pepr,
+        version: cfg.version,
         description: cfg.description,
       });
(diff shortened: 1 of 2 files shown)


bdw617 (dismissed):
looks ok to me

jeff-mccoy (commented):
I don't think we have any tests yet for the zarf.yaml itself, so can you add a screenshot or output `zarf.yaml` to this pr comment for our records validating the fix.

Michael-Kruggel:
Here is the screenshot validating the fix:
image

jeff-mccoy (approved):

PR #172, opened 2023-06-27
metrics endpoint in pepr

bdw617:
added a simple metrics route to pepr, not exposed via servicemonitor a WIP for now.

for now, I've only added 3 counters that live in the /mutate endpoint:
alert: if you use the wrong `API KEY`
summary: it uses a full summary for OK requests. 
error: a counter if there are issues. 

I've not yet figured out the best way to allow extended metrics into each of the capabilities, but for now, the simplicity of this should help us drive that.

https://github.com/defenseunicorns/pepr/issues/24

### DIFF
--- a/src/lib/controller.ts
+++ b/src/lib/controller.ts
@@ -10,7 +10,6 @@ import { Request, Response } from "./k8s/types";
 import Log from "./logger";
 import { processor } from "./processor";
 import { ModuleConfig } from "./types";
-import { performance } from "perf_hooks";
 import { MetricsCollector } from "./metrics";
 
 export class Controller {
@@ -137,7 +136,7 @@ export class Controller {
   };
 
   private mutate = async (req: express.Request, res: express.Response) => {
-    const startTime = performance.now();
+    const startTime = this.metricsCollector.observeStart();
 
     try {
       // Validate the token
@@ -177,7 +176,7 @@ export class Controller {
         kind: "AdmissionReview",
         response,
       });
-      this.metricsCollector.observe(startTime);
+      this.metricsCollector.observeEnd(startTime);
     } catch (err) {
       console.error(err);
       res.status(500).send("Internal Server Error");
--- a/src/lib/controller.ts
+++ b/src/lib/controller.ts
@@ -10,10 +10,13 @@ import { Request, Response } from "./k8s/types";
 import Log from "./logger";
 import { processor } from "./processor";
 import { ModuleConfig } from "./types";
+import { performance } from "perf_hooks";
+import { MetricsCollector } from "./metrics";
 
 export class Controller {
   private readonly app = express();
   private running = false;
+  private metricsCollector = new MetricsCollector("pepr");
 
   // The token used to authenticate requests
   private token = "";
@@ -32,6 +35,9 @@ export class Controller {
 
     // Health check endpoint
     this.app.get("/healthz", this.healthz);
+
+    // Metrics endpoint
+    this.app.get("/metrics", this.metrics);
 
     // Mutate endpoint
     this.app.post("/mutate/:token", this.mutate);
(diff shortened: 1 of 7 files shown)


src/lib/controller.ts
```diff
@@ -121,14 +127,26 @@ export class Controller {
...
+    try {
+      res.send(await this.metricsCollector.getMetrics());
+    } catch (err) {
+      console.error(err);
+      res.status(500).send("Internal Server Error");
+    }
+  };
+
   private mutate = async (req: express.Request, res: express.Response) => {
+    const startTime = performance.now();
```
jeff-mccoy: would probably be better to just use the request time tracking in the `logger` function

src/lib/metrics.ts
```diff
@@ -0,0 +1,50 @@
...
+    });
+
+    this.summary = new promClient.Summary({
+      name: `${prefix}_summary`,
+      help: "summary",
+      registers: [this.registery],
+    });
+  }
+
+  error(): void {
```
jeff-mccoy: We should at least have jsdoc for public functions (`/** */` format).

src/lib/metrics.ts
```diff
@@ -0,0 +1,50 @@
...
+    this.errors.inc();
+  }
+
+  alert(): void {
+    this.alerts.inc();
+  }
+
+  observe(startTime: number): void {
```
jeff-mccoy: it feels odd to me to have a time entry pulled by the callee to start and then also in the called method to end--seems like it should be generated and tracked internally, if nothing else:

E.g. 

```typescript
observe() {
  this._timer = performance.now();
}

observeEnd() {
this.summary.observe(performance.now() - this._timer);
}
```

src/lib/metrics.ts
```diff
@@ -0,0 +1,50 @@
...
+    this.summary.observe(performance.now() - startTime);
+  }
+
+  async getMetrics(): Promise<string> {
```
jeff-mccoy: style-wise, we don't typically include the return info unless it is needed; it just makes the code more verbose and doesn't help intellisense for the most part.

src/lib/metrics.ts
```diff
@@ -0,0 +1,50 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2023-Present The Pepr Authors
+
+import promClient from "prom-client";
+import { performance } from "perf_hooks";
+
+export class MetricsCollector {
+  private registery: promClient.Registry;
```
jeff-mccoy: spelling `registery`

src/lib/metrics.ts
```diff
@@ -0,0 +1,50 @@
...
+  private errors: promClient.Counter<string>;
```
jeff-mccoy: ironically, everywhere but the `controller.ts`, we are using the same convention of prefixing the private fields with an `_`,:

`private _errors: promClient.Counter<string>;`

jeff-mccoy (changes requested):
This will be a very useful addition, good work. More code comments would be nice--especially jsdoc for public methods, we need to add some test to the e2e suite to verify the metrics are working. Also, could you drop a sample metrics output in the PR comments for reference, assuming it will output [Prometheus format](https://prometheus.io/docs/instrumenting/exposition_formats/)?

src/lib/metrics.ts
bdw617: I'd have to create a new instance of the class for that...

How's this instead?
  observeStart(): number {
    return performance.now()
  }

  observeEnd(startTime: number): void {
    this.summary.observe(performance.now() - startTime);
  }

src/lib/metrics.ts
bdw617: :| 

so like this instead?

  async getMetrics() {
    return this.registry.metrics();
  }

jeff-mccoy: yes plz!

src/lib/metrics.ts
jeff-mccoy: yeah that's prob fine, thx

jeff-mccoy:
@bdw617 still need a test for this to get it merged.

jeff-mccoy (approved):
lgtm, thanks!

PR #179, opened 2023-07-06
Fix delete metadata filters

jeff-mccoy:


bdw617 (approved):
ready to test it when ready :)

PR #180, opened 2023-07-07
PeprStore, K8s Fluent API & Watch Action

### LINKED ISSUE #49: Update K8s client to 1.x
Waiting for the next RC, but we need to upgrade to the 1.x version of the K8s client (currently under RC only) so that we can remove some findings for deprecated packages.

Related to https://github.com/kubernetes-client/javascript/issues/754

Screenshot 2023-04-24 at 12 11 51 PM

### LINKED ISSUE #139: Add more example capabilities that include the creation of resources
As a user of this repo I found it difficult to understand how I would implement more complex capabilities like the creation of resources. I believe if there were more examples of capabilities that included a wider scope of options it would be incredibly helpful as I work with Pepr.

### LINKED ISSUE #156: Complete Implementation of alwaysIgnore for for Labels in package.json
Complete implementation of `alwaysIgnore` in the `zarf` section of generated `package.json`.

_Reasoning_
**Given** you are using Pepr to create a module to replace the Zarf Agent,
**When** a request comes to the admission controller with labels `zarf.dev/agent: ignore` or `zarf-agent: patched`,
**Then** the hook should ignore the request.

Links to Zarf code:
- [Namespace ignore](https://github.com/defenseunicorns/zarf/blob/1aeae1b584ec36cd3ce252358c8ccbeabbce0dfe/src/internal/cluster/state.go#L94)
- [Pod Ignore](https://github.com/defenseunicorns/zarf/blob/1aeae1b584ec36cd3ce252358c8ccbeabbce0dfe/src/internal/cluster/injector.go#L290)
- [Patched example](https://github.com/defenseunicorns/zarf/blob/1aeae1b584ec36cd3ce252358c8ccbeabbce0dfe/src/internal/agent/hooks/pods.go#L50)

```json
  "pepr": {
    "name": "zarf-agent",
    "uuid": "cb9dbea4-d6c1-5b60-b4f1-758ed8843298",
    "onError": "audit",
    "alwaysIgnore": {
      "namespaces": [],
      "labels": ["zarf.dev/agent: ignore"]
    }
  },
```

### LINKED ISSUE #173: Pepr K8s-native state system
As we've started to develop use cases for Pepr, it is becoming apparent that there are times when you need to share data between different Capabilities and CapabilityActions. After discussing several ways we might do this, @bburky @cmwylie19 @bdw617 @rjferguson21 and I landed on the general idea for persisting shared state within K8s and maintaining an automated watch system for changes.  More thoughts and details are below:

Problem: we need to share data, and using K8s resource metadata exclusively is cumbersome and brittle. One option is to generate more K8s resources, which is more complex and slower and requires resource cleanup.

Idea: define a simple key-value store native to Pepr core and transparently available for users of the SDK.  More details:

- Leverage a K8s CRD/CR (`PeprStore`) for pepr to maintain key-value state data per-capability
- Provide an API that generally matches the [Web Storage API](https://developer.mozilla.org/en-US/docs/Web/API/Storage)
- A watcher for all `PeprStore` resources (watchers are batch operations) that updates the internal state of the object for read
- A `Store` method available from `Capabilities` to bind data
- This can be exported for sharing data between `Capabilities`

```typescript
// Use the 'When' function to create a new Capability Action, 'Store' to persist data
const { When, Store } = HelloPepr;

// Save some data to the store
Store.setItem("hello-pepr", "Hello Pepr!");

// Add an item to the store
Store.setItem("hello-pepr-2", "Hello Pepr! I'm a second item!");

// Read the data back from the store
console.log(Store.getItem("hello-pepr"));

// Update the data in the store
Store.setItem("hello-pepr", "Hello Pepr! I've been updated!");

// Get the length of the store
console.log(Store.length);

// Delete an item from the store
Store.removeItem("hello-pepr");

// Clear the store
Store.clear();
```

- Discuss write operations: batching, collisions, performance, failures, etc
- Discuss read/write contracts: i.e., don't return write until successful persistence, require force K8s resource get before returning a read

jeff-mccoy:
Fixes #173 
Fixes #49 
Fixes #156 
Fixes #139 

- Introduce new `Watch` action for observing resource events outside of the admission controller lifecycle
- Introduces `Pepr Store` for lightweight key-value stores persistence
- Introduces new `K8s(kind.<Name>)` fluent api from [kubernetes-fluent-client](https://github.com/defenseunicorns/kubernetes-fluent-client)
- moves node-fetch wrapper & model to GVK mapping to [kubernetes-fluent-client](https://github.com/defenseunicorns/kubernetes-fluent-client)
- Moves back to arrow functions vs binds due to complexities with `Pepr Store` binding sequence
- Add higher test coverage around core library code
- Leverage SSA for deploying K8s components in journey tests and `npx pepr deploy`
- Add internal `destroyModule()` for purging a pepr module from a cluster


jeff-mccoy:
The method binding pattern proved to be very messy with the interactions between the `Controller`/`PeprControllerStore`/`Capability`/`PeprStore` classes. It proved too brittle (even with additional lint rules) to avoid runtime-only failures when bindings weren't set right vs. just using arrow functions, which guarantee `this scoping`.  Moving to binding in #238 was initially done to support sinon, which was dropped and the hopes for a more memory-efficient use of class methods. Unfortunately, arrow functions and runtime-bound function calls using `myFunction.bind(this)` format still take up the same memory.

src/lib/assets/store.ts
```diff
@@ -0,0 +1,48 @@
...
+                  type: "string",
+                },
+              },
+            },
+          },
+        },
+      },
+    ],
+    scope: "Namespaced",
+    names: {
```
cmwylie19, line 43: do we want to throw a shortname here?

src/lib/controller/store.ts
cmwylie19, line 1: impressive 🤩

src/lib/k8s/fluent/kube.ts
cmwylie19, line 1: elegant

cmwylie19 (dismissed):
lgtm

src/lib/assets/store.ts
jeff-mccoy: I actually chose not to in order to avoid possible conflicts, it would be a minor convenience, but decided it wasn't worth the conflict risk. Thoughts?

src/lib/k8s.ts
```diff
@@ -13,30 +11,21 @@ export enum Operation {
...
-  /** The K8s resource kind, e..g "Pod". */
-  readonly kind: string;
-  readonly group: string;
-  readonly version?: string;
-  /** Optional, override the plural name for use in Webhook rules generation */
-  readonly plural?: string;
-}
+export const peprStoreGVK = {
+  kind: "PeprStore",
+  version: "v1",
```
bdw617, line 24: do we want to follow standards on using alpha/beta here?

bdw617 (commented):
will keep looking at this

src/lib/k8s.ts
jeff-mccoy: I think it's fine as v1 since we have no intention of ever changing, it's effectively a custom ConfigMap to avoid things tracking it for performance reasons.

README.md
```diff
@@ -12,10 +12,15 @@
 
 Pepr is on a mission to save Kubernetes from the tyranny of YAML, intimidating glue code, bash scripts, and other makeshift solutions. As a Kubernetes controller, Pepr empowers you to define Kubernetes transformations using TypeScript, without software development expertise thanks to plain-english configurations. Pepr transforms a patchwork of forks, scripts, overlays, and other chaos into a cohesive, well-structured, and maintainable system. With Pepr, you can seamlessly transition IT ops tribal knowledge into code, simplifying documentation, testing, validation, and coordination of changes for a more predictable outcome.
 
+#### _Note: Pepr is still in active development so breaking changes may occur, but will be document in release notes._
+
 ## Features
 
-- Zero-config K8s webhook mutations and validations.
+- Zero-config K8s webhook mutations and validations
+- Automatic leader-elected K8s resource watching
```
btlghrants, line 20: So... did we decide that "leader-elected" was overkill for the current impl of Watch then?  Should this still be in here, or did I misunderstand what we talked about during sync today (totally possible)?

README.md
```diff
@@ -27,7 +32,7 @@ Pepr is on a mission to save Kubernetes from the tyranny of YAML, intimidating g
 
 ## Example Pepr Action
 
-This quick sample shows how to react to a ConfigMap being created or updated in the cluster. It adds a label and annotation to the ConfigMap and adds some data to the ConfigMap. Finally, it logs a message to the Pepr controller logs. For more see [actions](./docs/actions.md).
+This quick sample shows how to react to a ConfigMap being created or updated in the cluster. It adds a label and annotation to the ConfigMap and adds some data to the ConfigMap. It also creates a Validating Webhook to make sure the "pepr" label still exists. Finally, after the ConfigMap is created, it logs a message to the Pepr controller logs and creates or updates a separate ConfigMap with the [kubernetes-fluent-client](https://github.com/defenseunicorns/kubernetes-fluent-client) using server-side apply. For more details see [actions](./docs/actions.md) section.
```
btlghrants: `It also creates a Validating Webhook to make sure the "pepr" label still exists.`

Does it?  I was sorta expecting to see a .Validate() call in there but is this something the .Mutate() also adds?

btlghrants: `it logs a message to the Pepr controller ~~logs~~`

Random unimportant nit -- would read better with out doubling up on "log".

docs/actions.md
```diff
@@ -4,7 +4,7 @@ An action is a discrete set of behaviors defined in a single function that acts
 
 For example, an action could be responsible for adding a specific label to a Kubernetes resource, or for modifying a specific field in a resource's metadata. Actions can be grouped together within a Capability to provide a more comprehensive set of operations that can be performed on Kubernetes resources.
 
-Actions are either `Mutate()` or `Validate()` actions. Mutate actions are used to modify Kubernetes resources, while Validate actions are used to validate Kubernetes resources.
+Actions are `Mutate()`, `Validate()`, or `Watch()` actions. Both Mutate and Validate actions run during the admission controller lifecycle, while Watch actions run in a separate controller that watches for changes to resources, including existing resources.
```
btlghrants, line 7: "Actions are `Mutate()`, `Validate()`, or `Watch()` ~~actions~~." 

Dupe'd word.

docs/store.md
```diff
@@ -0,0 +1,42 @@
+# Pepr Store: A Lightweight Key-Value Store for Pepr Modules
+
+The nature of admission controllers and general watch operations (the `Mutate`, `Validate` and `Watch` actions in Pepr) make some types of complex and long-running operations difficult. There are also times when you need to share data between different actions. While you could manually create your own K8s resources and manager their cleanup, this can be very hard to manage or keep performant at scale.
```
btlghrants, line 3: "While you could manually create your own K8s resources and manage~~r~~ their cleanup, this can be very hard to ~~manage or~~ keep performant at scale."

Word stuff.

README.md
jeff-mccoy: no I think we can reword that--the effect is the same, but it's not a technical leader-election process anymore. open to suggestions.

docs/store.md
```diff
@@ -0,0 +1,42 @@
+# Pepr Store: A Lightweight Key-Value Store for Pepr Modules
+
+The nature of admission controllers and general watch operations (the `Mutate`, `Validate` and `Watch` actions in Pepr) make some types of complex and long-running operations difficult. There are also times when you need to share data between different actions. While you could manually create your own K8s resources and manager their cleanup, this can be very hard to manage or keep performant at scale. 
+
+The Pepr Store solves this by exposing a simple API compatible with the standard [Web Storage API](https://developer.mozilla.org/en-US/docs/Web/API/Storage) for use within capabilities that abstracts away the complexity of managing the data in K8s. Additionally, as Pepr runs multiple replicas of the admission controller and a single watch controller, the Pepr Store provides a unique way to share data between these different instances automatically.
```
btlghrants, line 5: ~~"The Pepr Store solves this by exposing a simple API compatible with the standard [Web Storage API](https://developer.mozilla.org/en-US/docs/Web/API/Storage) for use within capabilities that abstracts away the complexity of managing the data in K8s."~~

"The Pepr Store solves this by exposing a simple, [Web Storage API](https://developer.mozilla.org/en-US/docs/Web/API/Storage)-compatible data storage mechanism for use within capabilities."

Brevity.

docs/store.md
```diff
@@ -0,0 +1,42 @@
...
+
+Each Pepr Capability has a `Store` instance that can be used to get, set and delete data as well as subsribe to any changes to the Store. Behind the scenes, all capability stores instances in a single Pepr Module are stored within a single CRD in the cluster. This CRD is automatically created when the Pepr Module is deployed. Care is taken to make the read and write operations as efficient as possible by using K8s watches, batch processing and patch operations for writes.
```
btlghrants: "Each Pepr Capability has a `Store` instance that can be used to get, set and delete *arbitrary* data as well as *subscribe* to any changes to the Store."

Explanatory & typo.

btlghrants: "Behind the scenes, all capability stores instances in a single Pepr Module are stored within a single CRD in the cluster. "

Not 100% on what this is saying.  Something about "all capability stores instances in a single Pepr Module are stored" feels like the plurality/subject aren't clear.

btlghrants: "Care is taken to make the read and write operations as efficient as..."

Something about "Care is taken" feels tonally off -- something like "Focus has been put on making the read and write operations... etc." would read better.

README.md
btlghrants:
```suggestion
This quick sample shows how to react to a ConfigMap being created or updated in the cluster. It adds a label and annotation to the ConfigMap and adds some data to the ConfigMap. It also creates a Validating Webhook to make sure the "pepr" label still exists. Finally, after the ConfigMap is created, it logs a message to the Pepr controller and creates or updates a separate ConfigMap with the [kubernetes-fluent-client](https://github.com/defenseunicorns/kubernetes-fluent-client) using server-side apply. For more details see [actions](./docs/actions.md) section.
```

docs/actions.md
btlghrants:
```suggestion
Actions are `Mutate()`, `Validate()`, or `Watch()`. Both Mutate and Validate actions run during the admission controller lifecycle, while Watch actions run in a separate controller that tracks changes to resources, including existing resources.
```

docs/store.md
jeff-mccoy: Yeah this whole thing might just need to be rewritten. Probably should have just had chatgpt write it instead of me 😆

README.md
```diff
@@ -12,10 +12,15 @@
 
 Pepr is on a mission to save Kubernetes from the tyranny of YAML, intimidating glue code, bash scripts, and other makeshift solutions. As a Kubernetes controller, Pepr empowers you to define Kubernetes transformations using TypeScript, without software development expertise thanks to plain-english configurations. Pepr transforms a patchwork of forks, scripts, overlays, and other chaos into a cohesive, well-structured, and maintainable system. With Pepr, you can seamlessly transition IT ops tribal knowledge into code, simplifying documentation, testing, validation, and coordination of changes for a more predictable outcome.
 
+#### _Note: Pepr is still in active development so breaking changes may occur, but will be document in release notes._
```
Noxsios:
```suggestion
#### _Note: Pepr is still in active development so breaking changes may occur, but will be documented in release notes._
```

journey/k8s.ts
```diff
@@ -1,52 +1,50 @@
...
-kc.loadFromDefault();
+import { PeprStore } from "../src/lib/k8s";
 
-const k8sApi = kc.makeApiClient(AppsV1Api);
-const k8sCoreApi = kc.makeApiClient(CoreV1Api);
-
-function delay2Secs() {
+export function delay2Secs() {
   return new Promise(resolve => setTimeout(resolve, 2000));
 }
```
Noxsios:
```suggestion
export function sleep(seconds: number) {
  return new Promise(resolve => setTimeout(resolve, seconds*1000));
}
```

idk, when I do stuff like this, I like calling the fn `sleep` so it looks just like Python's `time.Sleep(n)`

src/lib/assets/rbac.ts
```diff
@@ -55,3 +55,41 @@ export function serviceAccount(name: string): ServiceAccount {
     },
   };
 }
+
+export function storeRole(name: string): kind.Role {
+  name = `${name}-store`;
+  return {
+    apiVersion: "rbac.authorization.k8s.io/v1",
```
Noxsios, line 62: is there any way we could make an enum out of known API versions? prob not

src/lib/controller/store.ts
```diff
@@ -0,0 +1,197 @@
...
+import { K8s } from "kubernetes-fluent-client";
+import { startsWith } from "ramda";
+
+import { Capability } from "../capability";
+import { PeprStore } from "../k8s";
+import Log from "../logger";
+import { ModuleConfig } from "../module";
+import { DataOp, DataSender, DataStore, Storage } from "../storage";
+
+const namespace = "pepr-system";
```
Noxsios, line 14: prob not a concern yet, but what is the plan for pepr in a multi-tenant cluster, where you might not have access to create your own namespaces? though there are tools to create virtual clusters within a namespace

Noxsios (commented):
my 2 🪙 feel free to ignore

docs/actions.md
```diff
@@ -52,6 +52,18 @@ When(a.ConfigMap)
   });
 ```
 
+---
+
+In this example, a Watch action the name and phase of any ConfigMap.
```
cmwylie19:
```suggestion
In this example, a Watch action on the name and phase of any ConfigMap.
```

docs/store.md
cmwylie19:
```suggestion
Each Pepr Capability has a `Store` instance that can be used to get, set and delete data as well as subscribe to any changes to the Store. Behind the scenes, all capability store instances in a single Pepr Module are stored within a single CRD in the cluster. This CRD is automatically created when the Pepr Module is deployed. Care is taken to make the read and write operations as efficient as possible by using K8s watches, batch processing and patch operations for writes.
```

bdw617 (dismissed):
Store.setItem("tenantGateway", "istio-system/tenant");
  Store.getItem("tenantGateway")
  
  produces nothing.
Store.onReady happens after the controller starts, will expand testing to Mutate/Validate. Right now the Store not usable (at least with Watch)

src/lib/assets/deploy.ts
```diff
@@ -1,179 +1,121 @@
...
+    throw new Error("No code provided");
+  }
+
+  await setupRBAC(name);
+  await setupController(assets, code, hash);
+  await setupWatcher(assets, hash);
+}
 
+async function setupRBAC(name: string) {
+  Log.info("Applying cluster role binding");
```
cmwylie19, line 65: not that it's a big deal at all but weird to see the `crb` before the `cr`

README.md
btlghrants:
```suggestion
- Cluster event-driven resource watching
```
...or maybe just cut it since it's not really selling much of anything anymore?

docs/store.md
btlghrants:
```suggestion
The nature of admission controllers and general watch operations (the `Mutate`, `Validate` and `Watch` actions in Pepr) make some types of complex and long-running operations difficult. There are also times when you need to share data between different actions. While you could manually create your own K8s resources and manage their cleanup, this can be very hard to track and keep performant at scale. 
```

docs/store.md
btlghrants:
```suggestion
The Pepr Store solves this by exposing a simple, [Web Storage API](https://developer.mozilla.org/en-US/docs/Web/API/Storage)-compatible mechanism for use within capabilities. Additionally, as Pepr runs multiple replicas of the admission controller along with a watch controller, the Pepr Store provides a unique way to share data between these different instances automatically.
```

src/lib/assets/store.ts
cmwylie19: sorry for the late reply, possible conflicts with other `ps` resources?

I'm okay either way. At Solo, Gloo Edge had shortName for `vs` which would clobber Istio's `vs`, the work around was fairly easy to remove the CRD. 

As kubernetes tool i like the shortname because it feels more native and I use shortnames constantly anyway, `k get po,svc,deploy,cm,etc`

src/lib/controller/store.ts
cmwylie19: IMO in a multi-tenant cluster, the Pepr Capability would need to be restricted to the namespaces owned by said tenant which is possibly when initializing a new capability:

```typescript
export const ExampleCapability = new Capability({
  name: "example-capability",
  description: "Restricted to a couple of namespaces.",
  namespaces: ["tenant-a", "tenant-b"],  // <- define ns here
});
```

cmwylie19: but would definitely need to be installed by some `user` or `sa` that has the permissions to install cluster scoped objects for the `cr` and `crb` and forgetting whatever else is not `ns` scoped

cmwylie19 (dismissed):
lgtm

src/lib/controller/store.ts
jeff-mccoy: the real issue is even if you scope to a ns, webhooks and clusterrole need admin privs to create, so you dont get much by scoping down as you already need the permissions to generate the global things anyway

jeff-mccoy:
> Store.setItem("tenantGateway", "istio-system/tenant"); Store.getItem("tenantGateway")
> 
> produces nothing. Store.onReady happens after the controller starts, will expand testing to Mutate/Validate. Right now the Store not usable (at least with Watch)

right we brought that up yesterday, working that fix rn. Mutate/Validate should work, but you can't getItems until the store is ready--but today it doesnt throw an error.

README.md
jeff-mccoy: resource watching is a major feature--but leader-election isn't even a problem most ppl are aware of

btlghrants (dismissed):
Don't see any red flags so let's do it, wee!

src/lib/assets/rbac.ts
jeff-mccoy: it can be pretty dynamic and even confusing (see events vs audit events), probably best to not for now.

cmwylie19 (approved):
lgtm

bdw617 (approved):
with caveats

btlghrants (approved):
:+1:

PR #186, opened 2023-07-11
Fix k8s import/export

Noxsios:
Fixes the k8s export so it can be used properly.  Unsure how this affects the bundle as before the k8s re-export was not actually doing anything and now it is properly pulling in all of `@kubernetes/client-node`.

Pretty sure that this will increase the size of the JS by a bit because its now importing _everything_ from the k8s client and not tree shaking imports.


jeff-mccoy (approved):
Thanks that should be fine as we don't bundle this lib intentionally, it's only on disk in the controller--which is why we reexport to begin with.

PR #203, opened 2023-08-04
check if kind.kind ends in y, if so remove y and add ies, else is the…

### LINKED ISSUE #201: Incorrect MutatingWebhookConfiguration
When using `a.GenericKind` Pepr [assumes](https://github.com/defenseunicorns/pepr/blob/419510839102ce34268230d48f9648fb0d1e2f05/src/lib/k8s/webhook.ts#L217) the plural for resources. We need logic to check if the kind.kind ends in y, and if so, add `ies` instead of `s`

Example of error:

in Capability
```javascript
When(a.GenericKind, {
  group: "source.toolkit.fluxcd.io",
  version: "v1beta2",
  kind: "GitRepository",
})
```

Resulting MutatingWebhookConfiguration
```yaml
  - apiGroups:
    - source.toolkit.fluxcd.io
    apiVersions:
    - v1beta2
    operations:
    - CREATE
    - UPDATE
    resources:
    - gitrepositorys
```

expectation
```bash
gitrepositories                   gitrepo            source.toolkit.fluxcd.io/v1beta2       true         GitRepository
```

No biggy, i can work around this and I will pick up this PR.

cmwylie19:
… same

Closes #201 by checking if `kind.kind` ends in `y`, if so, removes the `y` and adds `ies`. @jeff-mccoy lmk if there is a better place for this function

### DIFF
--- a/src/lib/k8s/webhook.ts
+++ b/src/lib/k8s/webhook.ts
@@ -213,8 +213,8 @@ export class Webhook {
               operations.push(event);
             }
 
-            // Use the plural property if it exists, otherwise use lowercase kind + s
-            const resource = kind.plural || `${kind.kind.toLowerCase()}s`;
+            // Use the plural property if it exists, otherwise use lowercase kind + s or -y + ies if it ends in y
+            const resource = kind.plural || createPluralFromKind(kind.kind.toLowerCase());
 
             rules.push({
               apiGroups: [kind.group],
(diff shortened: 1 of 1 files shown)


cmwylie19:
The way to handle this is with plurals

```javascript
When(a.GenericKind, {
  group: "source.toolkit.fluxcd.io",
  version: "v1beta2",
  kind: "GitRepository",
  plural: "gitrepositories",
});
```

Resolved comments in issue #201

PR #205, opened 2023-08-07
updated metrics for validate/mutate.

bdw617:
can also handle more metrics, could hook into the capabilities fairly easily.


docs/metrics.md
```diff
@@ -8,7 +8,8 @@ The `MetricsCollector` exposes the following metrics:
 
 - `pepr_errors`: A counter that increments when an error event occurs in the application.
 - `pepr_alerts`: A counter that increments when an alert event is triggered in the application.
-- `pepr_summary`: A summary that provides the observed durations of events in the application.
+- `pepr_Mutate`: A summary that provides the observed durations of mutation events in the application.
```
jeff-mccoy, line 11: does it matter that these are capitalized and the previous one wasn't?

bdw617: welllllll. I'm using Mutate and Validate strings internally, so we could lower case them, If this is killing you, we can do that.

bdw617:
will redo this.

PR #208, opened 2023-08-14
Dependency Updates (including peer deps) & Dependabot Grouping

jeff-mccoy:


bdw617 (approved):
looks ok, we will need to make sure the typescript change doesn't break anything downstream

PR #228, opened 2023-08-28
Base64 decode and encode non-ASCII values in secret data

### LINKED ISSUE #219: Support for base64 encoding/decoding non-ASCII characters in Secret data
**Given** I need to read/write data that contains non-ASCII characters from/to a Kubernetes secret
**When** I use Pepr to read/write the data
**Then** Pepr encodes and decodes the data for me

Context:

The current behavior of Pepr not encoding/decoding non-ASCII characters is intentional, see https://github.com/defenseunicorns/pepr/pull/131. That PR relates to issue https://github.com/defenseunicorns/pepr/issues/129, which may be related to https://github.com/rancher/rancher/issues/27439.

### LINKED ISSUE #279: secrets, base64 and binary versus text need to be easier to use and simpler in the code.
right now we do a best effort for base64 decoding/encoding the secrets. In golang, the underlying kubernetes API decodes/encodes for you, which it can since you can put binary data in a string in golang. In javascript you can't. 

@jeff-mccoy built a system for keeping track of how we change things, but I think it might be better to just not modify the existing secrets. We can provide convenience functions on top of the secret to make it easier for people but the integration into the API. The current implementation has issues:

read values from a secret as text, then you need to write binary, you need to own the base64encoding of it. 
read values from a secret as binary, but if you replace it with text or binary, you need to base64 encode it again. 

Creating functions on top of a secret so you don't directly work with the Raw data will help both in capabilities and in working with the fluent api directly. This would remove having users know anything about base64 and not worry about the underlying secret storage.

The V1Secret model (it's got an example of 1/2 fixing this problem with stringData).
https://github.com/kubernetes-client/javascript/blob/0.19.0/src/gen/model/v1Secret.ts#L40

### Additional context

lucasrod16:
Closes https://github.com/defenseunicorns/pepr/issues/219


bdw617 (changes requested):
This code is fine, but we need to hold up this PR until the Validate decoding works, I'll PR that too.

jeff-mccoy:
So one concern with this is it breaks the type contract for the upstream k8s type. It expects a string, going to play with this a little as javascript does some fun magic with auto string conversion that _might_ make this sort of okay. More than likely though we'll want to change the class to account for the `Buffer` vs `string` type.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#type_coercion


Screenshot 2023-08-31 at 12 02 13 AM

jeff-mccoy:
Will also need to migrate these unit tests to Jest, https://github.com/skovhus/jest-codemods may help

src/lib/utils.ts
```diff
@@ -3,55 +3,63 @@
...
-  for (const key in obj.data) {
-    const value = obj.data[key];
-    // Only encode ascii values
-    obj.data[key] = skip.includes(key) ? value : base64Encode(value);
+interface Base64Map {
+  data?: Record<string, Base64ConvertibleValue>;
+}
+
+/** Base64 encode all values in a map */
+export function convertToBase64Map(obj: Base64Map): Base64Map {
```
bdw617, line 13: so looking at the main code. we're not actually copying the object, only creating a place for the skip. Since we're getting rid of the skip completely, we can overwrite the data in the existing object and not return anything.

// here's what I think it should be, we could even get rid of the first line and replace it with obj?.data, but downstream it might expect that to be {} versus undefined.
```
export function convertToBase64Map(obj: { data?: Record<string, string> }) {
  obj.data = obj.data ?? {};
  for (const key in obj.data) {
    obj.data[key] =  base64Encode(obj.data[key]);
  }
}
```
and
```
export function convertFromBase64Map(obj: { data?: Record<string, string> }) {
  obj.data = obj.data ?? {};
  for (const key in obj.data) {
    if (obj.data[key] == undefined) {
      obj.data[key] = "";
    } else {
      const decoded = base64Decode(obj.data[key]);
        obj.data[key] = decoded;
    }     
  }
}
```

src/lib/utils.ts
```diff
@@ -3,55 +3,63 @@
...
-      }
+export function base64Decode(data: Base64ConvertibleValue): Buffer {
+  try {
+    if (typeof data === "string") {
+      return Buffer.from(data, "base64");
+    } else if (Buffer.isBuffer(data)) {
+      return data;
     }
+  } catch (error) {
+    Log.error(`Error decoding base64 data: ${error}`);
```
bdw617, line 52: can you reproduce this problem in kubernetes? I wasn't able to shove invalid base64 data into a .spec.data[]

bdw617 (changes requested):
it's probably good to make these changes, it keeps it more consistent with the existing code.

cmwylie19:
closing this because this will likely be handled through helper functions. Thanks so much Lucas for the contribution. We will keep the branch

PR #232, opened 2023-08-30
Validate will now decode base64 secret data

### LINKED ISSUE #231: 0.13.0 does different things with secrets between validate and mutate.
Mutate works, Validate doesn't decode the base64 for you. @lucasrod16's change to Mutate helped me find this :)

bdw617:


jeff-mccoy (approved):
prob need a test for validate secret things, but I think #228 will cover that

PR #236, opened 2023-08-31
visible errors

### LINKED ISSUE #233: `pepr build` can silently fail
If there is a json syntax error in package.json, `npx pepr build` has no output and silently fails:

```
$ npx pepr build -l debug
{"level":20,"time":1693417225078,"pid":12932,"hostname":"Blakes-MacBook-Pro.local","msg":"Unexpected token / in JSON at position 301"}
```

If the user has no `node_modules/` (has not yet run `npm install`), `npx pepr build` has no output and silently fails:
```
$ npx pepr build -l debug
{"level":20,"time":1693417811820,"pid":13682,"hostname":"Blakes-MacBook-Pro.local","msg":"Command failed: ./node_modules/.bin/tsc\n/bin/sh: ./node_modules/.bin/tsc: No such file or directory\n"}
{"level":20,"time":1693417811821,"pid":13682,"hostname":"Blakes-MacBook-Pro.local","msg":""}
{"level":20,"time":1693417811821,"pid":13682,"hostname":"Blakes-MacBook-Pro.local","msg":"/bin/sh: ./node_modules/.bin/tsc: No such file or directory\n"}
```

These errors should print an error message.

bdw617:
if you have a syntax error in your module, `npx pepr build` will now tell you without added `-l debug`
errors will not be logged as Error, so in the event any build function fails there's a visible message

### DIFF
--- a/src/cli/build.ts
+++ b/src/cli/build.ts
@@ -177,14 +177,14 @@ export async function buildModule(reloader?: Reloader, entryPoint = peprTS) {
 
     return { ctx, path, cfg, uuid };
   } catch (e) {
-    Log.debug(e.message);
+    Log.error(e.message);
 
     if (e.stdout) {
       const out = e.stdout.toString() as string;
       const err = e.stderr.toString();
 
-      Log.debug(out);
-      Log.debug(err);
+      Log.error(out);
+      Log.error(err);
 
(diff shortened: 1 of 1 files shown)


jeff-mccoy (approved):
lgtm, thx

PR #246, opened 2023-09-06
added a doc on how to use aws sdk within pepr

bdw617:


jeff-mccoy (changes requested):
This is very specific and not sure it belongs in the Pepr docs, will have to think abou this one.

bdw617:
@jeff-mccoy ok, but we need to demonstrate capabilities that use APIs like AWS. I've pushed @lucasrod16 to use IAC for provisioning AWS resources versus trying to jam this into pepr). There are other complexities to creating an ECR within pepr that involve IAM and VPCs :|

PR #248, opened 2023-09-07
Fix capability namespace list being ignored

### LINKED ISSUE #243: capability namespaces ignored in 0.13.1 (working in shifting-sands)
tested in 0.13.1 and the callback hits even if the pod's namespace is default. In shifting-sands, it works as expected. not sure we want to fix it, but if we need to make sure there's a test for it...

export const HelloPepr = new Capability({
  name: "test",
  description: "test",
  namespaces: ["test"],
});

const { When } = HelloPepr;

When(a.Pod)
  .IsCreated()
  .Mutate(p => {
      Log.info("we're here!")
  })

jeff-mccoy:
Fixes #243.

This PR correctly handles namespace filters provided when a Capability is created, such as in HelloPepr.

### DIFF
--- a/journey/k8s.ts
+++ b/journey/k8s.ts
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: 2023-Present The Pepr Authors
 
-import { AppsV1Api, CoreV1Api, KubeConfig } from "@kubernetes/client-node";
+import { AppsV1Api, CoreV1Api, KubeConfig, V1ConfigMap } from "@kubernetes/client-node";
 
 const kc = new KubeConfig();
 kc.loadFromDefault();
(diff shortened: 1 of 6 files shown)


jeff-mccoy:
Added unit test and 2 journey test to validate functionality, see images below before the code change was applied. 

Screenshot 2023-09-06 at 11 38 14 PM


Screenshot 2023-09-06 at 11 50 29 PM

bdw617:
I'll run through some e2e tests today to see how it works... making sure it keeps things aligned with shifting-sands

bdw617 (commented):
I've tested this and it works, but this is maybe not the way I would have implemented it.

Why not like this in the mutatingwebhook configuration?

```
  namespaceSelector:
    matchExpressions:
    - key: pepr.dev
      operator: NotIn
      values:
      - ignore
    - key: kubernetes.io/metadata.name
      operator: In
      values:
      - pepr-demo
```

jeff-mccoy:
We would have to do this selection regardless because each capability can have different namespace filters, challenge is as soon as any capability doesnt have a namespace filter you'd have to remove it from the webhook config, so keep it here I think makes the most sense.

bdw617 (approved):

PR #249, opened 2023-09-07
Make the user templates more accessible

jeff-mccoy:
This PR moves the template files (used during `npx pepr init` and `npx pepr update`) from `src/cli/init/templates/` → `src/templates/`


bdw617 (approved):

cmwylie19 (approved):
lgtm

PR #250, opened 2023-09-07
updated kinds.test to a loop.

bdw617:
updated kinds unit test to use an array and loop, also added two tests to complete coverage. In the event a test fails it will log it and throw an exception so we know specifically which one has failed.


cmwylie19 (dismissed):
e2e failing

jeff-mccoy (approved):
much cleaner, lgtm

PR #257, opened 2023-09-08
.gitignore and function return type

### LINKED ISSUE #256: pepr-test-module should be added to .gitignore
Since pepr-test-module does not belong in the repo but is pertinent to developing we should add it to the gitignore

cmwylie19:
Update .gitignore to include `pepr-test-module` and add a return type for `generateWebhookRules` function

Closes #256

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,7 @@
 
+# pepr-test-module should be ignored
+pepr-test-module/
+
 # Binaries for programs and plugins
 *.exe
 *.exe~
--- a/src/lib/assets/webhooks.ts
+++ b/src/lib/assets/webhooks.ts
@@ -20,7 +20,7 @@ const peprIgnoreLabel: V1LabelSelectorRequirement = {
 
 const peprIgnoreNamespaces: string[] = ["kube-system", "pepr-system"];
 
-export async function generateWebhookRules(assets: Assets, isMutateWebhook: boolean) {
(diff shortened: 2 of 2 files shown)


src/lib/assets/webhooks.ts
bdw617, line 1: not sure this change matters? @jeff-mccoy prefers documentation for functions over explicit return types

bdw617 (changes requested):

.gitignore
```diff
@@ -1,4 +1,7 @@
 
+# pepr-test-module should be ignored
+pepr-test-module/
```
jeff-mccoy, line 3: this is already on line 21, was there a reason to add it here too?

src/lib/assets/webhooks.ts
jeff-mccoy: style-wise I typically leave any types can be inferred as such in typescript, return stuff is just jsdoc and not really enforced by the IDE

jeff-mccoy: same reason I don't do `const words:string = "words";` but instead do `const words = "words";`

cmwylie19: Got it, i thought this was a one off but i could have checked more.

.gitignore
cmwylie19: oversight, saw it in my editor as new code and didn't realize

PR #316, opened 2023-10-12
chore: security updates

cmwylie19:
## Description

Provide security patches mainly around CI 

## Related Issue

Fixes #

Relates to #

## Type of change

## Checklist before merging

### DIFF
--- a/.github/workflows/commitlint.yml
+++ b/.github/workflows/commitlint.yml
@@ -21,5 +21,8 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@5e21ff4d9bc1a8cf6de233a3057d20ec6b3fb69d # v3.8.1
 
+      - name: Install commitlint
+        run: npm install --save-dev @commitlint/{config-conventional,cli}
+
       - name: Lint PR title
         run: echo "${{ github.event.pull_request.title }}" | npx commitlint@v17.7.2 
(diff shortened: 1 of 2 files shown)


.github/workflows/commitlint.yml
```diff
@@ -24,4 +25,4 @@ jobs:
         run: npm install --save-dev @commitlint/{config-conventional,cli}
 
       - name: Lint PR title
-        run: echo "${{ github.event.pull_request.title }}" | npx commitlint
+        run: echo "${{ github.event.pull_request.title }}" | npx commitlint@v17.7.2
```
btlghrants: Maybe I'm just unfamiliar with the syntax here, but it looks like we're `npm install`ing commit lint at no-particular-version (so, latest..?) but then we're specifying v17.7.2 when we run it.  Is that right?

What happens when a new version of commitlint gets released?  Wouldn't this `npm install` the newer version but then _also_ try to run commit lint with @v17.7.2?

cmwylie19: I believe that is correct, good catch. Maybe we can take out that `npm i` completely? I don't think it is doing anything

btlghrants: Or, wait, I guess `npm install` is just going to install the version we have stored in the package-lock.json file, innit?  In that case, _why_ do we need to specify the version in our workflow file too?

cmwylie19: this is resolved together through a conversation in gather.

btlghrants:
```suggestion
        run: echo "${{ github.event.pull_request.title }}" | npx commitlint
```
We decided that we shouldn't need this cuz it's spec'd in the package-lock.json file, right?

btlghrants (approved):
Looks good; just like we talked 'bout.

PR #324, opened 2023-10-17
feat: least priv rbac creation

### LINKED ISSUE #31: Make ClusterRole, Webhook bindings restricted to current module

cmwylie19:
## Description

Build time flag `--rbac-mode` determines whether whether the `ClusterRole` for the Pepr `ServiceAccount` receives cluster-admin permissions (default) or scoped permissions based on the capability bindings. 

Uses a reduce function to iterate over the capability's bindings and determine the `verbs`, `plural`, and `groups` needed in order to build a `ClusterRoleBinding`. That function is used in the `createClusterRole` function.

Includes:
- docs
- development
- unit test
- e2e test

## Related Issue

Fixes #31 

Relates to #

## Type of change

## Checklist before merging


src/lib/assets/index.ts
```diff
@@ -40,8 +40,8 @@ export class Assets {
 
   zarfYaml = (path: string) => zarfYaml(this, path);
 
-  allYaml = async () => {
+  allYaml = async (rbacMode: string = "") => {
```
btlghrants: I think I remember us talking about how the only two acceptable rbacMode values would be `admin` or `scoped` -- it's what we're enforcing in the cli cmds, right?   And, since we've decided that the default-right-now behavior would be considered `admin`, I wonder what the utility of setting a default value of empty string here would be?  As is, this implies that empty string is a valid value but is it?

I would sort have expected this to have a default value of "admin" (if we wanted to push the validation into the lib code) or to not have a default at all (if we wanted to just rely on the cli invocation).

src/lib/assets/yaml.ts
```diff
@@ -40,7 +40,7 @@ export function zarfYaml({ name, image, config }: Assets, path: string) {
   return dumpYaml(zarfCfg, { noRefs: true });
 }
 
-export async function allYaml(assets: Assets) {
+export async function allYaml(assets: Assets, rbacMode: string = "") {
```
btlghrants: Similar to question in src/lib/assets/index.ts, what's the purpose of a default empty string for rbacMode here?  Seems a non-default-valued param would be better here because it wouldn't imply that "" was an expected value.  :thinking:

btlghrants (approved):
LGTM

PR #352, opened 2023-10-31
feat: add flag to change directory of built artifacts

### LINKED ISSUE #350: build output flag

Provide a flag during `npx pepr build` for output to define where the contents of the dist folder go. This request came up during consideration for module testing

- **Given** you are writing module tests and want to test the `dist` folder contents
- **When** you issue `npx pepr build -o somedir` 
- **Then** `somedir` has the contents of the `dist` folder

cmwylie19:
## Description

Provide a flag during `npx pepr build` for output to define where the contents of the dist folder go. This request came up during consideration for module testing.

## Related Issue

Fixes #350 

Relates to #

## Type of change

## Checklist before merging

### DIFF
--- a/journey/pepr-build-wasm.ts
+++ b/journey/pepr-build-wasm.ts
@@ -9,17 +9,22 @@ import { resolve } from "path";
 
 import { cwd } from "./entrypoint.test";
 
+// test npx pepr build -o dst
+const outputDir = "dst";
 export function peprBuild() {
+
+  execSync(`mkdir ${outputDir}`);
+
   it("should successfully build the Pepr project with arguments", async () => {
-    execSync("npx pepr build -r gchr.io/defenseunicorns --rbac-mode scoped", { cwd: cwd, stdio: "inherit" });
+    execSync(`npx pepr build -r gchr.io/defenseunicorns --rbac-mode scoped -o ${outputDir}`, { cwd: cwd, stdio: "inherit" });
   });
 
   it("should generate produce the K8s yaml file", async () => {
-    await fs.access(resolve(cwd, "dist", "pepr-module-static-test.yaml"));
+    await fs.access(resolve(cwd, outputDir, "pepr-module-static-test.yaml"));
   });
 
   it("should generate a custom image in zarf.yaml", async () => {
-    await fs.access(resolve(cwd, "dist", "zarf.yaml"));
+    await fs.access(resolve(cwd, outputDir, "zarf.yaml"));
     await validateZarfYaml();
   });
 
@@ -30,7 +35,7 @@ export function peprBuild() {
 
 async function validateClusterRoleYaml() {
   // Read the generated yaml files
-  const k8sYaml = await fs.readFile(resolve(cwd, "dist", "pepr-module-static-test.yaml"), "utf8");
+  const k8sYaml = await fs.readFile(resolve(cwd, outputDir, "pepr-module-static-test.yaml"), "utf8");
   const cr = await fs.readFile(resolve("journey", "resources", "clusterrole.yaml"), "utf8");
 
   expect(k8sYaml.includes(cr)).toEqual(true)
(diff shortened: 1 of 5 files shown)


journey/pepr-build-wasm.ts
```diff
@@ -9,17 +9,24 @@ import { resolve } from "path";
 
 import { cwd } from "./entrypoint.test";
 
+// test npx pepr build -o dst
+const outputDir = "dst";
 export function peprBuild() {
+
+  it("should build artifacts in the dst folder", async () => {
+    execSync(`mkdir ${outputDir}`, { cwd, stdio: "inherit" });
```
btlghrants: I'm being a little pedantic here, but... I think I'd prefer that we used node methods to interact with the underlying system where we can... just so we get/stay in the habit of not assuming the programs/versions/supported-flags that are installed on dev machines (and have a better time handling errors, etc.). :thinking: 

Would you be okay with moving to something like this..?
```suggestion
    await fs.mkdir(outputDir, { recursive: true })
```
Note: added that `recursive` option so that if anyone tries to add more path parts to the `outputDir` variable, node will just create the nested structures -- a bit like adding `-p / --parents` to mkdir would.

src/cli/build.ts
```diff
@@ -29,6 +29,10 @@ export default function (program: RootCmd) {
       "-r, --registry-info [<registry>/<username>]",
       "Registry Info: Image registry and username. Note: You must be signed into the registry",
     )
+    .option(
+      "-o, --output-dir [output directory]",
+      "Define where the contents of the dist folder go",
```
btlghrants:
```suggestion
      "Define where to place build output",
```
...or something like it should read more clearly to someone new (who probably won't know what "contents of the dist folder" means.

journey/pepr-build-wasm.ts
```diff
@@ -9,17 +9,24 @@ import { resolve } from "path";
 
 import { cwd } from "./entrypoint.test";
 
+// test npx pepr build -o dst
+const outputDir = "dst";
```
btlghrants: Have we tested that this works with deeper / multi-part paths -- something like `this/folder/might/not/exist/yet/dst`?

btlghrants (changes requested):

journey/pepr-build-wasm.ts
cmwylie19: 100% good eye, should have thought of thought

journey/pepr-build-wasm.ts
cmwylie19: should work now, good eye

src/cli/build.ts
```diff
@@ -29,12 +30,26 @@ export default function (program: RootCmd) {
...
+      if (opts.outputDir) {
+        outputDir = opts.outputDir;
+        createDirectoryIfNotExists(outputDir).catch(error => {
+          console.error(
+            "Could not create output directory. ",
+            error,
+            " Defaulting to dist folder.",
+          );
+          outputDir = "dist";
+        });
```
btlghrants, line 46: I think... I like the error message, but I'm not sure I like the use-dist-by-default, personally. :thinking: 

It feels a bit like it violates the [Principle of Least Surprise](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) in that, if a user is telling Pepr to _not_ use the default directory (ostensibly, for a reason), then Pepr probably _shouldn't_ use it.

Personally, if I'm saying to Pepr, "Do your thing, Pepr, but put the results 'here'" and "here" isn't available (because it's unreachable / or I made a typo or whatever) then I'd want Pepr to just tell me that & exit.  Cuz that's the quickest way for me correct my mistake & try again.

If we just push on using defaults then we're essentially telling the caller that:
1. we know what they "really" want (which isn't what they asked for),
2. it might take them a while to _notice_ that what's happening isn't what they asked for (cuz lots of people don't watch CLI output all that closely as it scrolls by), and
3. it introduces the possibility that we're stomping on things they really want to keep (since, by passing the -o flag, they were attempting to instruct Pepr _not_ to touch the default location).

btlghrants (changes requested):

src/cli/build.ts
cmwylie19: You are right, lets stick to best practices when it comes to user interface/interactions. Changed

btlghrants (approved):
LGreatTM

PR #365, opened 2023-11-09
feat: setItemAndWait transactional way to set an item in the store

### LINKED ISSUE #359: Transactional store during a watch

A transactional store makes sense during a `Watch` binding, but not a `Mutate` or `Validate` due to  [webhook timeouts](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#:~:text=The%20timeout%20value%20must%20be%20between%201%20and%2030%20seconds.&text=The%20timeout%20for%20an%20admission%20webhook%20defaults%20to%2010%20seconds.).

- **Given** there is a need to use the `PeprStore` in a transactional way
- **When** a watch action happens
- **Then** there should be a helper like `setItemAndWait` to `await` that the item exists in the store before moving on.

cmwylie19:
## Description

Provides a function to set an item in the store in a transactional way during a `Watch` action.

## Related Issue

Fixes #359 

Relates to #

## Type of change

## Checklist before merging

### DIFF
--- a/src/lib/storage.test.ts
+++ b/src/lib/storage.test.ts
@@ -21,11 +21,11 @@ describe("Storage", () => {
     expect(mockSender).toHaveBeenCalledWith("add", ["key1"], "value1");
   });
 
-  it("should set an item and wait", async () => {
+  it("should set an item and wait", () => {
     const mockSender = jest.fn();
     storage.registerSender(mockSender);
 
-    await storage.setItemAndWait("key1", "value1");
+    storage.setItemAndWait("key1", "value1");
 
     expect(mockSender).toHaveBeenCalledWith("add", ["key1"], "value1");
   });
--- a/src/lib/storage.test.ts
+++ b/src/lib/storage.test.ts
@@ -21,6 +21,15 @@ describe("Storage", () => {
     expect(mockSender).toHaveBeenCalledWith("add", ["key1"], "value1");
   });
 
+  it("should set an item and wait", async () => {
+    const mockSender = jest.fn();
+    storage.registerSender(mockSender);
+
+    await storage.setItemAndWait("key1", "value1");
+
+    expect(mockSender).toHaveBeenCalledWith("add", ["key1"], "value1");
+  });
+
   it("should remove an item", () => {
     const mockSender = jest.fn();
     storage.registerSender(mockSender);
--- a/src/lib/storage.test.ts
+++ b/src/lib/storage.test.ts
@@ -24,10 +24,12 @@ describe("Storage", () => {
   it("should set an item and wait", () => {
     const mockSender = jest.fn();
     storage.registerSender(mockSender);
+    jest.useFakeTimers();
 
     storage.setItemAndWait("key1", "value1");
 
     expect(mockSender).toHaveBeenCalledWith("add", ["key1"], "value1");
+    jest.clearAllTimers();
   });
 
   it("should remove an item", () => {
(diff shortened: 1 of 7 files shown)


src/lib/storage.ts
```diff
@@ -88,6 +94,27 @@ export class Storage implements PeprStore {
...
+  /**
+   * Creates a promise and subscribes to the store, the promise resolves when
+   * the key and value are seen in the store.
+   *
+   * @param key - The key to add into the store
+   * @param value - The value of the key
+   * @returns
+   */
+  setItemAndWait = (key: string, value: string) => {
+    Log.warn(`setItemAndWait should only be used on a Watch`);
```
cmwylie19: remove warning and rely on docs warning

src/lib/storage.ts
```diff
@@ -88,6 +94,27 @@ export class Storage implements PeprStore {
...
+    this.#dispatchUpdate("add", [key], value);
+    return new Promise<void>(resolve => {
+      const unsubscribe = this.subscribe(data => {
+        if (data[key] === value) {
+          unsubscribe();
+          resolve();
+        }
+      });
```
cmwylie19, line 114: `setTimeout` here to resolve at 10 seconds to clean up subscribe and reject incase there are two manipulations of the same key at the same time

src/lib/storage.test.ts
```diff
@@ -21,6 +21,15 @@ describe("Storage", () => {
     expect(mockSender).toHaveBeenCalledWith("add", ["key1"], "value1");
   });
 
+  it("should set an item and wait", () => {
+    const mockSender = jest.fn();
+    storage.registerSender(mockSender);
+
+    storage.setItemAndWait("key1", "value1");
```
btlghrants:
```suggestion
  it("should set an item and wait", async () => {
    const mockSender = jest.fn();
    storage.registerSender(mockSender);

    await storage.setItemAndWait("key1", "value1");
```

I think... you have to await the result of this call else the set call & the expect assertion are going to be racing each other.

src/lib/storage.ts
```diff
@@ -88,6 +95,31 @@ export class Storage implements PeprStore {
...
+        if (data[key] === value) {
+          unsubscribe();
+          resolve();
+        }
+        // If promise has not resolved before MAX_WAIT_TIME reject
+        setTimeout(() => {
+          unsubscribe();
+          reject();
+        }, MAX_WAIT_TIME);
+      });
```
btlghrants:
```suggestion
      const unsubscribe = this.subscribe(data => {
        if (data[key] === value) {
          unsubscribe();
          resolve();
        }
      });
      
      // If promise has not resolved before MAX_WAIT_TIME reject
      setTimeout(() => {
        unsubscribe();
        reject();
      }, MAX_WAIT_TIME);
```

Not 100% sure here on the syntax (cuz the PR suggestion editor's kinda not-great) but it seems like you'd want to establish the setTimeout() just the one time -- right after you subscribe to the store event -- rather than _every_ time the subscribe handler is called.  No?  :thinking:

btlghrants (changes requested):

src/lib/storage.test.ts
cmwylie19: i saw this, the tests oddly fail using this condition because of how they are mocked. I originally had this implementation.

src/lib/storage.ts
cmwylie19: yes, good catch

src/lib/storage.test.ts
btlghrants: Ahh, yeah... I mean, we might need to update how the mock happens then, I suppose.  It's too weird to have a test that doesn't really simulate the real thing. :see_no_evil:

btlghrants (approved):
Aaaaannnddddd.... action!

Issue #288, opened 2023-10-03
Replace hello-pepr.ts with pepr-excellent-examples repo + docs


bdw617:
### Is your feature request related to a problem? Please describe.
hello-pepr is great, but it's better as a test framework and a stream of examples.

### Describe the solution you'd like

we should discuss in deeper what this should be, but it needs a lot of work.

Should include Prometheus example for scraping metrics with Grafana dashboard with PolicyReport

cmwylie19:
Thanks for the idea. External documentation like a GitHub repo is an excellent thought. We are planning an entire epic where we plan on addressing the underlying concerns to improve the project's documentation and examples.

Here is a [link of the issues](https://github.com/defenseunicorns/pepr/issues?q=is%3Aopen+is%3Aissue+label%3Adocumentation) that we are planning on prioritizing after a few initial high priority items which includes a website where the idea is that it will be similar to [Zarf's](https://docs.zarf.dev/docs/zarf-overview) (in time).

However, `hello-pepr.ts` doesn't need to be removed in order to provide excellent documentation and examples. It is a fundamental part of the current documentation/training process of Pepr and gets the user developing immediately after initializing a Pepr module. External docs/examples and `hello-pepr` can all coexist together. 

I'll tag this issue in a similar issue and we can track progress there.

cmwylie19:
I chatted with Jeff and the team around this and you are right, apologies for closing. 

This issue is more like an epic that is never really done. This repo will start a capability demonstrating [OnSchedule](https://github.com/defenseunicorns/pepr/pull/340)

[WIP](https://github.com/defenseunicorns/pepr-excellent-examples)

**TODO**: brainstorm a way to hook this into CI.

cmwylie19:
This repository will undergo a change where it will consist of folders of actual Pepr modules with `e2e` tests. It will be a continual WIP

cmwylie19:
- mutate
- filters

PR #369, opened 2023-11-13
feat: schedule implementation

### LINKED ISSUE #308: OnSchedule implementation

Add a helper function that only runs in the watch controller to run code on a given schedule.
(RunAt)
```typescript
OnSchedule({
  every: 5,
  unit: "minutes",
  run: () => {
     // run some code
  },
});
```

- **Given** `OnSchedule` is created from a capability
- **When** the desired interval has occurred
- **Then** the code is run 

We have considered using Kubernetes native `CronJobs` for this

cmwylie19:
## Description

`OnSchedule` implementation. This PR includes documentation and will include examples as eluded to in #288 .

Creates a shadow store if and only if there is a schedule created to conserve resources.

## Related Issue

Fixes #308 

Relates to #288 

## Type of change

## Checklist before merging

### Expectations

- Schedule store lastTimestamp while store increments count.
- When the `OnSchedule` is not used, there should be no schedulestore


docs/OnSchedule.md
```diff
@@ -0,0 +1,86 @@
+# OnSchedule
+
+The `OnSchedule` feature allows you to schedule and automate the execution of specific code at predefined intervals or schedules. This feature is designed to simplify recurring tasks and can serve as an alternative to traditional CronJobs. This code is designed to be run at the top level on a Capability, not within a function like `When`. 
+  
+> **Note -** To use this feature in dev node `npx pepr dev`, you MUST set `PEPR_WATCH_MODE="true"`. This is because the scheduler only runs in the watch controller not started in dev mode by default. `PEPR_WATCH_MODE="true" npx pepr dev`.
```
btlghrants:
```suggestion
> **Note -** To use this feature in dev mode you MUST set `PEPR_WATCH_MODE="true"`. This is because the scheduler only runs on the watch controller and the watch controller is not started by default in dev mode.

For example: `PEPR_WATCH_MODE="true" npx pepr dev`
```

Typo / flow.

docs/OnSchedule.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Best Practices
+
+`OnSchedule` is designed for targeting intervals equal to or larger than 30 seconds due to the storage mechanism used to archive schedule info, anything less than that is _use at your own risk_.
```
btlghrants:
```suggestion
`OnSchedule` is designed for targeting intervals equal to or larger than 30 seconds due to the storage mechanism used to archive schedule info.
```

Feels like a stronger statement if we just tell people that they shouldn't use <30 sec intervals -- don't want to suggest that people could use shorter intervals _even_ at their own risk, IMO, if we don't want to support that. :thinking:

docs/OnSchedule.md
```diff
@@ -0,0 +1,86 @@
...
+  
+## Usage
+
+Create a schedule by calling the OnSchedule function with the following parameters:
```
btlghrants:
```suggestion
Create a recurring task execution by calling the OnSchedule function with the following parameters:
```

Swapped subject to put a focus on the "thing to be accomplished" (i.e. the scripted task) rather than on the mechanism for accomplishing it (i.e. the schedule).

docs/OnSchedule.md
```diff
@@ -0,0 +1,86 @@
...
+
+**name** - The unique name of the schedule.
+
+**every** - (Optional) An integer that represents the frequency of the schedule.
```
btlghrants:
```suggestion
**every** - (Optional) An integer that represents the frequency of the schedule in number of _units_.
```

My first read of this made me wonder what unit of measure was being used... which is, of course, clarified below, but this edit should help others avoid the same (brief) confusion.

docs/OnSchedule.md
```diff
@@ -0,0 +1,86 @@
...
+
+**unit** - (Optional) A string specifying the time unit for the schedule (e.g., `seconds`, `minute`, `minutes`, `hour`, `hours`).  
+
+**startTime** - (Optional) A UTC timestamp indicating when the schedule should start. All date times must be provided in GMT.  
+
+**run** - A function that contains the code you want to execute on the defined schedule.  
+
+**completions** - (Optional) An integer indicating the maximum number of times the schedule should run.
```
btlghrants:
```suggestion
**completions** - (Optional) An integer indicating the maximum number of times the schedule should run to completion.
```

Minor edit to put focus on how OnSchedule isn't just _starting_ some number of times, it's _completing_ a certain number of runs -- helps clarify why the field is named as it is, IMO.

docs/OnSchedule.md
btlghrants, line 1: I feel like the doc is missing one-or-more notes about how OnSchedule'll be have if the "(optional)" inputs aren't set (i.e. every, unit, startTime, completions).  Not sure if that's better put as an overall note, or as something noted next to each field , like a default-value-if-not-set  note or something.  :thinking:

docs/OnSchedule.md
```diff
@@ -0,0 +1,86 @@
...
+
+
+## Examples
+
+Update the curr ConfigMap every 15 seconds and use the store to track the current count:
```
btlghrants:
```suggestion
Update a ConfigMap every 30 seconds:
```

Cut "curr" since it's a code item & not really representative of the ConfigMap itself.

Update time period to match 1) >30 sec limitation (described above), and 2) the example code (laid out below).

Cut note about "the store" since it's an implementation detail / not really important for the thrust of the example.

docs/OnSchedule.md
```diff
@@ -0,0 +1,86 @@
...
+
+```typescript
+let curr = 0
```
btlghrants: What is this variable for, actually?

docs/OnSchedule.md
```diff
@@ -0,0 +1,86 @@
...
+        });
+
+      } catch (error) {
+        Log.error(error, "Failed to apply ConfigMap using server-side apply.");
+      }
+    },
+  });
+```
+
+Every 24 hours refresh the AWSToken, delays the start for 30 seconds, and only run 3 times:
```
btlghrants:
```suggestion
Refresh an AWSToken every 24 hours, with a delayed start of 30 seconds, running a total of 3 times:
```

Reorg for flow.

docs/OnSchedule.md
```diff
@@ -0,0 +1,86 @@
...
+
+- Simplifies scheduling recurring tasks without the need for complex CronJob configurations.
+- Provides flexibility to define schedules in a human-readable format.
+- Allows you to execute code with precision at specified intervals.
+- Supports limiting the number of schedule completions for finite tasks.
+
+## Considerations
+
+- Ensure that all date times are provided in GMT to maintain consistent scheduling.
+- Review your scheduling needs to determine if OnSchedule is suitable for your use case compared to traditional CronJobs.
```
btlghrants: I'm a little unclear on the intended function of this line -- what is it warning me about?  Is there any sort of statement we could make about why someone would choose to look into using a CronJob over using OnSchedule that might imply the suitability of one over the other?

If not, this line might not be needed..? :thinking:

docs/OnSchedule.md
```diff
@@ -0,0 +1,86 @@
...
+- OnSchedule waits for the store to be ready before executing because the schedule is stored in etcd. This means that if the store is not ready, the schedule will not run.
```
btlghrants:
```suggestion
- OnSchedule is backed by the Pepr Store so it has to wait for the store to be ready before executing.  This means that there will be a (slight) delay between the time Pepr starts up and the time OnSchedule can begin executing scheduled tasks.
```

Etcd is an implementation detail of the Pepr Store & not all that clarifying... unless you're trying to get the reader to look into the details of the Pepr Store (which doesn't really seem to be the case), so I cut it out.

website/content/en/docs/OnSchedule.md
btlghrants, line 1: Same notes on content as in the non-website docs file: docs/OnSchedule.md

btlghrants (commented):
I'm not a huge fan of seeing those few "GitHub Actions / format" suggestions sprinkled through the code... so I wonder if there's a way us to figure out how to squash those?

docs/OnSchedule.md
cmwylie19: not needed, meant to take it out. will remove

docs/OnSchedule.md
cmwylie19: I think it is more of an ADR type doc, due to that I removed it.

docs/OnSchedule.md
cmwylie19: i hear ya, i was hoping the intellisense would cover that but i can make it explicit

cmwylie19: sorry, misinterpreted, so you would like to me to add what happens if the optional values are not set. I can cover that. I think more documentation is generally better.

btlghrants (approved):
LGTM.  Assuming all the tests pass we can merge in the a.m., for sure.  :+1:

PR #384, opened 2023-11-20
fix: updated capabilities to warn the user in the case there could be capability overlap

### LINKED ISSUE #311: multiple builder callbacks that match the same k8s object will only cause the first one to be called.

at `npx pepr dev/build` identify if multiple builder callbacks could match the same object. 

```
When(a.ConfigMap)
  .IsCreated()
  .Mutate(request => {
      // will only be called for Created() requests
  });

When(a.ConfigMap)
  .IsCreatedOrUpdated()
  .Mutate(request => {
  // will only be called for Updated() requests but not Created() requests
  });
```

example 2:
```
// this will always be called, regardless of the data.
When(a.ConfigMap)
  .IsCreatedOrUpdated()
  .Mutate(request => {
      if (request.Raw.data?.["hello"]) {
        // DO SOMETHING
      }
  });

// this will NEVER be called, regardless of the data.
When(a.ConfigMap)
  .IsCreatedOrUpdated()
  .Mutate(request => {
      if (!request.Raw.data?.["hello"]) {
        // DO SOMETHING ELSE
      }
  });

// THIS IS THE CORRECT WAY:
When(a.ConfigMap)
  .IsCreatedOrUpdated()
  .Mutate(request => {
      if (request.Raw.data?.["hello"]) {
        // DO SOMETHING
      } else {
         // DO SOMETHING ELSE
      }
  });
```

if multiple builders match the same object, only the first one will be called. We don't chain them (and we do it silently, which is confusing)

bdw617:
## Description

Pepr capabilities can overlap each other. Right now it creates a possible ambiguous problem where 2 bindings/callbacks can match an object. 

In the following example, the order matters, and might not be obvious to the developer why a configmap with name test in namespace test will be denied. This fix does NOT change the behavior, but will provide a warning (which is WIP) that this could happen. 

```
When(a.ConfigMap)
  .IsCreated()
  .WithName("test")
  .Validate(request => {
    return request.Deny();
  });


When(a.ConfigMap)
  .IsCreated()
  .InNameSpace("test")
  .Validate(request => {
    return request.Approve();
  });
```

The suggested fix would be to merge these two into a single one where the application developer clearly understands the relationship and what they're trying to do. Think of this as linting for pepr capabilities.

## Related Issue

Fixes: https://github.com/defenseunicorns/pepr/issues/311

## Type of change

- [x ] Bug fix (non-breaking change which fixes an issue)

## Checklist before merging


jeff-mccoy (changes requested):
I wouldn't put all these lower-level functions on the class itself and would also make any that had to be on the class private using the `#` syntax

cmwylie19:
@bdw617 as you can see from both the project board and the issue #311 which you referenced this issue is already assigned and a work in progress.. If you want to contribute in the future you should checkout an issue that is unassigned.

cmwylie19:
Closing this for now due to this reason. Feel free to open a new issue if you need to 

https://github.com/defenseunicorns/pepr/issues/311#issuecomment-1832846453

PR #391, opened 2023-11-22
feat!: update `pepr build --entry-point` flag to allow output of deployable module

### LINKED ISSUE #392: feat!: update `pepr build --entry-point` flag to allow output of deployable module
As part of trying to establish a nice e2e test pattern for individual Pepr Capabilities within a Module project, I want to be able to specify both the target output directory (`--output-dir`) _and_ the desired entrypoint .ts file (`--entry-point`) **while also** outputting a fully-packed & deployable Module.

While calling `pepr build` with this combination of the flags is already possible, the (current) `--entry-point` flag implementation allows specification of the entrypoint file to use _but it also_ changes the way the module gets built such that the output module is no longer independently deployable.

This PR changes the way the `--entry-point` flag works to make it output fully a deployable module and instead adds a new flag: `--no-embed` to cover the pre-existing behavior.  This allows use of the flags `--output-dir` and `--entry-point` together to fully specify what-gets-built-and-where while still allowing callers to opt-in to the previous behavior withh the new flag (`--no-embed`).

btlghrants:
## Description
As part of trying to establish a nice e2e test pattern for individual Pepr Capabilities within a Module project, I want to be able to specify both the target output directory (`--output-dir`) and the desired entrypoint .ts file (`--entry-point`) **while also** outputting a fully-packed & deployable Module.

While calling `pepr build` with this combination of the flags is already possible, and while the (current) `--entry-point` flag implementation allows specification of the entrypoint file to use _it also_ changes the way the module gets built such that the output module is no longer independently deployable.

This PR changes the way the `--entry-point` flag works to make it output fully a deployable module and instead adds a new flag: `--no-embed` to cover the pre-existing behavior.  This allows use of the flags `--output-dir` and `--entry-point` together to fully specify what-gets-built-and-where while still allowing callers to opt-in to the previous behavior with the new flag (`--no-embed`).

## Related Issue

Fixes #392 

## Type of change

- [x] Breaking-change-as-part-of-adding-new-feature

## Checklist before merging

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -55,6 +55,9 @@ Create a [zarf.yaml](https://zarf.dev) and K8s manifest for the current module. 
 
 **Options:**
 
-- `-r, --registry-info [<registry>/<username>]` - Registry Info: Image registry and username. Note: You must be signed into the registry
-- `--rbac-mode [admin|scoped]` - Rbac Mode: admin, scoped (default: admin)
 - `-l, --log-level [level]` - Log level: debug, info, warn, error (default: "info")
+- `-e, --entry-point [file]` - Specify the entry point file to build with. (default: "pepr.ts")
+- `-n, --no-embed` - Disables embedding of deployment files into output module. Useful when creating library modules intended solely for reuse/distribution via NPM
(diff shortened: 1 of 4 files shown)


cmwylie19 (approved):
lgtm

PR #398, opened 2023-11-27
fix: watch does not respect ns testing

### LINKED ISSUE #395: Watch() does not respect namespace rules

cmwylie19:
## Description

Watch is not respecting namespace criteria. This PR is to make watch respect the ignoredNamespaces and the capability namespaces. If those two criteria are not respected, a warning is given to the user.

Bindings must respect ignoreNamespace and capability namespace, if not the build will not occur.

The `alwaysIgnore.namespaces` is the first priority, then capability namespaces

## Related Issue

Fixes #395 

Relates to #

## Type of change

## Checklist before merging


src/lib/helpers.ts
```diff
@@ -50,3 +50,69 @@ export async function createDirectoryIfNotExists(path: string) {
...
+
+  // check if binding uses an ignored namespace
+  if (ignoredNamespaceConflict(ignoredNamespaces, bindingNamespaces)) {
+    err += "Binding uses a Pepr ignored namespace.";
+  }
+
+  // ensure filter namespaces are part of capability namespaces
+  if (bindingAndCapabilityNSConflict(bindingNamespaces, capabilityNamespaces)) {
+    err += "Binding uses namespace not governed by capability.";
+  }
```
btlghrants: I was kinda surprised that these error messages don't announce _which namespace/s_ is/are causing the error..?  Is there a reason we're not just outputting what's in conflict?  Seems kinda awkward to tell the user "There's a problem so go find it" instead of just stating what the conflict is.

src/lib/helpers.ts
```diff
@@ -50,3 +50,69 @@ export async function createDirectoryIfNotExists(path: string) {
...
+  const namespaceError = generateWatchNamespaceError(
+    ignoredNamespaces ? ignoredNamespaces : [],
+    bindingNamespaces,
+    capabilityNamespaces ? capabilityNamespaces : [],
+  );
+  if (namespaceError !== "") {
+    console.error(
+      `Error in ${name} capability. A binding violates namespace rules. Please check ignoredNamespaces and capability namespaces: ${namespaceError}`,
+    );
+    process.exit(1);
```
btlghrants: I'm generally not a huge fan of seeing "process.exit()" calls inside helpers / library code / anywhere _outside_ of the CLI code because:
- it doesn't allow the caller an opportunity to correct issues before blowing up the thread,
- it doesn't allow the caller to a way to "do things" (e.g. close child procs, clean up files, and so on) before the process dies, and
- it doesn't allow the caller to modify the exit codes / content or syntax of messages that are output (in the case they have a certain style of log output, etc.).

Generally, things like this are better handled by throwing an error rather than killing the process because it allows the caller the opportunity to act on those points (and probably more).  How hard would it be to push the responsibility to exit up to a higher (preferably CLI-owned) layer and just have this func throw to quit?

btlghrants (commented):
Just a couple Qs.

src/lib/helpers.ts
cmwylie19: I hear you, does it make a different that this is only called from the `cli` in `npx pepr build`?

image

src/lib/helpers.ts
cmwylie19: definitely agree

src/lib/helpers.ts
cmwylie19: the reason it is there is to stop the build of the yaml assets

```typescript
  allYaml = async (rbacMode: string) => {
    this.capabilities = await loadCapabilities(this.path);
    // give error if namespaces are not respected
    for (const capability of this.capabilities) {
      namespaceComplianceValidator(capability, this.alwaysIgnore.namespaces);
    }
    
    // dont want it to proceed

    return allYaml(this, rbacMode);
  };
 ```

cmwylie19: What are your thoughts?

btlghrants: Does it make a difference that it's only called from the CLI..?  Kinda, but kinda not. :thinking: 

I think it's mostly about optionality & not _implicitly_ tying that function to the CLI  / assuming the caller is architecturally-empowered kill the whole execution environtment, more than anything.  Because that function isn't declared in a CLI-specific file / namespace or anything -- its declared in src/lib which is exactly the same place we'd put code that we expected to employ in both CLI and non-CLI contexts, right?  Admittedly, we're not making a great distinction between CLI & non-CLI in our source artifacts, but... from that location, there's not really anything even suggesting to a future caller that they might be importing a function that will explode their process without the chance of recovery.

So, is it a show stopper?  Nah.  But is it good style / a pattern I think we'd want to propagate throughout our code base?  Also probably _not_.  And any code that makes it into main has at least the potential to be copied by new / future devs, so... you know.  Kind of a judgement call.  :shrug: 

Better to just make it exit with a throw and let that propagate up to the top level & kill the process IMO.  At least that way we leave the option for someone in the future to use this call in a way that allows recovery / retry / etc.

btlghrants (approved):
Looks great.

PR #405, opened 2023-11-29
feat: chain validation responses

### LINKED ISSUE #404: Multiple validates in a single capability do not behave as expected
### Environment
Device and OS: macOS
App version: 0.17.1
Kubernetes distro being used: k3d

### Steps to reproduce
1. Create a pepr capability that has two validations for the same resource type (as separate `validate`s)
2. Run a pod/svc/... that should be denied by the first but not by the second.

### Expected result

Pod/svc/... would be denied.

### Actual Result

The object is admitted. Denial still shows in the pepr logs, but does not actually affect the admission state.

I believe the source of the issue is this line - https://github.com/defenseunicorns/pepr/blob/main/src/lib/validate-processor.ts#L52 - the `response.allowed` gets clobbered by each subsuqent validation, so only the final validation triggers an approval/denial.

It would be great to see a fix and enhancement here:
- Ensure that a denial is "saved" and not clobbered
- Build out a list of everything that denied the object, so that all denial conditions are presented to the end user (don't short circuit the validation)

### Severity/Priority

Medium? This does break the ability to properly validate multiple things with separate `When...Validate`s.

### Additional Context

Some examples to test this could be pulled from https://github.com/defenseunicorns/uds-core/blob/3163c6a87b309faa065e5138414fbbe29a3ef17f/src/policies/networking.ts . For example, if you create an externalname type service you would expect it to be denied but it will be admitted, and only nodeport types will be denied.

cmwylie19:
## Description

Create an array of responses in the `validateProcessor` and handle accordingly in the controller by adjusting the `AdmissionReview` object based on whether the admission kind was Mutate or Validate.

If multiple bindings operate on the same kubernetes resource and just one fails, then the object is NOT admitted into the cluster and is denied, BUT the all failed validation responses are chained together and can be seen in the Pepr controller logs and in the api-server response from `kubectl`.

## Related Issue

Fixes #404

Relates to #

## Type of change

## Checklist before merging

## Example

Capability Bindings

```typescript
When(a.Pod)
.IsCreatedOrUpdated()
.Validate(po => {
  if(po.Raw.metadata?.labels["pepr"]){
    return po.Deny("first validation")
  }
  return po.Approve()
})

When(a.Pod)
.IsCreatedOrUpdated()
.Validate(po => {
  if(po.Raw.metadata?.labels["pepr"]){
    return po.Deny("second validation")
  }
  return po.Approve()
})
```

Pepr Controller Logs

```bash
[17:35:37.857] DEBUG (/aadone/92098): Outgoing response
    uid: "0576a189-461a-4fb1-8eee-21c2ea11085c"
    namespace: "default"
    response: [
      {
        "uid": "0576a189-461a-4fb1-8eee-21c2ea11085c",
        "allowed": false,
        "status": {
          "code": 400,
          "message": "first validation"
        }
      },
      {
        "uid": "0576a189-461a-4fb1-8eee-21c2ea11085c",
        "allowed": false,
        "status": {
          "code": 400,
          "message": "second validation"
        }
      }
    ]
```

User Output

```
└─[1] <git:(main✈) > k run aadone --image=nginx -l pepr=done
Error from server: admission webhook "pepr-static-test.pepr.dev" denied the request: first validation; second validation
```

### DIFF
--- a/src/lib/controller/index.ts
+++ b/src/lib/controller/index.ts
@@ -220,7 +220,7 @@ export class Controller {
         Log.debug({ ...reqMetadata, request }, "Incoming request body");
 
         // Process the request
-        let response: MutateResponse | ValidateResponse;
+        let response: MutateResponse | ValidateResponse[];
 
         // Call mutate or validate based on the admission kind
         if (admissionKind === "Mutate") {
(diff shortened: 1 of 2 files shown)


btlghrants (approved):
LGTM -- pair programming FTW.

PR #411, opened 2023-12-04
feat: add pepr monitor command

jeff-mccoy:
## Description

For discussion: adds a simple `npx pepr monitor` command to summarize in-cluster logs. More of an idea to think through.

Screenshot 2023-12-04 at 3 02 23 AM

## Type of change

## Checklist before merging


cmwylie19:
Enjoy this command a lot from a standpoint of consumption of validations. 

I'm seeing some problems with the way we implemented responses as arrays.

Before checking if failures is an array

image


After adding a check 
image


sometimes, `payload.response` is an array which causes the `payload.response.patchType`
image

image


What im wondering is, if it would be better to have every response on a separate line instead of having an array. it seems to add a bit more complexity

jeff-mccoy:
yeah i think the arrays are making things a little messy, I just dont want to drop a bunch of duplicate allows either

cmwylie19:
Feedback from team:

1. Match on a different pattern when iterating for the debug log around responses, other than `Outgoing response`
2. in line 239, the debug statement should print the respones for either the Validate or Mutate from the `kube-apiserver` AdmissionReview response

btlghrants (approved):
LGTM

PR #415, opened 2023-12-04
feat: create a library of helper functions that make working with pepr easier

### LINKED ISSUE #299: Create a library of helper functions that make working with Pepr easier.
 
Could be worth adding it to the `PeprValidateRequest` class as something like `request.getContainers()` or at least as a helper function you can import. 

- **Given** you are in a validate block 
- **When** parameter name is passed into the callback
- **Then** the parameter has a helper function `.getContainers()`

### Additional context

```typescript
// Returns all containers in the pod
export function containers(request: PeprValidateRequest<a.Pod>) {
  return [
    ...(request.Raw.spec?.containers || []),
    ...(request.Raw.spec?.initContainers || []),
    ...(request.Raw.spec?.ephemeralContainers || []),
  ];
}
```

schaeferka:
## Description

Added module-helpers.ts and module-helpers.spec.ts.

module-helpers.ts has a containers function that accepts a request parameter and an optional containerTypes parameter. The request is a PeprValidateRequest<a.Pod> that identifies the pod from which a list of containers should be returned. 

If no containerTypes parameter is included, all containers in the pod will be returned. If "containers" is passed in for the containerTypes, then only spec.containers will be returned. If "initContainers" is passed in for the containerTypes, then only spec.initContainers will be returned. If ephemeralContainers is passed in for the containerTypes, then only spec.ephemeralContainers will be returned.

In order to keep the flexibility of pulling the module-helpers into a separate module in the future, the containers function was put into the new module-helpers.ts file instead of modifying the PeprValidateRequest class.

## Related Issue

Fixes #

Relates to #

## Type of change

## Checklist before merging


cmwylie19 (changes requested):
I like the test, like the function, but this would be even cleaner on the [`PeprValidationRequest`](https://github.com/defenseunicorns/pepr/blob/main/src/lib/validate-request.ts). Overall great work

cmwylie19:
> I like the test, like the function, but this would be even cleaner on the [`PeprValidationRequest`](https://github.com/defenseunicorns/pepr/blob/main/src/lib/validate-request.ts). Overall great work.


sorry missed your comment about flexibility in the PR.. I'm not sure I am following what you mean though. Can you elaborate " keep the flexibility of pulling the module-helpers into a separate module in the future"? Thanks!

This feels very coupled to the `PeprValidateRequest`.

The important element on this is to maintain parody with our Fluent API, `When(a.Pod)....Mutate(pod => pod.SetLabel(""))`. This is a major part of our user experience that is built straight into the intellisense, which makes it easier to write even without having deep knowledge of Pepr

src/lib/module-helpers.ts
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2023-Present The Pepr Authors
+
+import { PeprValidateRequest } from "./validate-request";
+import { a } from "../lib";
+
+// Returns all containers in the pod
+export function containers(
```
cmwylie19, line 8: I think `GetContainers` would align closer to the current helper functions

cmwylie19: similar to SetLabel, HasLabel, SetAnnotation, HasAnnotation

cmwylie19 (changes requested):

jeff-mccoy:
Btw, we ended up needing it in mutate & validate: https://github.com/defenseunicorns/uds-core/blob/common-pepr/src/policies/common.ts#L23

src/lib/module-helpers.ts
schaeferka: In looking at the below-linked uds-core section, would it make sense to update the function to something like:

export function GetContainers(
  request: PeprValidateRequest<a.Pod> | PeprMutateRequest<a.Pod>,
  containerTypes?: "containers" | "initContainers" | "ephemeralContainers",
)

schaeferka: We considered a couple of different ways to access the GetContainers function and it didn't seem especially tied to validation -- seemed more like a selector function (since it just looks up a set of container keys in a K8s.Pod object). By keeping it in a separate module-helpers file, we can also use it for mutate requests as well by updating the request parameter to PeprValidateRequest<a.Pod> | PeprMutateRequest<a.Pod>.

cmwylie19: That is a great point

cmwylie19: since it is not going to be a part of the fluent api in the same way i feel less strongly about the renaming

cmwylie19 (approved):

PR #436, opened 2023-12-12
chore: connect failurePolicy to package.json onError

### LINKED ISSUE #434: ValidatingWebhookConfig should fail on error

Depending on validating webhook configuration, they can fail open when they are unreachable. Because Pepr is doing a security function, we want to fail on error. Ensure `failurePolicy: Fail` is set in `ValidatingWebhookConfiguration`

- **Given** an error evaluating a match condition
- **When** the webhook is never called
- **Then** reject the request

### Additional context
https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-matchconditions

cmwylie19:
## Description

@bburky brought up a potential security thread related to the webhook configurations's `failurePolicy`. Currently, the failurePolicy is always ignore and there is no way to change it, this prove be danger to a ValidatingAdmission Webook or even a Mutating that is supposed to remove capability and securityContexts because it can cause the webhoook to _ignore the object and pass it into the cluster_.

During `npx pepr init` there is a prompt about:

```bash
? How do you want Pepr to handle errors encountered during K8s operations? › - Use arrow-keys. Return to submit.
❯   Ignore - Pepr will continue processing and generate an entry in the Pepr Controller log.
    Log an audit event
    Reject the operation
```

Based on this response, the package.json is populated with `onError` section:

```json
 "pepr": {
    "name": "pepr-test-module",
    "uuid": "static-test",
    "onError": "reject", // <----- HERE
    "alwaysIgnore": {
      "namespaces": [],
      "labels": []
    },
    "includedFiles": []
  },
```

If `reject` is present on the onError section then the failurePolicy generated during a `npx pepr build` will be `Fail`.

## Related Issue

Fixes #434 

Relates to #

## Type of change

## Checklist before merging

### DIFF
--- a/src/lib/assets/webhooks.ts
+++ b/src/lib/assets/webhooks.ts
@@ -129,7 +129,7 @@ export async function webhookConfig(
         name: `${name}.pepr.dev`,
         admissionReviewVersions: ["v1", "v1beta1"],
         clientConfig,
-        failurePolicy: "Ignore",
+        failurePolicy: config.onError === "reject" ? "Fail" : "Ignore" ,
         matchPolicy: "Equivalent",
         timeoutSeconds,
         namespaceSelector: {
--- a/src/lib/assets/webhooks.ts
+++ b/src/lib/assets/webhooks.ts
@@ -129,7 +129,7 @@ export async function webhookConfig(
         name: `${name}.pepr.dev`,
         admissionReviewVersions: ["v1", "v1beta1"],
         clientConfig,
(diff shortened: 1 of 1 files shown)


bburky:
This looks good. The only other note is the uptime and stability of a webhook is critical when it is `failurePolicy: Fail`. I think you already said the webhook pods of Pepr are HA though, so I think you're fine.

jeff-mccoy (approved):

PR #439, opened 2023-12-14
feat: add controller env variables from package.json Pepr config

jeff-mccoy:
## Description

This allows user to specify custom env variables in the `package.json` Pepr Config that will be injected into the controller env variables. The immediate use for [UDS Core](https://github.com/defenseunicorns/uds-core) will be in combination with Zarf variables to allow for a sort of simple global config without needing to watch or get another K8s resource and deal with race conditions.

## Type of change

## Checklist before merging


jeff-mccoy:
Added an E2E test to assert the values were passed to the deployment spec.

Screenshot 2023-12-14 at 3 24 33 AM
Screenshot 2023-12-14 at 3 24 17 AM

cmwylie19 (approved):

PR #445, opened 2023-12-18
feat: add sane defaults to the pod and containers in pepr deployments

### LINKED ISSUE #444: Include securityContext for generated resources
Pepr currently does not include a `securityContext` for generated deployments.   In clusters that run Kyverno (like DUBBD/BigBang), Pepr is in violation with enforced policies.  To ensure we're making it easy for folks to be secure and compliant out of the box it would be great if we could take a stab at adding a default `securityContext` for generated resources.

- **Given** A Pepr Module,
- **When** Module is build with `npx pepr build`
- **Then** Pepr would include appropriate `securityContext` including  `runAsUser` and `runAsGroup` to ensure compliance with Kyverno policies and smooth deployment on platforms like BigBang.

We could also potentially make this configurable if that was desirable.

### Additional context
As of https://repo1.dso.mil/big-bang/bigbang/-/merge_requests/3453 (included in BigBang 2.16, DUBBD 0.15.0) Pepr is getting denied  without adding explicit exclusions. 

This is not an issue in environments that run UDS-Core; at the moment `pepr-system` is excluded from the Pepr based policy validation.

cmwylie19:
## Description

Adds sane defaults to `securityContext` in Pepr's Admission and Watcher Deployment in the podSpec and containerSpec.

## Related Issue

Fixes #444 

Relates to #

## Type of change

## Checklist before merging


src/lib/assets/pods.ts
```diff
@@ -70,6 +70,12 @@ export function watcher(assets: Assets, hash: string) {
         },
         spec: {
           serviceAccountName: name,
+          securityContext: {
+            runAsUser: 1000,
```
rjferguson21: From what I can tell the image pepr uses currently uses a UID and GID of `65532` https://github.com/chainguard-images/images/tree/main/images/node#compatibility-notes

cmwylie19: TY i will fix that in the PR

cmwylie19: TY i will fix that in the PR

rjferguson21:
Testing this against DUBBD 0.15.0 looks good.  All policies that are currently set to enforce are no longer showing failures.  

There is one remaining policy that is currently failing but it is currently only in `audit` mode:
```
(⎈|k3d-uds-test-cluster:default) ~ kubectl get policyreports -o json -n pepr-system | jq '.items[] | .results[] | select(.result == "fail") | {policy: .policy, rule: .rule, message: .message, resources: .resources }'
{
  "policy": "disallow-auto-mount-service-account-token",
  "rule": "automount-service-accounts",
  "message": "validation error: Automount Kubernetes API Credentials isn't turned off. The field automountServiceAccountToken  must be set to false. rule automount-service-accounts failed at path /automountServiceAccountToken/",
  "resources": [
    {
      "apiVersion": "v1",
      "kind": "ServiceAccount",
      "name": "pepr-static-test",
      "namespace": "pepr-system",
      "uid": "9b5bf683-8955-49a9-b34a-8972a4546675"
    }
  ]
}
```

The policy is roughly based on this - https://kyverno.io/policies/other/disable-automountserviceaccounttoken/disable-automountserviceaccounttoken/

Looking into what is a reasonable way to address that I found this example: https://cert-manager.io/docs/installation/best-practice/#restrict-auto-mount-of-service-account-tokens

I would be fine to move this along and create a follow on issue to address that particular policy because it seems a bit more involved.

cmwylie19:
totally got tripped up testing locally b/c I was not able to create the `configMap` with SSA... turns out the code was right, but I had assigned the role `create` for the SSA, which of course didn't work

rjferguson21:
Retested this against a baseline DUBBD `0.15.0` after @cmwylie19 added fixes specifically for `disallow-auto-mount-service-account-token` and there are no longer any Kyverno policy violations. 

LGTM

rjferguson21 (dismissed):

jeff-mccoy (changes requested):
let's discuss this after the break please

RothAndrew:
Status update? Looking forward to this fix so I can stop excluding the pepr-system namespace from my deployment's kyverno policies

jeff-mccoy:
If we break this up, I don't have a problem with security context at all. I have an issue with the automount portion. Is that possible @cmwylie19?

cmwylie19:
> If we break this up, I don't have a problem with security context at all. I have an issue with the automount portion. Is that possible @cmwylie19?

Yes, definitely possible. I think that the automount is taken care of by intentionally and purposefully assigning serviceAccounts to your resources and not using the defaults.

@RothAndrew should be done today and I will look to get it reviewed tomorrow

cmwylie19:
Tested upgrade process manually details here -- https://github.com/defenseunicorns/pepr/issues/464

jeff-mccoy (approved):
sgtm

PR #483, opened 2024-01-12
feat: add timestamp to podSpec annotations force rollout on deployments

### LINKED ISSUE #477: Upgrading a pepr capability with no changes puts webhooks in nonfunctional state
### Environment
Device and OS: macOS
App version: 0.21.1
Kubernetes distro being used: k3d

### Steps to reproduce
1. `npx pepr init`
2. `npx pepr build`
3. `zarf dev deploy dist`
4. `kubectl apply -f capabilities/hello-pepr.samples.yaml` (everything works as expected)
5. `npx pepr build` (no code changes, but api token/certs are regenerated here)
6. `zarf dev deploy dist`
7. `kubectl apply -f capabilities/hello-pepr.samples.yaml`
8. Note that the evil configmap was allowed and nothing was mutated. 

### Expected result

On an upgrade with no changes pepr should continue to behave as expected without any "interruption" in service.

### Actual Result

On upgrade with no changes pepr ends up in a nonfunctional state.

My suspicion: As expected pepr pods do not cycle since no changes are made to the pod spec. However the API token does update in both the secret and the webhook configuration. While the secret value in the running pod (`/app/api-token/value`) is [eventually consistent](https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-files-from-a-pod) and updates to the correct value, I suspect that the actual node process does not reload this value ever. In order to get pepr back in a working state I have to restart the pods. It's also possible that the webhook cert changes could also be problematic for a similar reason although the token seems more likely?

None, steps above for reproduction are more useful in this case 😄 

### Severity/Priority

Mid. Restarting pods is a workaround, but it is an annoying behavior. Also not clear what types of changes would actually modify the pod spec/cause a reload of the token, so this might be more than just a "I redeployed without changes" issue.

### Additional Context

For context on why someone might redeploy the exact same package - this makes more sense in the context of a UDS bundle. The bundle (and all contained components) may be rebuilt together to publish a new bundle version. Even if pepr capabilities don't change other portions of the bundle may change. When trying to upgrade this bundle you would encounter the mentioned bug.

Not sure on a good solution here as I'm not super familiar with how/where the token is handled (assuming that is the problem). If there is a way to "hot reload" the token in the running process that would (for the most part) handle this situation, although you would be relying on the eventual consistency of k8s to update the secret in the running pods. Alternatively any re-deploy could trigger the pods cycle through a checksum annotation or similar for the token value which might be excessive but would be a reasonably simple way to approach it.

cmwylie19:
## Description

Consecutive builds of `npx pepr build` or upgrading versions and rebuilding rotates tls credentials and the module code (secrets) but does not guarantee that the deployment has any updates that will force rollout.

This PR adds an timestamp annotation to the Deployment's podSpec to force a new rollout on each apply in the cluster after a build. Thanks @mjnagel 

## Related Issue

Fixes #477 

Relates to #

## Type of change

## Checklist before merging

### DIFF
--- a/src/lib/assets/pods.ts
+++ b/src/lib/assets/pods.ts
@@ -63,6 +63,9 @@ export function watcher(assets: Assets, hash: string) {
       },
       template: {
         metadata: {
+          annotations: {
+            buildTimestamp: `${Date.now()}`,
+          },
           labels: {
             app,
             "pepr.dev/controller": "watcher",
@@ -180,6 +183,9 @@ export function deployment(assets: Assets, hash: string): kind.Deployment {
       },
       template: {
         metadata: {
+          annotations: {
+            buildTimestamp: `${Date.now()}`,
+          },
           labels: {
             app,
             "pepr.dev/controller": "admission",


btlghrants (approved):
As we talked it through, LGTM.

PR #493, opened 2024-01-18
chore: add test for npx pepr monitor 

### LINKED ISSUE #465: e2e test for `npx pepr monitor`
We currently don't have a test around `npx pepr monitor`

 related to #464 #368

### Additional context

cmwylie19:
## Description

journey test for `npx pepr deploy <uuid>` 
## Related Issue

Fixes #465 

Relates to #

## Type of change

## Checklist before merging

### DIFF
--- a/journey/pepr-deploy.ts
+++ b/journey/pepr-deploy.ts
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: 2023-Present The Pepr Authors
 
-import { expect, it } from "@jest/globals";
+import { describe, expect, it } from "@jest/globals";
 import { execSync, spawnSync, spawn } from "child_process";
 import { K8s, kind } from "kubernetes-fluent-client";
 import { resolve } from "path";
@@ -16,7 +16,7 @@ import {
   waitForPeprStoreKey,
   waitForSecret,
 } from "./k8s";
-import { describe } from "node:test";
+
 
 
 export function peprDeploy() {
--- a/journey/pepr-deploy.ts
+++ b/journey/pepr-deploy.ts
@@ -239,33 +239,3 @@ function testStore() {
     expect(key).toBe("This data was stored by a Watch Action.");
   });
 }
-
-// Get the output from the monitor command
-async function getMonitorData(childProcess: ChildProcess): Promise<string> {
-  return new Promise((resolve, reject) => {
-    let outputData = '';
-    if (childProcess.stdout) {
-      childProcess.stdout.on('data', (data) => {
-        outputData += data.toString();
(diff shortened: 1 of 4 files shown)


journey/pepr-deploy.ts
```diff
@@ -36,6 +38,32 @@ export function peprDeploy() {
...
+      childProcess.removeAllListeners();
+      done()
+    });
+
+    it("npx pepr monitor should display results to console", async () => {
+      const output: string = await getMonitorData(childProcess)
+      const shouldPass = output.includes("✅") || output.includes("❌");
+      expect(shouldPass).toBe(true);
+    });
+  });
```
btlghrants:
```suggestion
  describe("should monitor the cluster for admission changes", () => {

    const until = (predicate: () => boolean ) : Promise<void> => {
      const poll = (resolve: () => void) => {
        if ( predicate() ) { resolve() }
        else { setTimeout(_ => poll(resolve), 250) }
      }
      return new Promise(poll);
    }

    it("npx pepr monitor should display validation results to console", async () => {
      const cmd = [ 'pepr', 'monitor', 'static-test' ]

      const proc = spawn('npx', cmd, { shell: true })

      const state = { accept: false, reject: false, done: false}
      proc.stdout.on('data', (data) => {
        const stdout: String = data.toString()
        state.accept = stdout.includes("✅") ? true : state.accept
        state.reject = stdout.includes("❌") ? true : state.reject

        if (state.accept && state.reject) {
          proc.kill()
          proc.stdin.destroy()
          proc.stdout.destroy()
          proc.stderr.destroy()
        }
      })
      proc.on('exit', () => state.done = true );

      await until(() => state.done)

      // completes only if conditions are met, so... getting here means success!
    }, 5000);
  });

```

btlghrants: Something like this 1) kills cleanly, 2) doesn't have leave a hanging stdio pipe for Jest to cleanup, and 3) reads a bit easier (to me, at least) because it avoids the jest before/after lifecycle hooks for a single wait-until-seen test case.

Lot's of judgement-call type stuff in there, so happy to talk through any of it.

journey/pepr-deploy.ts
```diff
@@ -198,3 +227,33 @@ function testStore() {
...
+      resolve(outputData)
+    });
+
+    childProcess.on('close', () => {
+      resolve(outputData)
+    });
+
+  });
+}
+
```
btlghrants:
```suggestion
```

Not really needed if using suggestion above.  :man_shrugging:

btlghrants (commented):
Got rid of that jest-still-hanging-on, pipe ref thing.  Happy to talk through any of it.

btlghrants (approved):
Seems legit.

PR #547, opened 2024-02-02
feat: removeItemAndWait

### LINKED ISSUE #545: remoteItemAndWait

We need removeItemAndWait in the store for some operator work.

- **Given** removeItemAndWait is issued
- **When** removeItemAndWait is used
- **Then** there is a wait until the item is removed from the store

cmwylie19:
## Description

Needed for operator work in docs and tutorial

## Related Issue

Fixes #545 

Relates to #

## Type of change

## Checklist before merging

### DIFF
--- a/src/lib/storage.ts
+++ b/src/lib/storage.ts
@@ -138,7 +138,7 @@ export class Storage implements PeprStore {
     this.#dispatchUpdate("remove", [key]);
     return new Promise<void>((resolve, reject) => {
       const unsubscribe = this.subscribe(data => {
-        if (!data.hasOwnProperty(key)) {
+        if (!Object.hasOwn(data, key)) {
           unsubscribe();
           resolve();
         }
--- a/src/lib/storage.ts
+++ b/src/lib/storage.ts
@@ -47,6 +47,12 @@ export interface PeprStore {
    * Resolves when the key/value show up in the store.
    */
   setItemAndWait(key: string, value: string): Promise<void>;
+
+  /**
+   * Remove the value of the key.
+   * Resolves when the key does not show up in the store.
+   */
+  removeItemAndWait(key: string): Promise<void>;
 }
 
 /**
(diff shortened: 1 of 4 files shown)


src/lib/storage.ts
```diff
@@ -121,6 +127,31 @@ export class Storage implements PeprStore {
...
+   * the key is removed from the store.
+   *
+   * @param key - The key to add into the store
+   * @returns
+   */
+  removeItemAndWait = (key: string) => {
+    this.#dispatchUpdate("remove", [key]);
+    return new Promise<void>((resolve, reject) => {
+      const unsubscribe = this.subscribe(data => {
+        if (!data[key]) {
```
btlghrants:
```suggestion
        if (!data.hasOwnProperty(key)) {
```

Something like this would probably be better because `!data[key]` would return true anytime the content was falsy (0, false, null, undefined, etc.).

Using `.hasOwnProperty()` makes sure that there _is no property_ for the key that was removed and so won't get tripped up if someone tries to actually store a falsy value.

btlghrants (changes requested):

btlghrants (approved):
Neat.

src/lib/storage.ts
```diff
@@ -121,6 +127,31 @@ export class Storage implements PeprStore {
...
+   * the key is removed from the store.
+   *
+   * @param key - The key to add into the store
+   * @returns
+   */
+  removeItemAndWait = (key: string) => {
+    this.#dispatchUpdate("remove", [key]);
+    return new Promise<void>((resolve, reject) => {
+      const unsubscribe = this.subscribe(data => {
+        if (!data.hasOwnProperty(key)) {
```
btlghrants:
```suggestion
        if (!Object.hasOwn(data, key)) {
```

Maybe ESLint will like this better..?
- https://stackoverflow.com/a/72895649
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn

btlghrants (changes requested):
How 'bout now?

btlghrants (approved):
Hopefully now.

PR #603, opened 2024-02-27
feat: generate helm chart

### LINKED ISSUE #588: Generate a helm chart during `npx pepr build`.

@jeff-mccoy brought up the need to generate a helm chart during an `npx pepr build` to expose common values as in labels/annotations/resources/etc

- **Given** a user builds an pepr module
- **When** the dist folder is built
- **Then** there is also a helm chart

cmwylie19:
## Description

helm chart generation

## Related Issue

Fixes #588 

Relates to #

## Type of change

## Checklist before merging


src/cli/build.ts
```diff
@@ -139,6 +139,8 @@ export default function (program: RootCmd) {
       await fs.writeFile(yamlPath, yaml);
       await fs.writeFile(zarfPath, zarf);
 
+      await assets.generateHelmChart(outputDir);
```
btlghrants, line 142:
```js
const yamlFile = `pepr-module-${uuid}.yaml`;
const yamlPath = resolve(outputDir, yamlFile);
const yaml = await assets.allYaml(opts.rbacMode); // <-- generates string-ish raw manifest content

const zarfPath = resolve(outputDir, "zarf.yaml");
const zarf = assets.zarfYaml(yamlFile);           // <-- generates string-ish zarf manifest content

await fs.writeFile(yamlPath, yaml);               // <-- writes raw manifest to disk
await fs.writeFile(zarfPath, zarf);               // <-- writes zarf manifest to disk

await assets.generateHelmChart(outputDir);        // <-- generates AND writes helm chart to disk
```

Not necessarily something wrong, but was wondering if there was reasoning behind having the gen & write collapsed into a single line only for output of the new Helm chart? :thinking:

src/lib/assets/helm.ts
```diff
@@ -0,0 +1,200 @@
...
+      # Versions are expected to follow Semantic Versioning (https://semver.org/)
+      version: 0.1.0
+  
+      # This is the version number of the application being deployed. This version number should be
+      # incremented each time you make changes to the application. Versions are not expected to
+      # follow Semantic Versioning. They should reflect the version the application is using.
+      # It is recommended to use it with quotes.
+      appVersion: "1.16.0"
+  
+    `;
```
btlghrants:
```suggestion
  return `
    apiVersion: v2
    name: ${name}
    description: ${description || ""}

    # A chart can be either an 'application' or a 'library' chart.
    #
    # Application charts are a collection of templates that can be packaged into versioned archives
    # to be deployed.
    #
    # Library charts provide useful utilities or functions for the chart developer. They're included as
    # a dependency of application charts to inject those utilities and functions into the rendering
    # pipeline. Library charts do not define any templates and therefore cannot be deployed.
    type: application

    # This is the chart version. This version number should be incremented each time you make changes
    # to the chart and its templates, including the app version.
    # Versions are expected to follow Semantic Versioning (https://semver.org/)
    version: 0.1.0

    # This is the version number of the application being deployed. This version number should be
    # incremented each time you make changes to the application. Versions are not expected to
    # follow Semantic Versioning. They should reflect the version the application is using.
    # It is recommended to use it with quotes.
    appVersion: "1.16.0"
`;
```
One level too much indent & spaces leading empty lines.

src/lib/assets/index.ts
```diff
@@ -53,4 +68,108 @@ export class Assets {
...
+      await createDirectoryIfNotExists(`${CHART_DIR}/charts`);
+
+      // create templates dir
+      await createDirectoryIfNotExists(`${CHAR_TEMPLATES_DIR}`);
+
+      // create values file
+      await overridesFile(this, valuesPath);
+
+      // create the chart.yaml
+      await fs.writeFile(chartPath, chartYaml(this.config.uuid, this.config.description || ""));
```
btlghrants:
```suggestion
      await fs.writeFile(chartPath, dedent(chartYaml(this.config.uuid, this.config.description || "")));
```
Chart.yaml gen'd with a bunch of leading whitespace when I ran this block -- just looks like a forgotten dedent() call.

btlghrants: Or, actually... is there are reason you don't want to "dedent" the chartYaml return string _before_ you return it -- like, move the dendent call inside of chartYaml()?  Same thing with the rest of the methods that use dedent() in this file. :thinking:

1) Seems kinda strange to have the yaml generation functions return content that includes arbitrary whitespace, and 2) if these generation functions are used anywhere else, they're _also_ going to have to be deindented.

src/lib/assets/yaml.ts
```diff
@@ -11,6 +11,118 @@ import { deployment, moduleSecret, namespace, watcher } from "./pods";
 import { clusterRole, clusterRoleBinding, serviceAccount, storeRole, storeRoleBinding } from "./rbac";
 import { webhookConfig } from "./webhooks";
 
+// Overrides file generated from assets
```
btlghrants:
```suggestion
// Helm Chart overrides file (values.yaml) generated from assets
```
Embarrassingly, it took me a while to figure out that "overrides" meant "values.yaml" in this context.  Would you mind if we just popped "values.yaml" into the comment somewhere?  It'd have helped me get the point right away.

As a bonus, it'd provide a place for "people who see the output file & wanna search for the source that made it" land in the right place.

btlghrants (changes requested):
Just a couple minor things.

src/lib/assets/yaml.ts
cmwylie19: great idea

src/cli/build.ts
cmwylie19: Hmm that is a good point. There was not a reason for it other than I thought it would be a lot cleaner on the Asset class since it is using several class members there.

cmwylie19: could be two functions for sure though

src/lib/assets/index.ts
cmwylie19: nope, just an oversight. good catch

src/cli/build.ts
btlghrants: Yeah, I'm not too particular about which way is / would be better -- it just stood out to me as "different" & so I wondered why.

btlghrants (approved):
LGTM

PR #619, opened 2024-03-05
chore: added kfc crd to cli

### LINKED ISSUE #527: alias for kfc in pepr cli

`kfc` plays a major role in building operators and generating classes from CRDs. It would be nice to have it in the `npx pepr` cli to make it easier to use from the pepr project without having to worry about version. 

- **Given** you are building an operator
- **When** need to generate a class to handle the crd
- **Then** you can point the `npx pepr kfc crd [src] [dest]`

Part of the problem is that I encountered some unexpected behavior in kfc and i was not able to verify that I was on the latest version. So if i can verify that Pepr is on the latest then I can assume kfc is also up to date

schaeferka:
## Description
Added kfc crd to cli. Allows you to use `npx pepr kfc [source] [dest]` to generate a class from a CRD.
...

## Related Issue

Fixes #527 

Relates to #

## Type of change

## Checklist before merging

- [x ] Test, docs, adr added or updated as needed
- [x ] [Contributor Guide Steps](https://github.com/defenseunicorns/pepr/blob/main/CONTRIBUTING.md#submitting-a-pull-request) followed

### DIFF
--- a/src/cli/kfc.ts
+++ b/src/cli/kfc.ts
@@ -1,0 +1,42 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2023-Present The Pepr Authors
+
+import { execSync } from "child_process";
+import prompt from "prompts";
+
+import { RootCmd } from "./root";
+
+export default function (program: RootCmd) {
+  program
+    .command("kfc crd [source] [dest]")
+    .description("Generate class to handle CRD")
+    .action(async (source, dest) => {
+
+      const { confirm } = await prompt({
+        type: "confirm",
+        name: "confirm",
+        message:
+          "This will overwrite previously generated files in the destination folder.\n" +
+          "Are you sure you want to continue?",
+      });
+
+      // If the user doesn't confirm, exit
+      if (!confirm) {
+        return;
+      }
+
+      console.log("Creating the CRD generated class...");
+
+      try {
+        // Create the CRD generated class
+        execSync(`npx kfc crd ${source} ${dest}`, {
+          stdio: "inherit",
+        });
+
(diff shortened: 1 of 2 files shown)


src/cli/kfc.ts
```diff
@@ -0,0 +1,41 @@
...
+// SPDX-FileCopyrightText: 2023-Present The Pepr Authors
+
+import { execSync } from "child_process";
+import prompt from "prompts";
+
+import { RootCmd } from "./root";
+
+export default function (program: RootCmd) {
+  program
+    .command("kfc crd [source] [dest]")
```
cmwylie19:
```suggestion
    .command("kfc crd <source> <dest>")
```

cmwylie19:
```
You can specify an option which may be used as a boolean option but may optionally take an option-argument (declared with square brackets like --optional [value]).
```

```
The two most used option types are a boolean option, and an option which takes its value from the following argument (declared with angle brackets like --expect <value>). Both are undefined unless specified on command line.
```

We want to ensure this these values are not optional, [source](https://github.com/tj/commander.js?)

src/cli/kfc.ts
```diff
@@ -0,0 +1,41 @@
...
+    .description("Generate class to handle CRD")
+    .action(async (source, dest) => {
```
cmwylie19: This does not work as expected. `src` is crd

image

cmwylie19 (changes requested):
not working

cmwylie19 (approved):

cmwylie19 (approved):

cmwylie19 (approved):

PR #643, opened 2024-03-12
chore: remove watch cache from pepr

### LINKED ISSUE #640: Release pipeline references Zarf
### Environment
Device and OS:
App version:
Kubernetes distro being used:
Other:

### Steps to reproduce
1. Look at release.yaml
```yaml
 - name: "Zarf Agent: Login to GHCR"
```

https://github.com/defenseunicorns/pepr/blob/d75c0e58d6df97d8dced8ef779fcc97fc937c8e0/.github/workflows/release.yml#L31

### Expected result
It should say Pepr Controller instead of Zarf Agent 

### Additional Context

### LINKED ISSUE #642: Propose killing the Pepr Watch caching behavior due
### Overview

Propose killing the Pepr watch cache as it seems to create more problems that it is worth. KFC internally has a `lastSeen` mechanism that manages in the event of network failure or overly stale connections. Pepr seeds this with it's own tracked version of the last seen resources version per callback which is a function that uses `.toString()` which won't capture business logic changes in other areas of the code that function calls. 

Kubernetes recommends idempotency and eventual consistency which means re-calling callback should not result in inconsistent state or failure.

### Steps to reproduce
1.

### Additional Context

cmwylie19:
## Description

Watch Cache is causing more problems that it is fixing. Kubernetes Fluent Client is a way to track resource versions. We are considering removing the watch cache from Pepr 

## Related Issue

Fixes #642 
Fixes #640 (Because the security scans were stuck running for 20 mins and I needed to restart them) 

Relates to #

## Type of change

## Checklist before merging

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -28,7 +28,7 @@ jobs:
           registry-url: "https://registry.npmjs.org"
           cache: "npm"
 
-      - name: "Zarf Agent: Login to GHCR"
+      - name: "Pepr Controller: Login to GHCR"
         uses: docker/login-action@343f7c4344506bcbf9b4de18042ae17996df046d # v3.0.0
         with:
           registry: ghcr.io
--- a/journey/pepr-deploy.ts
+++ b/journey/pepr-deploy.ts
@@ -49,6 +49,8 @@ export function peprDeploy() {
     }
 
     it("npx pepr monitor should display validation results to console", async () => {
+      await testValidate();
+      
       const cmd = ['pepr', 'monitor', 'static-test']
 
       const proc = spawn('npx', cmd, { shell: true })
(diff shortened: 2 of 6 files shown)


journey/pepr-deploy.ts
```diff
@@ -49,6 +49,8 @@ export function peprDeploy() {
     }
 
     it("npx pepr monitor should display validation results to console", async () => {
+      await testValidate();
```
cmwylie19, line 52: Make sure the sample is deployed betfore running the `monitor` command. It is something timing out. `testValidate()` is idempotent and produces the same results every time

.github/workflows/release.yml
```diff
@@ -28,7 +28,7 @@ jobs:
           registry-url: "https://registry.npmjs.org"
           cache: "npm"
 
-      - name: "Zarf Agent: Login to GHCR"
```
cmwylie19, line 31: tech debt and i needed to make the pipeline re-run

btlghrants (approved):
M'kay.

PR #651, opened 2024-03-13
chore: filter bug

cmwylie19:
## Description

In Binding Filters, labels do not have to have values, they only have to have keys. In the event that there is not value, there is an implication that the value is "" (empty string).

This scenario was overlooked in the [checkOverlap](https://github.com/defenseunicorns/pepr/blob/ef780c4e449d17ecae29a792da267cab8dcba3cd/src/lib/helpers.ts#L26) function which checks if the binding's filters/annotations have overlap with the objects.

This was causing an error in UDS-Core as a pod was being ignored that contained the same key as the binding but a different value. 

```ts
export function checkOverlap(record1: Record<string, string>, record2: Record<string, string>) {
  if (Object.keys(record1).length === 0) {
    return true;
  }
  for (const key in record1) {
    if (
      Object.prototype.hasOwnProperty.call(record1, key) &&
      Object.prototype.hasOwnProperty.call(record2, key) &&
      record1[key] === record2[key] // VALUES MUST BE THE SAME
    ) {
      return true;
    }
  }
  return false;
}
```

For Example: 
```ts
When(a.ConfigMap)
  .IsCreated()
  .WithLabel("chuck-norris")
  .Mutate(async change => {
    // Try/catch is not needed as a response object will always be returned
    const response = await fetch<TheChuckNorrisJoke>(
      "https://api.chucknorris.io/jokes/random?category=dev",
    );
```
image

image

## Related Issue

Fixes #

Relates to #

## Type of change

## Checklist before merging

### DIFF
--- a/src/lib/helpers.test.ts
+++ b/src/lib/helpers.test.ts
@@ -944,26 +944,46 @@ describe("replaceString", () => {
 });
 
 describe("checkOverlap", () => {
-  test("returns true if binding annotation/label does not have value and object value is different", () => {
+  test("should return false since all binding annotations/labels do not exist on the object", () => {
+    expect(checkOverlap({ key1: "", key2: "" }, { key1: "something" })).toBe(false);
+  });
+  test("should return false since all binding annotations/labels values do not match on the object values", () => {
+    expect(checkOverlap({ key1: "key1", key2: "key2" }, { key1: "value1", key2: "key2" })).toBe(false);
+  });
+  test("should return true since all binding annotations/labels keys and values match the object keys and values", () => {
+    expect(checkOverlap({ key1: "key1", key2: "key2" }, { key1: "key1", key2: "key2" })).toBe(true);
+  });
+
+  test("should return true since all binding annotations/labels keys exist on the object", () => {
+    expect(checkOverlap({ key1: "", key2: "" }, { key1: "key1", key2: "key2" })).toBe(true);
+  });
+
+  test("(Mixed) should return true since key and key value match on object", () => {
+    expect(checkOverlap({ key1: "one", key2: "" }, { key1: "one", key2: "something" })).toBe(true);
+  });
+  test("(Mixed) should return false since key1 value is differnet on object", () => {
+    expect(checkOverlap({ key1: "one", key2: "" }, { key1: "different", key2: "" })).toBe(false);
+  });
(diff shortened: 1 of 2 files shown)


cmwylie19:
Need to accommodate for AND, when both labels should match. This functionality currently does not 
```ts
When(a.Pod)
  .IsUpdated()
  .WithLabel("batch.kubernetes.io/job-name")
  .WithLabel("service.istio.io/canonical-name")
  ```

File: src/lib/helpers.test.ts [hunk 1]
```diff
@@ -944,23 +944,46 @@ describe("replaceString", () => {
...
+  });
+
+  test("should return true since all binding annotations/labels keys exist on the object", () => {
+    expect(checkOverlap({ key1: "", key2: "" }, { key1: "key1", key2: "key2" })).toBe(true);
+  });
+
+  test("(Mixed) should return true since key and key value match on object", () => {
+    expect(checkOverlap({ key1: "one", key2: "" }, { key1: "one", key2: "something" })).toBe(true);
+  });
+  test("(Mixed) should return false since key1 value is differnet on object", () => {
```
btlghrants:
```suggestion
  test("(Mixed) should return false since key1 value is different on object", () => {
```

src/lib/helpers.test.ts
```diff
@@ -944,23 +944,46 @@ describe("replaceString", () => {
...
+    expect(checkOverlap({ key1: "one", key2: "" }, { key1: "different", key2: "" })).toBe(false);
+  });
+  test("should return true since binding annotation/label since object contains binding key", () => {
+    expect(checkOverlap({ key1: "" }, { key1: "value1" })).toBe(true);
+  });
```
btlghrants:
```suggestion
```
Still need it?  Seems like it's just a lesser form of the test [a few lines higher up](https://github.com/defenseunicorns/pepr/pull/651/files#diff-f1e16476bc27e2629bbaf4e6097a2fbed45c908424b6d585b11c6204c6664695R957):
```
  test("should return true since all binding annotations/labels keys exist on the object", () => {
    expect(checkOverlap({ key1: "", key2: "" }, { key1: "key1", key2: "key2" })).toBe(true);
  });
```

btlghrants (dismissed):
Seems pretty legit.

btlghrants (approved):
K!

PR #676, opened 2024-03-21
chore: named callbacks ADR

### LINKED ISSUE #675: named callbacks ADR

you have a system of callbacks that produce logs. The callbacks are anonymous, so debugging which log comes from which callback can be confusing unless you manually write a name in the log message from within the callback. This consideration needs an ADR

- **Given** we want to look why we decided or decided not to create named callbacks
- **When** we look in the adrs
- **Then** we find the reason

cmwylie19:
## Description

Pepr is a system of callbacks that produce logs. The callbacks are anonymous, so debugging which log comes from which callback can be confusing unless you manually write a name in the log message from within the callback. This consideration needs an ADR. 

## Related Issue

Fixes #675 

Relates to #

## Type of change

## Checklist before merging


jeff-mccoy:
Could you show a compete example of what this would look like?

adr/0008-named-callbacks.md
```diff
@@ -0,0 +1,55 @@
+# 7: LogAs Function in Capability Class "Named Callbacks"
```
btlghrants: "LogAs" seems kinda limited to me. :thinking: 

We talked about this being a general purpose feature that could / would (essentially) carry a label for any given callback around throughout the Pepr system, right?  As it stands, the only place we're currently showing callback function names is in the log messages but that's not all we think can / will be done with this IMO, and I'd like to try to avoid handicapping ourselves by giving it an overly limiting name.

What would you say to naming this something that hinted at a more general usage?  Something like:
- Alias()
- Designation()
- Handle()
- or maybe simply, As()

All of those give a good sense of what we want this new func to be thought of as _without_ necessarily limiting it to logs.

adr/0008-named-callbacks.md
```diff
@@ -0,0 +1,55 @@
...
+
+Date: 2024-03-21
+
+## Status
+
+Proposed
+
+## Context
+
+Pepr utilizes anonymous callbacks that produce logs during execution. This approach, while functional, has led to challenges in debugging and testing. The logs generated by these callbacks lack clear identifiers, making it difficult to ascertain which callback produced which log unless manual names are embedded in the log messages from within the callbacks themselves.
```
btlghrants:
```suggestion
Currently, Pepr utilizes function callback names to produce logs during execution. This approach -- when configured via nameless arrow function expression -- leads to challenges in debugging and testing. The logs generated by these callbacks lack clear, module author defined identifiers, making it difficult to ascertain which callback produced which effect unless manual names are embedded into log messages from within the callbacks themselves.
```

adr/0008-named-callbacks.md
```diff
@@ -0,0 +1,55 @@
...
+
+
+## Decision Drivers
+
+The key drivers for this decision include:
+
+- **Devex**: Instead of manual declaration of `const name = "my-action";` in each callback and each log, we can add a name to the callback to make it easier to identify the source of logs. `LogAs("my-action")`
```
btlghrants:
```suggestion
- **Devex**: Instead of manual declaration of `const name = "my-action";` in each callback and each log (which is easy to forget and tedious to do), we can add a name to the callback to make it easier to identify the source of logs. `LogAs("my-action")`
```

adr/0008-named-callbacks.md
```diff
@@ -0,0 +1,55 @@
...
+
+
+## Benefits
+
+- Simpler developer experience by reducing workload on the module author.
+- Improved Logs and Debugging by programmatically identifying the source of logs in a repeatable fashion across modules.
+
+## Consequences
+
+1. Add another optional function to the Capability class, `LogAs`, to add a name to the callback.
```
btlghrants:
```suggestion
1. Add another optional function to the Capability class, `Alias`, to add a name to the callback (and add a default, generated name for Pepr to use when a dev forgets to set one).
```

Or is this too implementation focused?

adr/0008-named-callbacks.md
cmwylie19: Interesting prospect here, not opposed by any means. Where else other than logs do you see this name being used?

cmwylie19: We like Alias because a user who has not read the docs is most likely to interpret this as a name.

- Pepr Sync

adr/0008-named-callbacks.md
cmwylie19: Agreed we should figure out a default name.

adr/0008-named-callbacks.md
cmwylie19:
```suggestion
# 7: Alias Function in Capability Class "Named Callbacks"
```

adr/0008-named-callbacks.md
```diff
@@ -0,0 +1,55 @@
...
+## Context
+
+Currently, Pepr utilizes function callback names to produce logs during execution. This approach -- when configured via nameless arrow function expression -- leads to challenges in debugging and testing. The logs generated by these callbacks lack clear, module author defined identifiers, making it difficult to ascertain which callback produced which effect unless manual names are embedded into log messages from within the callbacks themselves.
+
+
+## Decision Drivers
+
+The key drivers for this decision include:
+
+- **Devex**: Instead of manual declaration of `const name = "my-action";` in each callback and each log (which is easy to forget and tedious to do), we can add a name to the callback to make it easier to identify the source of logs. `LogAs("my-action")`
```
cmwylie19:
```suggestion
- **Devex**: Instead of manual declaration of `const name = "my-action";` in each callback and each log (which is easy to forget and tedious to do), we can add a name to the callback to make it easier to identify the source of logs. `Alias("my-action")`
```

adr/0008-named-callbacks.md
```diff
@@ -0,0 +1,55 @@
...
+- **Devex**: Instead of manual declaration of `const name = "my-action";` in each callback and each log (which is easy to forget and tedious to do), we can add a name to the callback to make it easier to identify the source of logs. `Alias("my-action")`
+- **Debugging Efficiency**: The debugging process in the current system is time-consuming and error-prone due to the anonymous nature of callbacks.  
+- **Code Readability**: Anonymous callbacks can make the code harder to understand, especially for new developers or when returning to old code that has a different form of naming.
+- **Maintenance and Scalability**: As the system grows, the manageability of anonymous callbacks becomes increasingly burdensome to maintain a consistent state of naming across actions.
+- **Error Tracing**: Pinpointing the source of errors or unexpected behavior in the system is complex without clear identification of each callback.
+
+
+## Considered Options
+
+1. **Named Callbacks**: Add an optional name to the callback to make it easier to identify the source of logs. `LogAs("my-action")`.
```
cmwylie19:
```suggestion
1. **Named Callbacks**: Add an optional name to the callback to make it easier to identify the source of logs. `Alias("my-action")`.
```

adr/0008-named-callbacks.md
```diff
@@ -0,0 +1,55 @@
...
+
+```ts
+// Watch for changes to the UDSPackage CRD to enqueue a package for processing
+When(UDSPackage)
+  .IsCreatedOrUpdated()
+  // Advanced CR Validation
+  .Validate(validator)
+  // Enqueue the package for processing
+  .Reconcile(reconciler)
+  .LogAs("UDS Operator Reconciler")
```
cmwylie19:
```suggestion
  .Alias("UDS Operator Reconciler")
```

adr/0008-named-callbacks.md
```diff
@@ -0,0 +1,55 @@
+# 7: Alias Function in Capability Class "Named Callbacks"
+
+Date: 2024-03-21
+
+## Status
+
+Proposed
```
cmwylie19:
```suggestion
Delayed - Revisiting this in the future and recommending that named callbacks be used in the interim. 
```

cmwylie19:
Was chatting with @brandtkeller about a Lula use-case where they need to validate that certain policies exist and it seems like this could be a useful mechanism to tell a compliance tool what a [binding](https://github.com/defenseunicorns/pepr/blob/0ec535efc6547f15fb4a02c35f6c3b8b4e5ecc22/src/lib/helpers.test.ts#L50) is responsible for.

```ts
When(a.Pod)
  .IsCreatedOrUpdated()
  .Validate(po => {
  // reject is running as root, or is privileged, etc
   })
  .Alias("reject:pods:runAsRoot:privileged:runAsGroup<10:allowPrivilegeEscalation")
  ```
  
  going to go ahead and get this prioritized because the Lula team needs it.

**Comment by cmwylie19** [comment]
We need to verify that we are able to inject the name of the `.Alias` into logs of the callback function.

If it were possible it would need to be done with the `Log` object

```ts
const Log = pino({
  transport,
  timestamp: pinoTimeFunction,
});
```

and import that into the [mutate processor](https://github.com/defenseunicorns/pepr/blob/0ec535efc6547f15fb4a02c35f6c3b8b4e5ecc22/src/lib/mutate-processor.ts#L80) and set

`Log.prefix = the bindings alias`

adr/0008-named-callbacks.md
```diff
@@ -0,0 +1,55 @@
+# 7: Alias Function in Capability Class "Named Callbacks"
+
+Date: 2024-03-21
+
+## Status
+
+Delayed - Revisiting this in the future and recommending that named callbacks be used in the interim.
```
schaeferka: Accepted

schaeferka (dismissed):
The new Alias function matches what is described in the ADR

adr/0008-named-callbacks.md
cmwylie19:
```suggestion
Accepted.
```

cmwylie19 (commented):
approved, just changed status to accepted!

PR #698, opened 2024-03-31
fix: refactor process.exit to exception

naveensrinivasan:
## Description

- refactor process.exit to exception

## Related Issue

Fixes #

Relates to #

## Type of change

## Checklist before merging

### DIFF
--- a/src/runtime/controller.test.ts
+++ b/src/runtime/controller.test.ts
@@ -1,0 +1,24 @@
+import { validateHash, ValidationError } from "./controller";
+import { describe, it, expect } from "@jest/globals";
+
+describe("validateHash", () => {
+  it("should throw ValidationError for invalid hash values", () => {
+    // Examples of invalid hashes
+    const invalidHashes = [
+      "", // Empty string
+      "12345", // Too short
+      "zxcvbnmasdfghjklqwertyuiop1234567890zxcvbnmasdfghjklqwertyuio", // Contains invalid character 'z'
+      "123456789012345678901234567890123456789012345678901234567890123", // 63 characters, one short
+    ];
+
+    invalidHashes.forEach(hash => {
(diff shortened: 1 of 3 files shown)


src/runtime/controller.test.ts
```diff
@@ -0,0 +1,24 @@
+import { validateHash, ValidationError } from "./controller";
+import { describe, it, expect } from "@jest/globals";
+
+describe("validateHash", () => {
```
cmwylie19:
```suggestion
import { describe, it, expect, jest, beforeEach, afterEach } from "@jest/globals";

describe("validateHash", () => {
  let originalExit: (code?: number) => never;

  beforeEach(() => {
    originalExit = process.exit;
    process.exit = jest.fn() as unknown as (code?: number) => never;
  });

  afterEach(() => {
    process.exit = originalExit;
  });
```

cmwylie19: thanks for the contribution. The unit test is working properly but it is actually exiting with a code of 1. This will mock the exit and allow the tests to pass

cmwylie19:
Hi @naveensrinivasan, Something really funky is going on.

There is an [startup](https://github.com/defenseunicorns/pepr/blob/2e814d442be3404e558c6e4e6e736632fbf52824/src/runtime/controller.ts#L79) function that is being invoke when you import `validateHash` from the `controller.ts` based on how controller is implemented. 

I suggest moving validateHash to helpers like in #707 and changing the package.json back. You can use that PR (#707)  as your reference and we can close it afterwards.

cmwylie19:
addressed in https://github.com/defenseunicorns/pepr/pull/707 I wanted to test something and did not want to pollute your PR. It was passing locally but not in CI. Thank you for the contribution

PR #716, opened 2024-04-04
fix: switch to node-lts chainguard image

jeff-mccoy:
Chainguard stopped publishing versioned images outside of `latest` last fall and so dependabot never picked up a newer version. Thanks to @eddiezane pointing this out, we were using a very stale image. This switches the upstream image to the equivalent lts version of Node, which will change version in the next couple of months, but matches our current NodeJS compatibility for Pepr.

PR #718, opened 2024-04-04
chore: vulnerability scan

cmwylie19:
## Description

Chainguard stopped publishing versioned images outside of `latest` last fall and so dependabot never picked up a newer version. This led to a stale Pepr Controller image that had vulnerabilities from not being maintained. This step in CI will fail if there are high vulnerabilities in the `pepr:dev` image which is the candidate image for release. If there are vulnerabilities it will trigger our team to research why the vulnerabilities are there, ie checking to ensure we have the latest and correct images.

CC: Thanks @eddiezane @jeff-mccoy for pointing it out 

## Related Issue

Fixes #

Relates to #716 

## Type of change

## Checklist before merging

- [x] [Contributor Guide Steps](https://docs.pepr.dev/main/contribute/contributor-guide/#submitting-a-pull-request) followed

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
     "gen-data-json": "node hack/build-template-data.js",
     "prebuild": "rm -fr dist/* && npm run gen-data-json",
     "build": "tsc && node build.mjs",
+    "build:image": "test:journey:build && docker buildx build --tag pepr:dev .",
     "test": "npm run test:unit && npm run test:journey",
     "test:unit": "npm run gen-data-json && jest src --coverage --detectOpenHandles --coverageDirectory=./coverage",
     "test:journey": "npm run test:journey:k3d && npm run test:journey:build && npm run test:journey:image && npm run test:journey:run",
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "gen-data-json": "node hack/build-template-data.js",
     "prebuild": "rm -fr dist/* && npm run gen-data-json",
     "build": "tsc && node build.mjs",
-    "build:image": "test:journey:build && docker buildx build --tag pepr:dev .",
+    "build:image": "npm run test:journey:build && docker buildx build --tag pepr:dev .",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -16,6 +16,7 @@
     "gen-data-json": "node hack/build-template-data.js",
     "prebuild": "rm -fr dist/* && npm run gen-data-json",
     "build": "tsc && node build.mjs",
+    "build:image": "npm run test:journey:build && docker buildx build --tag pepr:dev .",
```
btlghrants: I kinda don't like having a generic `build:image` command rely on the more-specialized `test:journey:build` script -- it seems a little confused to have a central-to-the-program script rely on a test script to me; feels snarly. :thinking: 

This change also results in a duplication of the docker CLI code for building "the dev image" (here & down in "test:journey:k3d") -- a task there is really only "one right way to do" but 2 separate code blocks performing independently.

The GitHubs PR UI is being weird and not letting me make a suggestion directly below, but what would you think of just updating the "test:journey:k3d" step to something like:

```
    "test:journey:image-build": "docker buildx build --tag pepr:dev .",
    "test:journey:image-import": "k3d image import pepr:dev -c pepr-dev",
    "test:journey:image": "npm run test:journey:image-build && npm run test:journey:image-import",
```

And then we use the split-out "test:journey:image-build" command instead of `build:image`?  It cuts the duplication & still gets us an independently-callable image build script?

btlghrants (changes requested):

package.json
cmwylie19: Okay, looking to follow but may be lost.

so
Step 1 -  you want to move `npm run test:journey:build` to ci?
Step 2 - create a new npm script for `docker buildx build --tag pepr:dev .` with a more detailed name?

cmwylie19: chatted sync and we decided to remove `npm pack` to make more efficient

btlghrants (approved):
After talking through it, I think we got to a good place!

PR #830, opened 2024-05-24
feat: add flag for chart based zarf.yaml and update templated fields in chart deployments

### LINKED ISSUE #822: Generated zarf.yaml reference helm chart rather than manifest

The current `npx pepr build` generates a helm chart + a manifest (from the templated chart?) for deployment of pepr. The `zarf.yaml` that is also generated references that manifest. This limits a lot of possible end user config options (labels, resources, affinity, etc). While an end user could make a new `zarf.yaml` or modify the generated one, it seems preferable that the default would be to use the chart.

When I `npx pepr build`, the `zarf.yaml` that is generated uses the helm chart. This allows me to import this `zarf.yaml` other locations and extend with values files.

As mentioned this could be done with a new `zarf.yaml` referencing the chart.

### Additional context

`uds-cli` heavily leans into helm charts for overriding values in a bundle. This does provide a way that someone could modify certain pepr pieces (for good or bad) so it may require some careful consideration on whether all existing items should still be exposed as helm values.

As an aside, this would also make things like "flavored pepr" closer aligned to the approach we take in `uds-core` with other flavored items, using values to override the image rather than the build flag.

zachariahmiller:
## Description
Updates `npx pepr build` to take in an optional argument `--zarf chart|manifest` so that the generated zarf yaml will use the generated chart vs the manifest. Default behavior is still to use the manifest. [pepr 822](https://github.com/defenseunicorns/pepr/issues/822)

Also adds templating to the pod annotations on the deployments and extra volume/volumeMounts
in support of [uds-core 417](https://github.com/defenseunicorns/uds-core/issues/417)

I tested the generation locally for both the --zarf flag as well as rendering the templates with a values file that added pod annotations and extra volumes. There didn't seems to be any area really covering testing this right now. The existing tests in helm.test.ts do still generically check the templates are valid that were updated here. 

## Related Issue

Fixes #
[822](https://github.com/defenseunicorns/pepr/issues/822)
Relates to #
[417](https://github.com/defenseunicorns/uds-core/issues/417)

## Type of change

## Checklist before merging

- [x] [Contributor Guide Steps](https://docs.pepr.dev/main/contribute/contributor-guide/#submitting-a-pull-request) followed

### DIFF
--- a/docs/030_user-guide/010_pepr-cli.md
+++ b/docs/030_user-guide/010_pepr-cli.md
@@ -89,6 +89,7 @@ Create a [zarf.yaml](https://zarf.dev) and K8s manifest for the current module. 
 - `-i, --custom-image [custom-image]` - Custom Image: Use custom image for Admission and Watcher Deployments.
 - `--registry [GitHub, Iron Bank]` - Container registry: Choose container registry for deployment manifests.
 - `-v, --version <version>. Example: '0.27.3'` - The version of the Pepr image to use in the deployment manifests.
+- `-z, --zarf [manifest|chart]` - The Zarf package type to generate: manifest or chart (default: manifest).
 
 ## `npx pepr kfc`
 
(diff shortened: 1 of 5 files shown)


cmwylie19:
@zachariahmiller this is really great work. Just went through and ran everything locally and added the new flags to the helm chart and everything is working well. Thanks for the PR.

cmwylie19 (approved):

PR #866, opened 2024-06-07
chore: clear cache on failure

### LINKED ISSUE #865: Clear Cache for Pepr Store on Failure
### Environment

If you try to remove an item from the Pepr store that is not there, it will always give a `422` error. 

```ts
      try {
        // Send the patch to the cluster
        await K8s(PeprStore, { namespace, name: this.#name }).Patch(payload);
      } catch (err) {
        Log.error(err, "Pepr store update failure");

        // On failure to update, re-add the operations to the cache to be retried
        for (const idx of indexes) {
          sendCache[idx] = payload[Number(idx)];
        }
      }
```

### Steps to reproduce
```ts
import { Capability, Log, a } from "pepr";

export const HelloPepr = new Capability({
  name: "hello-pepr",
  description: "A simple example capability to show how things work.",
});

const { When, Store } = HelloPepr;

When(a.Pod)
  .IsCreatedOrUpdated()
  .WithLabel("removeItem")
  .Watch(() => {
    Log.info("Pod with timeout label");
    for (let i = 0; i < 20; i++) {
      Store.removeItem(`element-${i}`);
    }
  });


  When(a.Pod)
  .IsCreatedOrUpdated()
  .WithLabel("a")
  .Watch(() => {
    Log.info("Pod with timeout label");
    Store.setItem("a","a")
  });

```

First, create a pod that gets the store's cache out of whack: `k run a --image=nginx -l removeItem=2`

Notice the 422 in the logs

Now try to update the store: `k run b --image=nginx -l a=a`

Notice you still get a 422 bc the cache still contains items that do not exist from the store

### Additional Context

cmwylie19:
## Description

We were seeing behavior where on a `Store.removeItem()`, if the item was not there the .Patch call returns a 422 but not clear the body of the patch. This means that every proceeding patch will also fail. This clears the cache on failure.

## Related Issue

Fixes #865 

Relates to #

## Type of change

## Checklist before merging

### DIFF
--- a/src/lib/controller/store.ts
+++ b/src/lib/controller/store.ts
@@ -167,10 +167,7 @@ export class PeprControllerStore {
       } catch (err) {
         Log.error(err, "Pepr store update failure");
 
-        // On failure to update, re-add the operations to the cache to be retried
-        for (const idx of indexes) {
-          sendCache[idx] = payload[Number(idx)];
-        }
+        Object.keys(sendCache).forEach(key => delete sendCache[key]);
       }
     };
 
--- a/src/lib/controller/store.ts
+++ b/src/lib/controller/store.ts
@@ -167,7 +167,14 @@ export class PeprControllerStore {
       } catch (err) {
         Log.error(err, "Pepr store update failure");
 
-        Object.keys(sendCache).forEach(key => delete sendCache[key]);
+        if (err.status === 422) {
+          Object.keys(sendCache).forEach(key => delete sendCache[key]);
+        } else {
+          // On failure to update, re-add the operations to the cache to be retried
+          for (const idx of indexes) {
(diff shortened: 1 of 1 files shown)


src/lib/controller/store.ts
```diff
@@ -167,10 +167,7 @@ export class PeprControllerStore {
       } catch (err) {
         Log.error(err, "Pepr store update failure");
 
-        // On failure to update, re-add the operations to the cache to be retried
-        for (const idx of indexes) {
-          sendCache[idx] = payload[Number(idx)];
-        }
+        Object.keys(sendCache).forEach(key => delete sendCache[key]);
```
cmwylie19: Scope this down to a 422 error, else keep the same implementation

btlghrants (approved):

PR #904, opened 2024-06-19
chore: add fuzzing and property based testing to Pepr

### LINKED ISSUE #742: Add fuzz testing to Pepr

Pepr needs fuzzing. Add Fuzz testing to Pepr so that we run a battery of tests through our functions. This will also enhance our scorecard score. 

- **Given** want to start adding fuzz tests to pepr to increase confidence 
- **When** we look to start
- **Then** we start on a single function, one of the [helpers](https://github.com/defenseunicorns/pepr/blob/main/src/lib/helpers.ts) would be a good candidate

cmwylie19:
## Description

Today in Pepr testing, we specify the inputs and outputs we want to use and observe.

The problem with our current approach to testing software is that it places the responsibility of identifying bugs solely on the developers writing the tests. They must think about all the things that could go wrong if they want their tests to find all potential bugs. While it's possible to think about common failure cases with enough experience, it's not really feasible to consider all possible edge cases without proving the code line by line.

This PR intends to **start** experimentally adding property based, fuzzing and fake data tests to Pepr.

We think it should also increase our Scorecard score

Shows library injecting random strings
image

## Related Issue

Fixes #742 

Relates to #

## Type of change

## Checklist before merging

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24,6 +24,7 @@
       "devDependencies": {
         "@commitlint/cli": "19.3.0",
         "@commitlint/config-conventional": "19.2.2",
+        "@fast-check/jest": "^1.8.2",
         "@jest/globals": "29.7.0",
         "@types/eslint": "8.56.10",
         "@types/express": "4.17.21",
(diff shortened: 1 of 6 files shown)


cmwylie19:
Current Score: 8.9

btlghrants (approved):
As we covered in standup, LGTM.

PR #918, opened 2024-06-27
feat: sanitizing store keys - store path v2

### LINKED ISSUE #915: Pepr Store Keys Need Sanitization
Certain keys are not making it into the store, the patch operation is returning `422`.

image

We need a way to sanitize the key so that the patch operation succeeds and users of the store can get their data in a consistent manner.

```plaintext
We ran into an issue in a few spots where we accept arbitrary keycloak client ids in our Package CR and then use those in other resources without "sanitizing" them. One of those places was the pepr store - we were setting weird client IDs like https://google.com/ for testing and then our pepr store key had that in it and failed during set. Just want to make sure we know what is allowed for those key names.
```

Reproduce:

```ts
When(a.Pod)
  .IsCreatedOrUpdated()
  .Mutate(po => {
    Log.info("Pod created or updated", po);
    //Store.setItem(`${po.Raw.metadata!.name}`, "a");
    Store.setItem(`https://google.com-${po.Raw.metadata!.name}`, "bcd");
  });
```

Definition of done:
- Folks can put in keys to the store like `https://google.com-${po.Raw.metadata!.name}`, the store sanitizes them before the pack, and they can get them back out by the same key (the store sanitizes the Store.getItem). 

`/src/lib/controller/store.ts`

### Additional context

cmwylie19:
## Description

This PR Migrates the Store to a v2 data path, migrates keys from previous version before a read of the CR, patches them to v2, then starts a watch. It also adds fuzzing, property-based, more unit and journey tests.

image

## Related Issue

Fixes #915 

Relates to PR https://github.com/defenseunicorns/pepr-excellent-examples/pull/46 updating the e2e test to expect the keys base64 encoded

## Type of change

## Checklist before merging


mjnagel (commented):
This definitely seems functional, I do have a few comments on things that might be painful from the core side with this path:
1. In the past I have relied on a `k describe` of the peprstore resource(s) to see what my currently store k/v pairs are. With the keys base64 encoded that's no longer the most transparent experience. I think that's _okay_ since it's not dissimilar from the experience viewing something like a secret, but it may be nice to consider a way to view the store easily (whether that's just a documented kubectl command, k9s plugin, or a new `npx pepr <store-viewer>` type thing?
2. The migration script is definitely usable, and we could probably find a way to make the upgrade completely seamless (i.e. not require any end user to do more than just a `zarf deploy`). I don't know how much effort this would be on this side of things but it would definitely be a more pleasant experience if pepr completely handled the migration for us. When I was playing around with this issue early on I tried some basic migration tactics in our code like when calling `getItem` first check if the old format key existed (non-base64 in this case). If it did, perform a "migration" by deleting and setting the new key, before using the value. Unsure if something similar would fit well in pepr.
3. Related to ^ but not directly tied to these changes, it might be nice to explore adding a `Store.migrate(oldKey, newKey)` type function for end users if they change their code in what keys they use. Might just be a lack of knowledge on my part but in exploring changing the store keys it didn't feel like the most seamless experience.

cmwylie19:
> This definitely seems functional, I do have a few comments on things that might be painful from the core side with this path:
> 
> 1. In the past I have relied on a `k describe` of the peprstore resource(s) to see what my currently store k/v pairs are. With the keys base64 encoded that's no longer the most transparent experience. I think that's _okay_ since it's not dissimilar from the experience viewing something like a secret, but it may be nice to consider a way to view the store easily (whether that's just a documented kubectl command, k9s plugin, or a new `npx pepr <store-viewer>` type thing?
> 2. The migration script is definitely usable, and we could probably find a way to make the upgrade completely seamless (i.e. not require any end user to do more than just a `zarf deploy`). I don't know how much effort this would be on this side of things but it would definitely be a more pleasant experience if pepr completely handled the migration for us. When I was playing around with this issue early on I tried some basic migration tactics in our code like when calling `getItem` first check if the old format key existed (non-base64 in this case). If it did, perform a "migration" by deleting and setting the new key, before using the value. Unsure if something similar would fit well in pepr.
> 3. Related to ^ but not directly tied to these changes, it might be nice to explore adding a `Store.migrate(oldKey, newKey)` type function for end users if they change their code in what keys they use. Might just be a lack of knowledge on my part but in exploring changing the store keys it didn't feel like the most seamless experience.

1. For sure, could not agree more. Maybe we can even do both.

2 & 3 are good ideas too. I just have a lot of open questions around how to guarantee we will get the capability name right without user intervention in the code as it is today because the capability name is tightly coupled to the module. However, i definitely think we can come up with something together to see how we can make this upgrade a seamless experience without the user even knowing it happened. 

Lets chat when you are freed up on current work and we will come up with a game plan. For now I am going to put this PR on old and do more investigation and research.

cmwylie19:
> > This definitely seems functional, I do have a few comments on things that might be painful from the core side with this path:
> > 
> > 1. In the past I have relied on a `k describe` of the peprstore resource(s) to see what my currently store k/v pairs are. With the keys base64 encoded that's no longer the most transparent experience. I think that's _okay_ since it's not dissimilar from the experience viewing something like a secret, but it may be nice to consider a way to view the store easily (whether that's just a documented kubectl command, k9s plugin, or a new `npx pepr <store-viewer>` type thing?
> > 2. The migration script is definitely usable, and we could probably find a way to make the upgrade completely seamless (i.e. not require any end user to do more than just a `zarf deploy`). I don't know how much effort this would be on this side of things but it would definitely be a more pleasant experience if pepr completely handled the migration for us. When I was playing around with this issue early on I tried some basic migration tactics in our code like when calling `getItem` first check if the old format key existed (non-base64 in this case). If it did, perform a "migration" by deleting and setting the new key, before using the value. Unsure if something similar would fit well in pepr.
> > 3. Related to ^ but not directly tied to these changes, it might be nice to explore adding a `Store.migrate(oldKey, newKey)` type function for end users if they change their code in what keys they use. Might just be a lack of knowledge on my part but in exploring changing the store keys it didn't feel like the most seamless experience.
> 
> 1. For sure, could not agree more. Maybe we can even do both.
> 
> 2 & 3 are good ideas too. I just have a lot of open questions around how to guarantee we will get the capability name right without user intervention in the code as it is today because the capability name is tightly coupled to the module. However, i definitely think we can come up with something together to see how we can make this upgrade a seamless experience without the user even knowing it happened.
> 
> Lets chat when you are freed up on current work and we will come up with a game plan. For now I am going to put this PR on old and do more investigation and research.

Now thinking back on this, what if we:
- used Store.onReady Callback `Store.onReady(()=>{})`
- added a Store.keyRaw like feature
- When we map through Store items, check if they are base64Encoded
- if not, migrate and delete the old key

What are you thoughts on that? We could also roll out the store view feature at the same time

journey/entrypoint.test.ts
```diff
@@ -30,3 +30,7 @@ describe("Journey: `npx pepr build`", peprBuild);
 describe("Journey: `npx pepr deploy`", peprDeploy);
 
 describe("Journey: `npx pepr dev`", peprDev);
+
+// describe("Journey: Test Store Migration", peprStoreMigrate);
```
cmwylie19:
```suggestion
```

journey/pepr-deploy.ts
```diff
@@ -260,22 +273,83 @@ function testMutate() {
...
+//   })
+
+//   it("should deploy the Pepr controller into the test cluster", async () => {
+//     execSync("kubectl rollout restart deploy/pepr-static-test -n pepr-system", { cwd, stdio: "inherit" })
+//     await Promise.all([waitForDeploymentReady("pepr-system", "pepr-static-test"), waitForDeploymentReady("pepr-system", "pepr-static-test-watcher")]);
+//   });
+
+//   // This asserts that the keys are v2
+//   describe("should upgrade the PeprStore", testStore);
+// }
```
cmwylie19:
```suggestion
```

journey/entrypoint.test.ts
```diff
@@ -30,3 +30,6 @@ describe("Journey: `npx pepr build`", peprBuild);
 describe("Journey: `npx pepr deploy`", peprDeploy);
 
 describe("Journey: `npx pepr dev`", peprDev);
+
+
+
```
cmwylie19:
```suggestion
```

journey/pepr-deploy.ts
```diff
@@ -164,6 +175,8 @@ function testIgnore() {
     expect(cm.metadata?.labels?.["pepr"]).toBeUndefined();
   });
 }
+
```
cmwylie19, line 178:
```suggestion
```

src/lib/storage.ts
```diff
@@ -173,7 +186,7 @@ export class Storage implements PeprStore {
   #onReady = () => {
     // Notify all ready handlers with a clone of the store
     for (const handler of this.#readyHandlers) {
-      handler(clone(this.#store));
+      handler(clone(this.store));
```
cmwylie19:
```suggestion
      handler(clone(this.#store));
```

src/lib/storage.ts
```diff
@@ -73,32 +83,35 @@ export class Storage implements PeprStore {
 
   receive = (data: DataStore) => {
     Log.debug(data, `Pepr store data received`);
-    this.#store = data || {};
+    this.store = data || {};
```
cmwylie19:
```suggestion
    this.#store = data || {};
```

src/lib/storage.ts
```diff
@@ -80,25 +90,28 @@ export class Storage implements PeprStore {
     // Notify all subscribers
     for (const idx in this.#subscribers) {
       // Send a unique clone of the store to each subscriber
-      this.#subscribers[idx](clone(this.#store));
+      this.#subscribers[idx](clone(this.store));
```
cmwylie19:
```suggestion
      this.#subscribers[idx](clone(this.#store));
```

src/lib/storage.ts
```diff
@@ -60,8 +69,9 @@ export interface PeprStore {
  *
  * The API is similar to the [Storage API](https://developer.mozilla.org/docs/Web/API/Storage)
  */
+
 export class Storage implements PeprStore {
-  #store: DataStore = {};
+  store: DataStore = {};
```
cmwylie19:
```suggestion
  #store: DataStore = {};
```

src/lib/storage.ts
```diff
@@ -85,20 +95,23 @@ export class Storage implements PeprStore {
...
+    const result = this.store[v2StoreKey(key)] || null;
+    if (result !== null && typeof result !== "function" && typeof result !== "object") {
+      return result;
+    }
+    return null;
   };
 
   clear = () => {
-    this.#dispatchUpdate("remove", Object.keys(this.#store));
+    this.#dispatchUpdate("remove", Object.keys(this.store));
```
cmwylie19:
```suggestion
    this.#dispatchUpdate("remove", Object.keys(this.#store));
```

src/lib/storage.ts
```diff
@@ -85,20 +95,23 @@ export class Storage implements PeprStore {
   };
 
   getItem = (key: string) => {
-    // Return null if the value is the empty string
-    return this.#store[key] || null;
+    const result = this.store[v2StoreKey(key)] || null;
```
cmwylie19:
```suggestion
    const result = this.#store[v2StoreKey(key)] || null;
```

btlghrants (dismissed):
Not much left to say after the in-depth, in-person walk-through -- seems to do what it says on the tin. :+1:

cmwylie19:
this runner is bjorked

cmwylie19:
> this runner is bjorked

image

cmwylie19:
The GitHub Runners are failing for node/npm

```bash
Run npm ci
  npm ci
  shell: /usr/bin/bash -e {0}
npm error Exit handler never called!
npm error This is an error with npm itself. Please report this error at:
npm error   <https://github.com/npm/cli/issues>
npm error A complete log of this run can be found in: /home/runner/.npm/_logs/[2](https://github.com/defenseunicorns/pepr/actions/runs/9980001407/job/27580881374#step:4:2)024-07-17T19_05_0[3](https://github.com/defenseunicorns/pepr/actions/runs/9980001407/job/27580881374#step:4:3)_312Z-debug-0.log
```

Relevant Issues:
- https://github.com/nodejs/node/issues/53902

Relevant news:
- https://github.com/nodejs/node/releases/tag/v22.5.0  Just released, thanks @mjnagel 

We need to talk about changing the matrix test from `latest` if this does not get fixed ASAP.

journey/pepr-deploy.ts
```diff
@@ -260,22 +272,57 @@ function testMutate() {
...
+  // Apply the store
+  const appliedStore = spawnSync("kubectl apply -f journey/resources/non-migrated-peprstore.yaml", {
+    shell: true, // Run command in a shell
+    encoding: "utf-8", // Encode result as string
+    cwd: resolve(cwd, ".."),
+  });
+  const { stdout } = appliedStore;
+
+  expect(stdout).toContain("peprstore.pepr.dev/pepr-static-test-store");
+}
```
btlghrants, line 328: Any reason why these shell-out to `kubectl apply` rather than just using KFC to apply the CR?  Seems like it'd be better to avoid using an external bin if we can & you're already using the fluent client elsewhere in the file, no?

Something like we do in the exex tests, [here](https://github.com/defenseunicorns/pepr-excellent-examples/blob/f22fcb756bfe82e2a12587eea0976aa9a262002d/_helpers/src/cluster.e2e.test.ts#L128) maybe?

src/lib/storage.test.ts
```diff
@@ -2,9 +2,91 @@
...
+      fc.property(fc.string(), fc.string(), (key, value) => {
+        storage.setItem(key, value);
+        storage.removeItem(key);
+        expect(storage.getItem(key)).toBeNull();
+      }),
+      { numRuns: 100 },
+    );
+  });
+
+  it("should ensure all set items are base64 encoded internally", () => {
```
btlghrants:
```suggestion
  it("should ensure all set items are v2-coded internally", () => {
```

Fix hold-over from previous impl..?

src/lib/storage.test.ts
```diff
@@ -15,21 +97,22 @@ describe("Storage", () => {
   it("should set an item", () => {
     const mockSender = jest.fn();
     storage.registerSender(mockSender);
-
+    const key = "key1";
     storage.setItem("key1", "value1");
```
btlghrants:
```suggestion
    storage.setItem(key, "value1");
```

Use variable instead of duplicating string.

btlghrants (commented):
Couple minor suggestions & one Q.

journey/pepr-deploy.ts
cmwylie19: you are right, but we should make an issue to take away all the of the `kubectl` applies but as the for journey tests, we have been using it for at least the last 11 months since they were built. Do you think the refactor should happen now?

btlghrants: Mmm, I'll leave it up to you, but since we're already using it elsewhere I'd be fine with just doing a cleanup on it all at once, later.

btlghrants (approved):

PR #933, opened 2024-07-02
chore: adr for store enhancement

### LINKED ISSUE #924: Store Enhancement Planning ADR

Extends existing Store functionality to v2 to enable storing keys like URLs, and provides a seamless hand-off migration path for users  while introducing additional tooling to view the store resource. 

Definition of Done:
- [ ] How to enable the store to set special characters as keys
- [ ] Address how to view the Store CR key/values in a user friendly way
- [ ] Hands off end user migration path is decided on
- [ ] Stakeholder feedback
- [ ] Consider that UDS Core processes `UDSExcemptions` on startup so we want to leave that process unbothered (We think this should probably be fine but need to test. Other solutions pre-starup would involve `initContainer`). If this were a problem i think it could be solved by moving [`startExemptionWatch`](https://github.com/defenseunicorns/uds-core/blob/2e1c53e939b99794c8e6994f20282974bd139917/pepr.ts#L17) to the [`Store.onReady`](https://github.com/defenseunicorns/uds-core/blob/2e1c53e939b99794c8e6994f20282974bd139917/pepr.ts#L38) callback. IMO this checkbox is part of stakeholder feedback.

- **Given** a key stakeholder is concerned around the enhancement PR
- **When** they take a look at the PR
- **Then** their questions are answered

cmwylie19:
## Description

Extends existing Store functionality to v2 to enable storing keys like URLs, and provides a seamless hand-off migration path for users while introducing additional tooling to view the store resource.

- [x] How to enable the store to set special characters as keys
- [x] Address how to view the Store CR key/values in a user friendly way
- [x] Hands off end user migration path is decided on
- [ ] Stakeholder feedback
- [ ] UDS Core processes `UDSExcemptions` on startup so we want to leave that process unbothered (We think this should probably be fine but need to test. Other solutions pre-starup would involve `initContainer`). If this were a problem i think it could be solved by moving [`startExemptionWatch`](https://github.com/defenseunicorns/uds-core/blob/2e1c53e939b99794c8e6994f20282974bd139917/pepr.ts#L17) to the [`Store.onReady`](https://github.com/defenseunicorns/uds-core/blob/2e1c53e939b99794c8e6994f20282974bd139917/pepr.ts#L38) callback. IMO this checkbox is part of stakeholder feedback.

Also, i don't think this Store change will affect the startExemptionWatch function bc they are not using the actual Pepr Store in that instance but populating a map[string]interface{} type of store object.

## Related Issue

Fixes #924 

Relates to #918

## Type of change

## Checklist before merging


adr/00080-store-key-sanitization.md
```diff
@@ -0,0 +1,28 @@
...
+
+Date: 2024-07-02
+
+## Status
+
+Under Review
+
+## Context
+
+Issues were discovered in a few spots when accepting arbitrary keycloak client ids and setting them into the Store without sanitization. IDs like https://google.com/ failed during `Store.setItem` due to the `/` character and returned 422.
```
btlghrants:
```suggestion
Issues were discovered in a few spots when accepting arbitrary keycloak client ids (i.e. URIs) as Store keys without sanitization. IDs like 'https://google.com/' fail during `Store.setItem()` calls due to '/' characters being interpreted as _separators within a JSON data path_ by the underlying implementation (resulting in a thrown error and returning a 422).
```
Just wanted to make it a bit more explicit for readers who weren't privy to our team convos.

adr/00080-store-key-sanitization.md
```diff
@@ -0,0 +1,28 @@
+# 8: Store Key Sanitization
```
btlghrants:
```suggestion
# 8: Store Key Encoding
```
The word "sanitization" usually describes swapping out / cleaning up of something & leaving only the "good stuff" behind, but... this feature isn't really changing what's input so the word "sanitization" seems kinda out of place to me.

For example, [here](https://www.esecurityplanet.com/endpoint/prevent-web-attacks-using-input-sanitization/#:~:text=Sanitizing%20consists%20of%20removing%20any,before%20insertion%20in%20a%20database.): 
"Sanitizing consists of removing any unsafe characters from user inputs, and validating will check to see if the data is in the expected format and type. Sanitizing modifies the input to ensure it's in a valid format for display, or before insertion in a database."

Seems to me that the word "encoding" better describes exactly what we're doing AND also implies some stuff of the downstream consequences (e.g. that the Store will no longer be readable as-is / will need to be decoded).  That said, I'm totally open to other descriptors too -- would just wanna see a word that doesn't imply that we're somehow "changing" the inputs, you know?

adr/00080-store-key-sanitization.md
```diff
@@ -0,0 +1,28 @@
...
+* Update the Store key prefix with v2: `/data/${capabilityName}-v2-${key}`
+* Migrate before calling `this.#onReady()` in the `#receive` function of `src/lib/controller/store.ts`, checking if each key that matches the old prefix, if so, migrating to the new prefix with base64 encoding. 
+* Add a new pepr command for viewing store items. like `npx pepr view-store`, maybe a `kubectl` plugin or alias too.
+* Enhance existing unit testing to cover the new base64 encoding and migration logic. Add a battery of new fuzz and property-based tests.
+
+## Consequences
+
+1. Users cannot easily read the PeprStore CR
+2. We need to migrate existing PeprStore CRs so that the `Store.getItem/Store.removeItem` will not fail due to the base64 encoding.
+3. Will slightly slow down the store startup since we need to check and possibly migrate store before onReady.
```
btlghrants:
```suggestion
3. Will slightly slow down the store startup since we need to check and possibly migrate store content before onReady.
```

adr/00080-store-key-sanitization.md
```diff
@@ -0,0 +1,28 @@
...
+## Context
+
+Issues were discovered in a few spots when accepting arbitrary keycloak client ids and setting them into the Store without sanitization. IDs like https://google.com/ failed during `Store.setItem` due to the `/` character and returned 422. 
+
+
+## Decision
+
+* Base64 encode the key after receiving it from the user and before setting it into the Store.
+* Update the Store key prefix with v2: `/data/${capabilityName}-v2-${key}`
+* Migrate before calling `this.#onReady()` in the `#receive` function of `src/lib/controller/store.ts`, checking if each key that matches the old prefix, if so, migrating to the new prefix with base64 encoding.
```
btlghrants:
```suggestion
* Migrate before calling `this.#onReady()` in the `#receive` function of `src/lib/controller/store.ts`: check each key to see if it matches the old prefix, and if so, migrating it to the new prefix with base64 encoding.
```
Extra words.

adr/00080-store-key-sanitization.md
```diff
@@ -0,0 +1,28 @@
...
+
+Issues were discovered in a few spots when accepting arbitrary keycloak client ids and setting them into the Store without sanitization. IDs like https://google.com/ failed during `Store.setItem` due to the `/` character and returned 422. 
+
+
+## Decision
+
+* Base64 encode the key after receiving it from the user and before setting it into the Store.
+* Update the Store key prefix with v2: `/data/${capabilityName}-v2-${key}`
+* Migrate before calling `this.#onReady()` in the `#receive` function of `src/lib/controller/store.ts`, checking if each key that matches the old prefix, if so, migrating to the new prefix with base64 encoding. 
+* Add a new pepr command for viewing store items. like `npx pepr view-store`, maybe a `kubectl` plugin or alias too.
```
btlghrants:
```suggestion
* Add a new `pepr` command for viewing store items--something like `npx pepr view-store`, or maybe a `kubectl` plugin and/or alias too.
```
Minor syntax stuff.

adr/00080-store-key-sanitization.md
```diff
@@ -0,0 +1,28 @@
...
+* Enhance existing unit testing to cover the new base64 encoding and migration logic. Add a battery of new fuzz and property-based tests.
+
+## Consequences
+
+1. Users cannot easily read the PeprStore CR
+2. We need to migrate existing PeprStore CRs so that the `Store.getItem/Store.removeItem` will not fail due to the base64 encoding.
+3. Will slightly slow down the store startup since we need to check and possibly migrate store before onReady.
+4. We need to build a new pepr command for viewing store items. like `npx pepr view-store`, maybe a `kubectl` plugin or alias too.
```
btlghrants:
```suggestion
4. We need to build a new `pepr` command for viewing store items. Something like `npx pepr view-store`, or maybe a `kubectl` plugin and/or alias too.
```
Minor syntax stuff.

btlghrants (commented):
Overall, not a bad start -- and I've added some suggested refinements -- but... it reads more like a plan of action than an ADR, honestly -- it describes an issue and then just hops straight into a list of TODOs supporting a specific, apparently pre-decided implementation.

Where's the reasoning / alternatives?  Will a future reader (or a soon-to-be reviewer) understand why we decided to solve the issue in the specific way we did?

For example:
- Why did we decided to code this up instead of just dumping something into the docs saying "Here's the rules!"?
- Why Base64 encode?  Why not _really_ sanitize the input by stripping out "bad" characters?
- Why did you decide to put a "v2" in _every_ key instead of having a v2 "container" injected in to keep the new / old keys separate (or whatever else)?
- Why is it worth it to us to add a new Pepr command to the CLI when we could just tell user's it's base64'd and leave it up to them to decide how to read it?

...and so on.  Happy to chat through any of that if you wanna, whenever.

adr/00080-store-key-sanitization.md
cmwylie19: committed below

cmwylie19:
> Overall, not a bad start -- and I've added some suggested refinements -- but... it reads more like a plan of action than an ADR, honestly -- it describes an issue and then just hops straight into a list of TODOs supporting a specific, apparently pre-decided implementation.
> 
> Where's the reasoning / alternatives? Will a future reader (or a soon-to-be reviewer) understand why we decided to solve the issue in the specific way we did?
> 
> For example:
> 
> * Why did we decided to code this up instead of just dumping something into the docs saying "Here's the rules!"?
> * Why Base64 encode?  Why not _really_ sanitize the input by stripping out "bad" characters?
> * Why did you decide to put a "v2" in _every_ key instead of having a v2 "container" injected in to keep the new / old keys separate (or whatever else)?
> * Why is it worth it to us to add a new Pepr command to the CLI when we could just tell user's it's base64'd and leave it up to them to decide how to read it?
> 
> ...and so on. Happy to chat through any of that if you wanna, whenever.

Sure, will add these fields. It does feel a little more like a KEP than an ADR

adr/0008-store-key-sanitization.md
```diff
@@ -0,0 +1,89 @@
...
+
+Date: 2024-07-02
+
+## Status
+
+Under Review
+
+## Context
+
+Issues were discovered when accepting arbitrary Keycloak client IDs and setting them into the Store without sanitization. For instance, IDs like `https://google.com/` failed during Store.setItem due to the `/` character, resulting in a `422` error.
```
btlghrants:
```suggestion
Issues were discovered when accepting arbitrary Keycloak client IDs as Store keys without sanitization. For instance, IDs like `https://google.com/` failed during `Store.setItem` due to the `/` character being interpreted as a path separator, resulting in a `422` error.
```

Similar to before, without clarifications like these there's no way for a future reader to really understand what's broken.

adr/0008-store-key-sanitization.md
```diff
@@ -0,0 +1,89 @@
...
+
+
+## Decision
+
+Implementing base64 would be a heavy shift for an extremely important part of the system. Instead, we will sanitize the key by replacing `/` with a character that does not break the `patch` operation before placing the key. We will also increase the tests to find other edge cases before releasing the change.
+
+#### Sanitize using String Replacement
+* Sanitize the key by replacing `/` with a character that does not break the `patch` operation before placing the key.
```
btlghrants: Haven't read too deeply into the specs, but... since the patch operation is a [JSON Patch](https://datatracker.ietf.org/doc/html/rfc6902/#section-4), it's "data" key is defined as a [JSON Pointer](https://datatracker.ietf.org/doc/html/rfc6901#section-3) object... which means that both "~" and "/" are going to have special meanings:

```
A JSON Pointer is a Unicode string (see [[RFC4627], Section 3](https://datatracker.ietf.org/doc/html/rfc4627#section-3))
   containing a sequence of zero or more reference tokens, each prefixed
   by a '/' (%x2F) character.

   Because the characters '~' (%x7E) and '/' (%x2F) have special
   meanings in JSON Pointer, '~' needs to be encoded as '~0' and '/'
   needs to be encoded as '~1' when these characters appear in a
   reference token.
   ```

**Review by btlghrants — COMMENTED** [review]
Still working it through, but these should be reasonable comments for the short-term.

File: adr/0008-store-key-sanitization.md [hunk 9]
```diff
@@ -0,0 +1,89 @@
...
+* Increasing the tests to find other edge cases before releasing the change.
```
btlghrants, line 20: It might be nice to think through / add some examples of the sorts of keys we'd like the Store to be able to handle in order to give "other edge cases" (and their respective, future tests) some shape.  Like:
- URIs (cuz that's what drove this PR to begin with)
- dot-delimited strings -- e.g. "a.b.c.d" -- just to be sure our escaping can handle things that look like JSON
- " " (empty strings)
- "null" (JS keywords)
- ...and so on.

adr/0008-store-key-sanitization.md
cmwylie19: yep good look, agree but respelled separator

adr/0008-store-key-sanitization.md
cmwylie19: added this link and spoke to the replacement but kept specific implementation as TBD by the testing done during the implementation.

adr/0008-store-key-sanitization.md
cmwylie19: I hear you on URIs but Store.*Item only accepts strings. 

It can already handle dot-delimited and empty strings. 

The only error we know of until doing more testing and finding more would be url as a string.

```yaml
  data:
    __pepr_do_not_delete__: k-thx-bye
    hello-pepr-: b #the key is empty here
    hello-pepr-a.a.a: a
  kind: PeprStore
```

```ts
OnSchedule({
  name: "set-items",
  unit: "seconds",
  every: 15,
  run: () => {
    Store.setItem("a.a.a","a")
    Store.setItem("","b")
  }
})
```

adr/0008-store-key-sanitization.md
```diff
@@ -0,0 +1,89 @@
...
+Under Review
+
+## Context
+
+Issues were discovered when accepting arbitrary Keycloak client IDs as Store keys without sanitization. For instance, IDs like `https://google.com/` failed during `Store.setItem` due to the `/` character being interpreted as a path separator, resulting in a `422` error.
+
+
+## Decision
+
+Sanitize the keys by replacing `/` with [a character that does not break the `patch` operation](https://datatracker.ietf.org/doc/html/rfc6902/#section-4) before setting, getting, or deleting a key in/from the store. We will also increase the tests to find other edge cases before releasing the change and deciding upon the exact character or pattern. We are prioritizing the cluster operator persona in this case who wants to quickly check all keys and values in the store by looking at the PeprStore CR.
```
btlghrants:
```suggestion
Sanitize the keys by replacing `/` with [one or more characters](https://datatracker.ietf.org/doc/html/rfc6901#section-3) that do not break the [`patch` operation](https://datatracker.ietf.org/doc/html/rfc6902/#section-4) before setting, getting, or deleting a key in/from the store. We will also increase the tests to find other edge cases before releasing the change and deciding upon the exact character or pattern. We are prioritizing the cluster operator persona in this case who wants to quickly check all keys and values in the store by looking at the PeprStore CR.
```

btlghrants (approved):

PR #1025, opened 2024-08-03
feat: sharded queue

### LINKED ISSUE #1019: `.Reconcile` queue based on resource ID rather than kind

In uds-core we have a custom resource named `Package` that we watch for create/update events:

https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/index.ts#L57-L62

Each event is queued via `Reconcile`. In effect this means that all `Package` resource creations/updates are placed in a queue together. As one quick example of the effects:
- Create `Package` A, B, and C - all in a single apply
- Result is that nothing happens to `Package` C until A and B have completed

The same is true on updates:
- Create `Package`s A, B, C
- Wait until all are ready
- Create `Package` D and update `Package` C
- `Package` C is not updated until D is fully created

Ideally I would like each `Package` to be queued separately - so updates on C wait for previous ones to complete, but A/B/D do not affect the timing/waiting.

Provide an option for `.Reconcile` to optionally be queued per resource. This could be done based on resource namespace + name, or some other unique identifier like UID. This would speed up reconcile time when multiple resources are created at the same time.

Separate bindings could be created per namespace or per name, but this does not work well for an extensible operator with resources in arbitrary namespaces.

### Additional context

This probably should be an option, but not the only option since certain reconcile code may not behave well on concurrent operations, even if restricted to different resources.

cmwylie19:
## Description

Changes the Queue to a queueRecord<string, Queue> so that we enqueue items of the same name. This creates multiple queues but cuts down on reconciliation time. This is based on issue #1019. 

e2e test is here - https://github.com/defenseunicorns/pepr-excellent-examples/pull/52 which includes the original tests

Think about adding a new metrics for queue, maybe items per queue

## Related Issue

Fixes #1019 

Relates to #

## Type of change

## Checklist before merging


btlghrants (approved):
LGTM (obvi) since I pushed the last commit, so... I'll approve but leave for @cmwylie19 to review / merge.

cmwylie19:
Nice job on the feature flagging!

cmwylie19:
Since I wrote the feature and you reviewed it, and you wrote the feature flags and I reviewed it and ran the e2e tests and we don't have another person I will merge this.

PR #1086, opened 2024-08-23
feat: add custom liveness and readiness probes to helm chart

### LINKED ISSUE #1085: Add capability to override readiness and liveness probes
When starting Pepr in a slower or resource constrained environment, the pods frequently encounter crashloops due to not responding to the readiness or liveness probes quickly enough.

- **Given** a resource constrained environment
- **When** a probe override is defined in the helm values
- **Then** the deployments are modified to use the probe definitions from the overridden helm values

The static probe definitions could be updated with a sufficiently long `initialDelaySeconds` but this assumes that all environments would have a long startup period which is not the case.  Or the number of probe failures could be increased, however this would mask true failure states and delay recovery.

### Additional context
N/A

nfoucha:
## Description
This PR replaces the hard-coded readiness and liveness probes in the helm chart with values references to allow for overrides.  `initialDelaySeconds` have also been added to the helm values defaults and the pod definitions to account for nominal startup time of the Pepr pods.
...

## Related Issue

Fixes #1085

## Type of change

## Checklist before merging

### DIFF
--- a/docs/030_user-guide/120_customization.md
+++ b/docs/030_user-guide/120_customization.md
@@ -68,6 +68,8 @@ Below are the available Helm override configurations after you have built your P
 | `annotations`                                | Deployment annotations                                              |
 | `labels`                                     | Deployment labels                                                   |
 | `securityContext`                            | Pod security context                                                |
+| `readinessProbe`                             | Pod readiness probe definition                                      |
+| `livenessProbe`                              | Pod liveness probe definition                                       |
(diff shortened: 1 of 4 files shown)


cmwylie19:
Thanks! This looks good. I will give it review by Monday!

cmwylie19 (approved):

cmwylie19:
Looks GREAT! Just tested it. Thank you for the contribution @nfoucha !

Issue #1022, opened 2024-08-02
Add native support/handling for Finalizers


mjnagel:
### Is your feature request related to a problem? Please describe.

When building operators with Pepr and other tools it is often useful to add a [finalizer](https://kubernetes.io/docs/concepts/overview/working-with-objects/finalizers/) to custom resources to handle cleanup when a resource is deleted.

To manage finalizers an end user has to:
- Add a finalizer to new CRs that are created (could be done via mutation or a patch)
- Watch for resource updates that contain the deletionTimestamp metadata field (when a resource with a finalizer is deleted it appears as an update, rather than a deletion)
- Based on ^ filter updates to a resource to either perform some cleanup OR process "reconciliation"
- Remove the finalizer to finish the resource deletion

### Describe the solution you'd like

It would be great if Pepr natively supported the concept of finalizers and handled the operations around them for me. What this might look like in my code (referencing uds-core):

```typescript
When(UDSPackage)
  .IsCreatedOrUpdated()
  // Advanced CR validation
  .Validate(validator)
  // Enqueue the package for processing
  .Reconcile(packageReconciler)
  // Run a `cleanupPackage` function for the finalizer
  .Finalizer(cleanupPackage);
```

In effect this would mean that Pepr would need to manage:
- Adding the finalizer to the resource
- Triggering the function I specify at the proper time (when a `deletionTimestamp` is added)
- Removing the finalizer after the function completes

### Describe alternatives you've considered

This could be left to individual module developers to handle on their own, which is a reasonable path all things considered (I did a rough draft of this [here](https://github.com/defenseunicorns/uds-core/pull/634)). However this does seem like it may be a common need as people develop operators with Pepr.

### Additional context

One thing to note - it is common to patch `status` on a resource when deletion kicks off, while finalizers are still running. Typically you'd see this reflected as a `Terminating` or `Deleting` phase on the resource. This could be beneficial for Pepr to also include in the finalization lifecycle, but I'm unsure if that would be possible since the CRD may have a restrictive/custom type for their status field. The other option would be for the module creator to have it as part of their "cleanup" function, and either Pepr or the cleanup function would need to properly filter out the status update event so that finalization only happens once.

References for other "operator builders" and how they handle this:
- [Operator SDK](https://sdk.operatorframework.io/docs/building-operators/ansible/reference/finalizers/): Would be nearly a 1:1 with the proposal in this issue, operator author adds a finalizer to their watch and the SDK manages the lifecycle of that
[...]

PR #1102, opened 2024-08-28
feat: add Finalize() to Pepr

btlghrants:
## Description

Add a fluent Finalize() method to Pepr.

End to End Test:  defenseunicorns/pepr-excellent-examples#58

## Related Issue

Relates to defenseunicorns/pepr#1022

## Type of change

## Checklist before merging


package.json
```diff
@@ -22,15 +22,15 @@
     "prebuild": "rm -fr dist/* && npm run gen-data-json",
     "version": "node scripts/set-version.js",
     "build": "tsc && node build.mjs",
-    "build:image": "npm run build && docker buildx build --tag pepr:dev .",
+    "build:image": "npm run build && docker buildx build --output type=docker --tag pepr:dev .",
```
cmwylie19, line 25:
```suggestion
    "build:image": "npm run build && docker buildx build --output type=docker --tag pepr:dev .",
```

I thought I only needed this, did you intend on adding this?

btlghrants: Yep, was intended -- it doesn't hurt the Linux box I work off of & it _definitely_ helps your Mac work so I don't see why we wouldn't just leave it in there. :man_shrugging:

cmwylie19:
this runs great, tested it thoroughly in e2e and locally

cmwylie19 (approved):

src/lib/watch-processor.ts
```diff
@@ -100,7 +101,20 @@ async function runBinding(binding: Binding, capabilityNamespaces: string[]) {
...
         if (filterMatch === "") {
-          await binding.watchCallback?.(obj, phase);
+          if (binding.isFinalize) {
+            if (!obj.metadata?.deletionTimestamp) {
+              return;
+            }
+            try {
+              await binding.finalizeCallback?.(obj);
+
+              // irrespective of callback success / failure, remove pepr finalizer
```
bburky, line 111: I can see some use cases where a retry might be preferred. But perhaps that's not the common case.

I'd at least suggesting logging when the `finalizeCallback` errors.

btlghrants: Yeah, that part is a little non-obvious, but because the `try/finally` that's attempting to run the `finalizeCallback()` is _itself_ inside a `try/catch`, the outer `catch` would handle logging a message in the case the callback fails, no?

Simple repro of that behavior below (just put it in a script & run `node <whatever>.js`, etc.):
```
try {
  try { throw "error" } finally { console.log( "fin" ) }
}
catch( e ) { console.log( e ) }
```

Or maybe you were thinking something different / more specific?

btlghrants: As for potential retry logic, agreed -- we think there're probably situations where it would be useful too, just didn't want to crowbar that in without proper research & a specific, user-requested use case -- partly to keep this feature as simple as possible, partly to be sure we weren't assuming what sorts of retry logic would be desired.

I wouldn't be surprised to see a feature like that roll out in the future though, especially once Finalize() sees some use & we get a feel for it's rough edges.

PR #1104, opened 2024-08-29
chore: support optional fields in ownerReferences

### LINKED ISSUE #1060: Enhance the ownerRef helper in SDK
Need to add more finalizer support and protect "owned" resources so they don't get deleted unless the owning object is deleted.

```ts
  const { name, uid } = instance.metadata!;

  return [
    {
      apiVersion: instance.apiVersion!,
      kind: instance.kind!,
      uid: uid!,
      name: name!,
      controller: true, // add
      blockOwnerDeletion: true, // add
    },
  ];
}
```
[Kubernetes Docs around ownerRefs](https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/object-meta/#System)

- **Given** a controller deploys a deployment that is owned by a `WebApp`
- **When** someone tries to delete the deployment
- **Then** deletion is blocked until the WebApp is deleted

samayer12:
## Description

This PR adds support for optional fields defined in [ownerReferences](https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/object-meta/#System). It also changes the test block to use a parameterized test.

End to End Test:  N/A

## Related Issue

Fixes #1060

## Type of change

## Checklist before merging

### DIFF
--- a/docs/030_user-guide/015_sdk.md
+++ b/docs/030_user-guide/015_sdk.md
@@ -50,6 +50,8 @@ let result = containers(peprValidationRequest, "ephemeralContainers")
 Returns the owner reference for a Kubernetes resource as an array. Accepts the following parameters:
 
 - **@param kubernetesResource: GenericKind** The Kubernetes resource to get the owner reference for
+- **@param blockOwnerDeletion: boolean** If true, AND if the owner has the "foregroundDeletion" finalizer, then the owner cannot be deleted from the key-value store until this reference is removed. 
+- **@param controller: boolean** If true, this reference points to the managing controller.
 
 **Usage:**
 
--- a/docs/030_user-guide/015_sdk.md
+++ b/docs/030_user-guide/015_sdk.md
@@ -47,7 +47,7 @@ let result = containers(peprValidationRequest, "ephemeralContainers")
 
 ## `getOwnerRefFrom`
 
-Returns the owner reference for a Kubernetes resource. Accepts the following parameters:
+Returns the owner reference for a Kubernetes resource as an array. Accepts the following parameters:
 
 - **@param kubernetesResource: GenericKind** The Kubernetes resource to get the owner reference for
 
--- a/src/sdk/sdk.test.ts
+++ b/src/sdk/sdk.test.ts
@@ -165,13 +165,6 @@ describe("writeEvent", () => {
 
 describe("getOwnerRefFrom", () => {
 
-  const V1OwnerReferenceFieldCount = Object.getOwnPropertyNames(V1OwnerReference).length
-  const crWithoutOptionals = {
-    apiVersion: "v1",
-    kind: "Package",
-    metadata: { name: "test", namespace: "default", uid: "1" },
-  };
-
 
   it("should return the owner reference for the CRD", () => {
     const cr = {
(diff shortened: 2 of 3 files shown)


src/sdk/sdk.ts
```diff
@@ -79,20 +79,27 @@ export async function writeEvent(
...
+  };
+
+  if (Object.prototype.hasOwnProperty.call(customResource, "blockOwnerDeletion")) {
+    ownerReference.blockOwnerDeletion = customResource.blockOwnerDeletion;
+  }
+  if (Object.prototype.hasOwnProperty.call(customResource, "controller")) {
+    ownerReference.controller = customResource.controller;
+  }
+
+  return [ownerReference];
```
samayer12: Why is this an array? Can a customResource provide multiple ownerReferences? What would that look like in practice?

src/sdk/sdk.ts
```diff
@@ -79,20 +79,27 @@ export async function writeEvent(
...
  * Get the owner reference for a custom resource
- * @param cr the custom resource to get the owner reference for
- * @returns the owner reference for the custom resource
+ * @param customResource the custom resource to get the owner reference for
+ * @returns the owner reference array for the custom resource
  */
-export function getOwnerRefFrom(cr: GenericKind): V1OwnerReference[] {
-  const { name, uid } = cr.metadata!;
+export function getOwnerRefFrom(customResource: GenericKind): V1OwnerReference[] {
+  const { name, uid } = customResource.metadata!;
```
samayer12: Maybe better to use the `spread` operator instead of making non-null assertions?

src/sdk/sdk.test.ts
```diff
@@ -179,7 +180,57 @@ describe("getOwnerRefFrom", () => {
...
+  }));
+  const ownerRefWithBlockOwnerDeletion = ownerRefWithoutOptionals.map(item => ({
+    ...item,
+    blockOwnerDeletion: false,
+  }));
+  const ownerRefWithAllFields = ownerRefWithoutOptionals.map(item => ({
+    ...item,
+    controller: true,
+    blockOwnerDeletion: true,
+  }));
```
samayer12: Those whole block feels tedious, but not necessarily better than when the tests weren't parameterized. For reference, [here's](https://github.com/defenseunicorns/pepr/pull/1104/commits/86ff9632998df4bc13f5d19d8a0c34e136d43449) what that looked like.

src/sdk/sdk.ts
```diff
@@ -79,20 +79,27 @@ export async function writeEvent(
...
-    },
-  ];
+  const ownerReference: V1OwnerReference = {
+    apiVersion: customResource.apiVersion!,
+    kind: customResource.kind!,
+    uid: uid!,
+    name: name!,
+  };
+
+  if (Object.prototype.hasOwnProperty.call(customResource, "blockOwnerDeletion")) {
```
samayer12: This could be `hasOwn()`, maybe some nullish-coalescing?

samayer12: Went with this implementation based off linter feedback: https://eslint.org/docs/latest/rules/no-prototype-builtins

src/sdk/sdk.test.ts
```diff
@@ -179,7 +180,57 @@ describe("getOwnerRefFrom", () => {
...
+
+  const V1OwnerReferenceFieldCount = Object.getOwnPropertyNames(V1OwnerReference).length;
+
+  test.each([
+    [crWithAllFields, ownerRefWithAllFields, V1OwnerReferenceFieldCount],
+    [crWithBlockOwnerDeletion, ownerRefWithBlockOwnerDeletion, V1OwnerReferenceFieldCount - 1],
+    [crWithController, ownerRefWithController, V1OwnerReferenceFieldCount - 1],
+    [crWithoutOptionals, ownerRefWithoutOptionals, V1OwnerReferenceFieldCount - 2],
```
samayer12: Pull the FieldCount checks out to it's own test.

src/sdk/sdk.ts
cmwylie19: Objects can have multiple owners in k8s for sure

src/sdk/sdk.ts
cmwylie19: a customResource does not contains these fields at all

```plaintext
KIND:       Pod
VERSION:    v1

DESCRIPTION:
    Pod is a collection of containers that can run on a host. This resource is
    created by clients and scheduled onto hosts.
    
FIELDS:
  apiVersion	<string>
    APIVersion defines the versioned schema of this representation of an object.
    Servers should convert recognized schemas to the latest internal value, and
    may reject unrecognized values. More info:
    https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources

  kind	<string>
    Kind is a string value representing the REST resource this object
    represents. Servers may infer this from the endpoint the client submits
    requests to. Cannot be updated. In CamelCase. More info:
    https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds

  metadata	<ObjectMeta>
    Standard object's metadata. More info:
    https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#metadata

  spec	<PodSpec>
    Specification of the desired behavior of the pod. More info:
    https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#spec-and-status

  status	<PodStatus>
    Most recently observed status of the pod. This data may not be up to date.
    Populated by the system. Read-only. More info:
    https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#spec-and-status
```

We have to place them on the CR
```plaintext
 > k explain pod.metadata.ownerReferences
KIND:       Pod
VERSION:    v1

FIELD: ownerReferences <[]OwnerReference>

DESCRIPTION:
    List of objects depended by this object. If ALL objects in the list have
    been deleted, this object will be garbage collected. If this object is
    managed by a controller, then an entry in this list will point to this
    controller, with the controller field set to true. There cannot be more than
    one managing controller.
    OwnerReference contains enough information to let you identify an owning
    object. An owning object must be in the same namespace as the dependent, or
    be cluster-scoped, so there is no namespace field.
    
FIELDS:
  apiVersion	<string> -required-
    API version of the referent.

  blockOwnerDeletion	<boolean>
    If true, AND if the owner has the "foregroundDeletion" finalizer, then the
    owner cannot be deleted from the key-value store until this reference is
    removed. See
    https://kubernetes.io/docs/concepts/architecture/garbage-collection/#foreground-deletion
    for how the garbage collector interacts with this field and enforces the
    foreground deletion. Defaults to false. To set this field, a user needs
    "delete" permission of the owner, otherwise 422 (Unprocessable Entity) will
    be returned.

  controller	<boolean>
    If true, this reference points to the managing controller.

  kind	<string> -required-
    Kind of the referent. More info:
    https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds

  name	<string> -required-
    Name of the referent. More info:
    http://kubernetes.io/docs/user-guide/identifiers#names

  uid	<string> -required-
    UID of the referent. More info:
    http://kubernetes.io/docs/user-guide/identifiers#uids

```

samayer12: Great, that `k explain` comparison was super useful for my context! Tracking the intended use now.

src/sdk/sdk.test.ts
samayer12: Discussed with @btlghrants for ways to make this block more concise.

src/sdk/sdk.test.ts
```diff
@@ -163,23 +164,55 @@ describe("writeEvent", () => {
...
+  test.each([
+    [true, false, ownerRefWithAllFields],
+    [false, undefined, ownerRefWithBlockOwnerDeletion],
+    [undefined, true, ownerRefWithController],
+    [undefined, undefined, ownerRef],
+  ])(
+    "should return owner reference for the CRD for combinations of V1OwnerReference fields - Optionals: blockOwnerDeletion (%s), controller (%s)",
+    (blockOwnerDeletion, controller, expected) => {
+      const result = getOwnerRefFrom(customResource, blockOwnerDeletion, controller);
+      expect(result).toStrictEqual(expected);
```
samayer12, line 205: `toStrictEqual()` is required here because a `toEqual()` would let something like `{ controller: undefined }` pass

src/sdk/sdk.ts
```diff
@@ -79,18 +79,25 @@ export async function writeEvent(
...
   return [
     {
-      apiVersion: cr.apiVersion!,
-      kind: cr.kind!,
+      apiVersion: apiVersion!,
+      kind: kind!,
       uid: uid!,
       name: name!,
+      ...(blockOwnerDeletion !== undefined && { blockOwnerDeletion }),
+      ...(controller !== undefined && { controller }),
```
samayer12, line 102: In English, this "Spreads the contents of the corresponding input parameter to the returned object if the parameter was set."

docs/030_user-guide/015_sdk.md
```diff
@@ -47,7 +47,7 @@ let result = containers(peprValidationRequest, "ephemeralContainers")
 
 ## `getOwnerRefFrom`
 
-Returns the owner reference for a Kubernetes resource. Accepts the following parameters:
+Returns the owner reference for a Kubernetes resource as an array. Accepts the following parameters:
 
 - **@param kubernetesResource: GenericKind** The Kubernetes resource to get the owner reference for
```
cmwylie19, line 52:
```suggestion
- **@param kubernetesResource: GenericKind** The Kubernetes resource to get the owner reference for
```
Add new params here
  blockOwnerDeletion?: boolean,
  controller?: boolean,

cmwylie19, line 52: add new params that are added
  
  blockOwnerDeletion?: boolean,
  controller?: boolean,

cmwylie19, line 52:
```suggestion
Returns the owner reference for a Kubernetes resource as an array. Accepts the following parameters:

- **@param kubernetesResource: GenericKind** The Kubernetes resource to get the owner reference for
- **@param blockOwnerDeletion: boolean** If true, AND if the owner has the "foregroundDeletion" finalizer, then the owner cannot be deleted from the key-value store until this reference is removed. 
- **@param controller: boolean** If true, this reference points to the managing controller.
```

src/sdk/sdk.ts
```diff
@@ -79,18 +79,25 @@ export async function writeEvent(
 
 /**
  * Get the owner reference for a custom resource
- * @param cr the custom resource to get the owner reference for
- * @returns the owner reference for the custom resource
+ * @param customResource the custom resource to get the owner reference for
+ * @returns the owner reference array for the custom resource
```
cmwylie19:
```suggestion
 * @param customResource the custom resource to get the owner reference for
 * @param blockOwnerDeletion if true, AND if the owner has the "foregroundDeletion" finalizer, then the owner cannot be deleted from the key-value store until this reference is removed. 
 * @param controller if true, this reference points to the managing controller.
 * @returns the owner reference array for the custom resource
```

cmwylie19 (changes requested):
Once we get the comments addressed, this one is ready for merge. Since I will be on PTO feel free to override my request changes

cmwylie19 (dismissed):

cmwylie19 (approved):
lgtm

PR #1108, opened 2024-08-30
chore: exclude dev deps

### LINKED ISSUE #1103: esbuild should not package devDependencies in modules
To keep size down, we need to ensure we are not adding dev dependencies to the secret created from the Pepr module.

- **Given** a build occurs
- **When** there are devDependencies, they are excluded
- **Then** the size is smaller since they are not necessary for prod

Links:
- Adding devDeps to externalLibs
- Through env var https://github.com/evanw/esbuild/issues/2377#issuecomment-1179344638 

cmwylie19:
## Description

This excludes devDependencies in the module other than TypeScript. This PR only cuts out the keys from the devDependencies in the built js.

End to End Test:    

## Related Issue

Fixes #1103 

Relates to #

## Type of change

## Checklist before merging

### DIFF
--- a/src/cli/build.ts
+++ b/src/cli/build.ts
@@ -244,6 +244,10 @@ export async function loadModule(entryPoint = peprTS) {
 export async function buildModule(reloader?: Reloader, entryPoint = peprTS, embed = true) {
   try {
     const { cfg, modulePath, path, uuid } = await loadModule(entryPoint);
+    const excludedDeps = [
+      ...externalLibs,
+      ...Object.keys(cfg.devDependencies).filter(dep => dep !== "typescript"),
+    ];
 
     const validFormat = await peprFormat(true);
 
@@ -265,7 +269,7 @@ export async function buildModule(reloader?: Reloader, entryPoint = peprTS, embe
     const ctxCfg: BuildOptions = {
       bundle: true,
       entryPoints: [entryPoint],
-      external: externalLibs,
+      external: excludedDeps,
       format: "cjs",
       keepNames: true,
       legalComments: "external",


cmwylie19:
externLibs Before
image

After:
image

src/cli/build.ts
```diff
@@ -244,6 +244,10 @@ export async function loadModule(entryPoint = peprTS) {
 export async function buildModule(reloader?: Reloader, entryPoint = peprTS, embed = true) {
   try {
     const { cfg, modulePath, path, uuid } = await loadModule(entryPoint);
+    const excludedDeps = [
+      ...externalLibs,
+      ...Object.keys(cfg.devDependencies).filter(dep => dep !== "typescript"),
```
btlghrants, line 249: This seems risky to me.

IF esbuild has included a devDependency into the built module it is because a module author has _referenced a dev dependency from their production code_, which means it's _necessary_ to package it.

Excluding the devDeps from the final module output means that such an erroneous module will build successfully but that it'll fail at runtime because we're not packaging the devDeps into the controller image.  This seems _much_ less ideal that just building the mis-configured dep into the module.

At least in that case the module will still work while they work on fixing their deps, no?

btlghrants (commented):
I think this change induces some risk / provides no real benefit.  My vote would be not to do this.  More comments inline.

src/cli/build.ts
cmwylie19: Which is an anti-pattern to reference devDepencencies in prod code in the first place. It was just a suggestion from UDS Core to exclude devDeps from prod code and I tend to agree.

Since you feel strongly we will just close the PR and open it again if it comes up in the future.

cmwylie19:
I'm not sure that it is risky

PR #1194, opened 2024-09-30
chore: remove deprecated types from k8s.ts

cmwylie19:
## Description

We had a circular dependency which caused us to move some types from src/lib/k8s.ts to src/lib/types.ts. We marked them as deprecated and published the news in the release notes. Now we need to remove the types from src/lib/k8s.ts

End to End Test:  N/A

## Related Issue

Fixes #1193 

Relates to #

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Other (security config, docs update, etc)

## Checklist before merging
- [x] Unit, [Journey](https://github.com/defenseunicorns/pepr/tree/main/journey), [E2E Tests](https://github.com/defenseunicorns/pepr-excellent-examples), [docs](https://github.com/defenseunicorns/pepr/tree/main/docs), [adr](https://github.com/defenseunicorns/pepr/tree/main/adr) added or updated as needed
- [x] [Contributor Guide Steps](https://docs.pepr.dev/main/contribute/#submitting-a-pull-request) followed

PR #1211, opened 2024-10-02
Revert "chore: remove deprecated types from k8s.ts" - slated for v0.38.0

cmwylie19:
Reverts defenseunicorns/pepr#1194

We need to do a patch today, PR is necessary before `v0.38.0` but we do not want to create any potential breaking changes due to deprecations before


btlghrants (approved):
Yep.

PR #1215, opened 2024-10-03
chore: add watch logs to soak test to determine reasons for failure

### LINKED ISSUE #1213: chore: add logs to soak test

Our soak test is a long running audit of our watcher as it watches resources in the cluster. A test failure is when a pod lives beyond the relist interval. We collect metrics around the time this occurs but we do not have insight into the reason why it fails. For that reason, we need to collect logs before exiting the test to determine the reason for failure. This will help when we are testing new implementations of the watcher.

cmwylie19:
## Description

Our soak test has a blind spot as to what goes wrong in a failure scenario. Adding logs should help us determine the reason. 

## Related Issue

Fixes #1213 

Relates to #

## Type of change

## Checklist before merging

### DIFF
--- a/.github/workflows/soak.yaml
+++ b/.github/workflows/soak.yaml
@@ -184,12 +184,14 @@ jobs:
           }
           touch logs/auditor-log.txt
           touch logs/informer-log.txt
+          touch logs/watch-logs.txt
 
           update_pod_map
 
           collect_metrics() {
             kubectl exec metrics-collector -n watch-auditor -- curl watch-auditor:8080/metrics | grep watch_controller_failures_total > logs/auditor-log.txt
             kubectl exec metrics-collector -n watch-auditor -- curl -k https://pepr-soak-ci-watcher.pepr-system.svc.cluster.local/metrics | egrep -E "pepr_cache_miss|pepr_resync_failure_count" > logs/informer-log.txt
+            kubectl logs deploy/pepr-soak-ci-watcher -n pepr-system > logs/watch-logs.txt
           }
 
           # Start collecting metrics every 5 minutes and checking pod counts every 30 minutes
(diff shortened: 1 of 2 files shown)


src/lib/watch-processor.test.ts
```diff
@@ -26,7 +26,9 @@ jest.mock("./metrics", () => ({
     incRetryCount: jest.fn(),
   },
 }));
-
+it("something", () => {
+  expect(1).toBe(1);
+});
```
cmwylie19:
```suggestion
```

src/lib/watch-processor.test.ts
```diff
@@ -26,7 +26,6 @@ jest.mock("./metrics", () => ({
     incRetryCount: jest.fn(),
   },
 }));
```
cmwylie19:
```suggestion
}));

```

.github/workflows/soak.yaml
```diff
@@ -208,6 +210,7 @@ jobs:
                 echo "$pod: ${pod_map[$pod]}"
                 if [ "${pod_map[$pod]}" -gt 1 ]; then
                   echo "Test failed: Pod $pod has count ${pod_map[$pod]}"
+                  kubectl logs deploy/pepr-soak-ci-watcher -n pepr-system
```
btlghrants, line 213: May be worth pushing this log dump into the `watch-logs.txt` file too, no?  Not sure that the bash logs get truncated like our JS-based-test log output (they don't, right?), but it'd be nice to have a version of it that can be downloaded & scripted through in case the output get's really large.

Or maybe it's not worth the trouble?

cmwylie19: Problem is the test exits and does not dump

https://github.com/defenseunicorns/pepr/actions/runs/11147804732/job/30983047281

cmwylie19: so this is like the last mile to throw it in the logs

btlghrants (approved):

cmwylie19:
the test that is failing is a version test but it does not indicate that any functionality is wrong. This PR does not touch Pepr Core. Going to go ahead and merge so I can use this updated logic in my soak for the native fetch

PR #1251, opened 2024-10-09
feat: rbac overrides in package.json

### LINKED ISSUE #814: Have a way to create overrides for RBAC in a source file.

We want to leverage `package.json` to defined RBAC settings that will work with RBAC and create an excellent example test. 

- **Given** we want to create a reproducible way to generate scoped RBAC 
- **When** we define the RBAC in package.json and run a build
- **Then** we have our customized RBAC

Definition of done: 
- [ ] bring rbac rules to values.yaml

schaeferka:
## Description
We want to leverage package.json to defined RBAC settings that will work with RBAC and create an excellent example test.

Describe the solution you'd like
Given we want to create a reproducible way to generate scoped RBAC
When we define the RBAC in package.json and run a build
Then we have our customized RBAC
...

End to End Test:    

## Related Issue

Fixes #814 

Relates to #

## Type of change

## Checklist before merging


src/lib/assets/helm.ts
```diff
@@ -207,13 +207,99 @@ export function admissionDeployTemplate(buildTimestamp: string) {
...
+  apiGroups: string[];
+  resources: string[];
+  verbs: string[];
+}
+
+export interface RoleRule {
+  apiGroups: string[];
+  resources: string[];
+  verbs: string[];
+}
```
cmwylie19: can you use upstream types, with kubernetes/client-node? 

These are the same type:

```plaintext
$ k explain clusterRole.rules
GROUP:      rbac.authorization.k8s.io
KIND:       ClusterRole
VERSION:    v1

FIELD: rules <[]PolicyRule>

$ k explain role.rules 
GROUP:      rbac.authorization.k8s.io
KIND:       Role
VERSION:    v1

FIELD: rules <[]PolicyRule>


```

schaeferka: the interfaces are now gone

journey/pepr-build.ts
```diff
@@ -37,57 +50,75 @@ export function peprBuild() {
...
+async function validateClusterRoleYaml() {
+  // Read the generated yaml files
+  const k8sYaml = await fs.readFile(
+    resolve(cwd, outputDir, "pepr-module-static-test.yaml"),
+    "utf8",
+  );
+  const cr = await fs.readFile(resolve("journey", "resources", "clusterrole.yaml"), "utf8");
+
+  console.log("K8s YAML:", k8sYaml);
+  console.log("CR YAML:", cr);
```
cmwylie19:
```suggestion
```

docs/030_user-guide/120_customization.md
```diff
@@ -132,5 +132,46 @@ Below are the available configurations through `package.json`.
...
+
+### Example Custom RBAC Rules
+
+The following example demonstrates how to add custom RBAC rules to the Pepr module.
+
+```json
+{
+  "pepr": {
+    "rbac": {
+      "roles": [
```
cmwylie19: this should be role and clusterRole since we only have one of each. The plural is misleading

src/lib/assets/yaml.ts
```diff
@@ -187,7 +218,7 @@ export function zarfYaml({ name, image, config }: Assets, path: string) {
     ],
   };
 
-  return dumpYaml(zarfCfg, { noRefs: true });
+  return dumpYaml(zarfCfg);
```
cmwylie19: what is the reason for this change?

src/lib/assets/yaml.ts
```diff
@@ -158,10 +175,24 @@ export async function overridesFile({ hash, name, image, config, apiToken }: Ass
...
     },
+
+    // Custom RBAC section to add ClusterRoles and Roles
+    rbac: {
+      clusterRoles: customClusterRoleRules.map((rule, index) => ({
+        name: `custom-cluster-role-${index + 1}`,
+        rules: rule,
+      })),
+      roles: customStoreRoleRules.map((rule, index) => ({
+        name: `custom-role-${index + 1}`,
```
cmwylie19: we need to use the RBAC that we create

src/lib/assets/yaml.ts
```diff
@@ -158,10 +175,24 @@ export async function overridesFile({ hash, name, image, config, apiToken }: Ass
         annotations: {},
       },
     },
+
+    // Custom RBAC section to add ClusterRoles and Roles
+    rbac: {
+      clusterRoles: customClusterRoleRules.map((rule, index) => ({
+        name: `custom-cluster-role-${index + 1}`,
```
cmwylie19: we need to use the rbac that we already create

src/lib/assets/yaml.ts
schaeferka: Reverted to original

src/lib/assets/yaml.ts
schaeferka: [@Case](https://defense-unicorns.slack.com/team/U05AJDH3ME0) Quick question for you on the RBAC PR. For this section of code in yaml.ts:

```yaml
 rbac: {
      clusterRoles: customClusterRoleRules.map((rule, index) => ({
        name: `custom-cluster-role-${index + 1}`,
        rules: rule,
      })),
      roles: customStoreRoleRules.map((rule, index) => ({
        name: `custom-role-${index + 1}`,
        rules: rule,
      })),
```

You commented that we need to use the RBAC we create. Do you mean the scoped/admin rules that are always generated, the custom rules from the package.json file, or both? I just want to make sure since the original yaml.ts file didn't include the scoped/admin rules in the overrides file.

src/lib/assets/yaml.ts
cmwylie19: use singular role and clusterRole

schaeferka: Changed to singular

docs/030_user-guide/120_customization.md
```diff
@@ -132,5 +132,46 @@ Below are the available configurations through `package.json`.
 | `alwaysIgnore`   | Conditions to always ignore            | `{namespaces: []}`  |
 | `includedFiles`  | For working with WebAssembly           | ["main.wasm", "wasm_exec.js"]   |
 | `env`            | Environment variables for the container| `{LOG_LEVEL: "warn"}`           |
+| `rbac`          | Custom RBAC rules                    | `{"rules": [{"apiGroups": ["<apiGroups>"], "resources": ["<resources>"], "verbs": ["<verbs>"]}]}` |
```
cmwylie19:
```suggestion
| `rbac`          | Custom RBAC rules                    | `"<role/clusterRole>": {"rules": [{"apiGroups": ["<apiGroups>"], "resources": ["<resources>"], "verbs": ["<verbs>"]}]}` |
```

src/lib/module.ts
```diff
@@ -9,6 +9,15 @@ import { CapabilityExport, AdmissionRequest } from "./types";
 import { setupWatch } from "./watch-processor";
 import { Log } from "../lib";
 
+// This should be the upstream type, needs to be exported from KFC
+export interface Rule {
```
cmwylie19: Upstream types `PolicyRule`

src/lib/types.ts
```diff
@@ -77,6 +77,9 @@ export interface CapabilityCfg {
 export interface CapabilityExport extends CapabilityCfg {
   bindings: Binding[];
   hasSchedule: boolean;
+  apiGroups?: string[];
```
cmwylie19: shouldn't this be RBAC and maybe an array of PolicyRules?

cmwylie19: Yeah i think it actually should since there can be more than one rule set

schaeferka: Updated to:

```ts
export interface CapabilityExport extends CapabilityCfg {
  bindings: Binding[];
  hasSchedule: boolean;
  rbac?: PolicyRule[];
}
```

src/lib/module.ts
schaeferka: Removed the Rule interface and updated to import and use PolicyRule.

src/lib/assets/index.ts
```diff
@@ -123,10 +130,12 @@ export class Assets {
...
-      await fs.writeFile(
-        clusterRolePath,
-        dumpYaml(clusterRole(this.name, this.capabilities, "rbac"), { noRefs: true }),
-      );
+      // change this to a helm chart template
+      await fs.writeFile(clusterRolePath, dedent(clusterRoleTemplate()));
+      // await fs.writeFile(
+      //   clusterRolePath,
+      //   dumpYaml(clusterRole(this.name, this.capabilities, "rbac"), { noRefs: true }),
+      // );
```
cmwylie19:
```suggestion
```

cmwylie19: remove comment

src/lib/assets/rbac.ts
```diff
@@ -109,3 +166,25 @@ export function storeRoleBinding(name: string): kind.RoleBinding {
     ],
   };
 }
+
+/**
+ * Reads and parses RBAC information from the package.json file under the 'pepr' key.
+ *
+ * @returns {RBACMap | null} - An object representing the RBAC rules from package.json or null if not found.
+ */
+export const readRBACFromPackageJson = (): RBACMap | null => {
```
cmwylie19: not sure we need to read this in again, we have access to this in the function `allYaml` that calls this function, it is the assets, is there a reason for this?

schaeferka: The `readRBACFromPackageJSON` function is used to get the custom RBAC rules from the `package.json`. It is called from the `clusterRole` function in `rbac.ts`, which then merges the custom rules into the generated RBAC rules. 

Currently, we are calling the `clusterRole` function twice - once in the `overridesFile` function and once in the `allYaml` function. The `assets/index.ts` file calls `overridesFile` and `allYaml` separately. But, as part of the `yaml.ts` refactoring we could improve this so that `clusterRole` is only called once. For example, in `assets/index.ts` we could call `clusterRole` once to get the merged list of rbac rules, and then hand that list to both `overrides` and `allYaml`.

src/lib/module.ts
```diff
@@ -35,6 +36,10 @@ export type ModuleConfig = {
   env?: Record<string, string>;
   /** Custom Labels for Kubernetes Objects */
   customLabels?: CustomLabels;
+  /** Custom RBAC rules */
+  rbac?: PolicyRule[];
+  /** The RBAC mode; if "scoped", generates scoped rules, otherwise uses wildcard rules. */
+  rbacMode?: string;
```
cmwylie19, line 42: This is out of scope for this issue, better scoped for a new issue

cmwylie19: Going to go ahead and prioritize issue https://github.com/defenseunicorns/pepr/issues/1327 so we can get this merged.

cmwylie19 (dismissed):

src/lib/assets/rbac.ts
cmwylie19: glad to see we are only reading the module config once again.

journey/pepr-build-wasm.ts
```diff
@@ -12,13 +12,15 @@ import { cwd } from "./entrypoint.test";
 // test npx pepr build -o dst
 const outputDir = "dist/pepr-test-module/child/folder";
 export function peprBuild() {
-
   it("should build artifacts in the dst folder", async () => {
-    await fs.mkdir(outputDir, { recursive: true })
+    await fs.mkdir(outputDir, { recursive: true });
   });
```
samayer12, line 17: This test doesn't actually assert anything, it just performs a `mkdir` operation without an `expect()` statement to ensure success. We should fix that. We should also perform test cleanup (i.e., removing the directory) after the test finishes.

Not sure if this will cause issues with the follow-on tests, since what we have in `journey/` usually aren't isolated from each other. Possibly out-of-scope for this issue and deserves a follow-on tech debt chore.

src/lib/assets/deploy.ts
```diff
@@ -84,18 +85,24 @@ export async function deploy(assets: Assets, force: boolean, webhookTimeout?: nu
     throw new Error("No code provided");
   }
 
-  await setupRBAC(name, assets.capabilities, force);
+  await setupRBAC(name, assets.capabilities, force, assets.config.rbacMode, assets.config.rbac);
```
samayer12: Why do we destructure the `assets.config` object here instead of passing in `assets.config`?

samayer12 (changes requested):

src/lib/assets/rbac.test.ts
```diff
@@ -0,0 +1,689 @@
...
+              annotations: {},
+              deletionTimestamp: false,
+            },
+          },
+        ],
+        hasSchedule: false,
+        name: "",
+        description: "",
+      },
+    ];
```
samayer12, line 231: These `CapabilityExport[]` objects seems to be re-used by a couple of tests. To be more direct about which keys were toggling for particular test cases, maybe we should have some sort of generator or a `defaultCapabilityExport` that we use as a starting point. Then, in each test case, the "Arrange" step involves setting the key(s) that are relevant to the particular test case.

The same idea goes for a lot of the "expected" JSON data

src/lib/assets/rbac.ts
```diff
@@ -2,35 +2,65 @@
...
 import { CapabilityExport } from "../types";
 import { createRBACMap } from "../helpers";
+
 /**
- * Grants the controller access to cluster resources beyond the mutating webhook.
+ * Creates a Kubernetes ClusterRole based on capabilities and optional custom RBAC rules.
  *
- * @todo: should dynamically generate this based on resources used by the module. will also need to explore how this should work for multiple modules.
- * @returns
+ * @todo: should dynamically generate this based on resources used by the module. will also need to explore how this
```
samayer12: Lingering TODO comment. Either fix it or write up an issue in the backlog so we don't lose track

src/lib/assets/deploy.ts
schaeferka: Updated to pass in assets.config

src/lib/assets/rbac.test.ts
schaeferka: Made a separate issue to revisit using a generator or `defaultCapabilityExport`.

src/lib/assets/rbac.ts
schaeferka: Removed TODO

src/lib/assets/deploy.ts
samayer12: We could do either way, I was curious if there was something about splitting them up that made more sense at first

samayer12 (dismissed):

src/lib/assets/deploy.ts
```diff
@@ -4,6 +4,7 @@
 import crypto from "crypto";
 import { promises as fs } from "fs";
 import { K8s, kind } from "kubernetes-fluent-client";
+import { PolicyRule } from "kubernetes-fluent-client/dist/upstream";
```
cmwylie19:
```suggestion
import { V1PolicyRule as PolicyRule } from "@kubernetes/client-node";
```

src/lib/assets/rbac.ts
```diff
@@ -2,35 +2,63 @@
 // SPDX-FileCopyrightText: 2023-Present The Pepr Authors
 
 import { kind } from "kubernetes-fluent-client";
+import { PolicyRule } from "kubernetes-fluent-client/dist/upstream";
```
cmwylie19:
```suggestion
import { V1PolicyRule as PolicyRule } from "@kubernetes/client-node";
```

src/lib/assets/rbac.test.ts
```diff
@@ -0,0 +1,689 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2023-Present The Pepr Authors
+import { clusterRole, clusterRoleBinding, storeRole, serviceAccount, storeRoleBinding } from "./rbac";
+import { CapabilityExport } from "../types";
+import { it, describe, expect, beforeEach, jest } from "@jest/globals";
+import { GenericClass } from "kubernetes-fluent-client";
+import { PolicyRule } from "kubernetes-fluent-client/dist/upstream";
```
cmwylie19:
```suggestion
import { V1PolicyRule as PolicyRule } from "@kubernetes/client-node";
```

cmwylie19 (changes requested):
need just one second on this.

src/lib/types.ts
```diff
@@ -9,6 +9,7 @@ import { PeprMutateRequest } from "./mutate-request";
 import { PeprValidateRequest } from "./validate-request";
 
 import { Logger } from "pino";
+import { PolicyRule } from "kubernetes-fluent-client/dist/upstream";
```
cmwylie19:
```suggestion
import { V1PolicyRule as PolicyRule } from "@kubernetes/client-node";
```

src/lib/module.ts
```diff
@@ -8,6 +8,7 @@ import { MutateResponse, ValidateResponse, WebhookIgnore } from "./k8s";
 import { CapabilityExport, AdmissionRequest } from "./types";
 import { setupWatch } from "./watch-processor";
 import { Log } from "../lib";
+import { PolicyRule } from "kubernetes-fluent-client/dist/upstream";
```
cmwylie19:
```suggestion
import { V1PolicyRule as PolicyRule } from "@kubernetes/client-node";
```

cmwylie19 (dismissed):

cmwylie19:
The runners are not refreshing the code, the complaints are due to a change in KFC that was needed to unblock CI but no longer exists in this code base.

PR #1259, opened 2024-10-10
chore: refactor store code

### LINKED ISSUE #959: refactor the store to share code

The controller Store has several nested function that are extremely similar, I think we can refactor the store code and reduce a lot of duplicated code. 

https://github.com/defenseunicorns/pepr/blob/main/src/lib/controller/store.ts

### Additional context

samayer12:
## Description

Refactor the `store.ts` file.

End to End Test:  

## Related Issue

Fixes #959 

Relates to #

## Type of change

## Checklist before merging


src/lib/controller/store.test.ts
```diff
@@ -1,131 +1,73 @@
...
     });
-  });
-  test("should not redact keys and values if PEPR_STORE_REDACT_VALUES is not true", () => {
-    addOperationKeys.forEach((key, i) => {
-      const redactedResult = redactedPatch({ [key]: addOperationValues[i] });
-      for (const [k, v] of Object.entries(redactedResult)) {
-        expect(k).not.toContain(`:${redactedValue}`);
-        expect(v).not.toEqual(expect.objectContaining({ value: redactedValue }));
-      }
+    it.skip("should migrate and setup the watch (with schedule)", async () => {
```
schaeferka, line 56:
```suggestion
    it("should migrate and setup the watch (with schedule)", async () => {
```

src/lib/controller/store.test.ts
```diff
@@ -1,131 +1,73 @@
...
+      testCapability.OnSchedule(mockSchedule);
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      const peprControllerStore = new PeprControllerStore([testCapability], `pepr-test-schedule`, () => {});
+      jest.advanceTimersToNextTimer();
+      await Promise.resolve();
+      expect(true).toBe(false); // store.ts only exposes a constructor, hard to test
+    });
+
+    it.skip("should create the store resource for a scheduled capability (without schedule)", async () => {
```
schaeferka, line 65:
```suggestion
    it("should create the store resource for a scheduled capability (without schedule)", async () => {
```

schaeferka (commented):
I wasn't sure if you meant to leave the "skip" tests in, so I left a comment on them.

A couple of months ago we had a discussion about fuzz testing. I'm not sure if it relates to the fuzzing items you removed, but it may be worth double checking.

src/lib/controller/store.test.ts
samayer12: This test doesn't do anything meaningful and is meant as a starting point for #1323.

src/lib/controller/store.test.ts
samayer12: This test doesn't do anything meaningful and is meant as a starting point for #1323.

samayer12:
> A couple of months ago we had a discussion about fuzz testing. I'm not sure if it relates to the fuzzing items you removed, but it may be worth double checking.

@schaeferka, the fuzzing that is on `main` didn't seem to provide any value to the unit tests for the refactored `logger.test.ts` test cases so I removed it. It looks like it was used to provide input to test against the "negative" case (i.e., should not redact) and I think the current tests cover that.

src/lib/capability.test.ts
```diff
@@ -100,11 +100,11 @@ describe("Capability", () => {
     const capability = new Capability(capabilityConfig);
 
     const storeResult = capability.registerStore();
-    expect(storeResult).toHaveProperty("store");
+    expect(storeResult).toBeDefined();
     expect(mockLog.info).toHaveBeenCalledWith(`Registering store for ${capabilityConfig.name}`);
 
     const scheduleStoreResult = capability.registerScheduleStore();
-    expect(scheduleStoreResult).toHaveProperty("scheduleStore");
+    expect(scheduleStoreResult.onReady).toBeDefined;
```
btlghrants:
```suggestion
    expect(scheduleStoreResult.onReady).toBeDefined();
```
Missing the invocation.

src/lib/controller/store.ts
```diff
@@ -3,14 +3,14 @@
 
 import { Operation } from "fast-json-patch";
 import { K8s } from "kubernetes-fluent-client";
-import { startsWith } from "ramda";
+import * as ramda from "ramda";
```
btlghrants: Would recommend moving this back to the specific-import syntax.

By using the "* as <lib>" syntax you're preventing build-time [tree shaking](https://developer.mozilla.org/en-US/docs/Glossary/Tree_shaking) from being able to remove all the (in this case) `ramda` functions you _aren't_ calling from the final distributable (which just increases `/dist` folder / module size / etc. for no benefit).

Because of that we try to use the "* as <lib>" syntax only inside test code where build output size doesn't matter.

src/lib/logger.ts
btlghrants, line 1: It makes sense to me that we define `redactedValue` here -- since it's a constant that we want to present in a uniform way across all outputs to the logs... but it makes less sense to me for the `logger.ts` to take a dependency on the PeprStore.  What's the reasoning there? :thinking: 

Any chance we could just export `redactedValue` from this file and put the functions that required knowledge of the `Store` back into the store files?

src/lib/controller/storeCache.test.ts
```diff
@@ -0,0 +1,101 @@
...
+        } as unknown as K8sInit<T, K>;
+      });
+
+      const input: Record<string, Operation> = {
+        entry: { op: "remove", path: "/some/path" },
+        entry2: { op: "add", path: "some/path", value: "value" },
+      };
+      const result = await sendUpdatesAndFlushCache(input, "some namespace", "some name");
+      expect(result).toStrictEqual({});
+    });
```
btlghrants, line 51:
```suggestion
    });

```

NBD.  Just missing a line break between tests.

btlghrants (changes requested):

src/lib/capability.test.ts
samayer12: Fixed

src/lib/controller/store.ts
samayer12: Fixed

src/lib/controller/storeCache.test.ts
samayer12: Fixed, figured the linter would've caught that :/

src/lib/logger.ts
samayer12: Good callout! I think this actually gets back to some issues with naming in the project. `store.ts` contains `PeprControllerStore` and both files depend upon `PeprStore`, which is defined in `k8s` (not to be confused with `K8s` from KFC).

That highlights a pattern of needlessly prefixing names that are used internal to `Pepr` with `pepr`. This is an issue because the naming can be redundant, makes maintenance harder (yep!), isn't specific, and it also goes against DDD principles that communicate intent. Not to mention it's just a poor abstraction that can lend itself to creating catch-all classes.

I've renamed some classes for clarity, which I think gets after the core issue. There's also a larger discussion about how we log in general, but that's out-of-scope for this PR.

btlghrants: Agree on that there's probably a larger discussion to be had.

I think I'm still concerned about the fact that we now have a `logger.ts` file -- that to this point has been purely concerned with configuring the Pino logger.  Now, it contains refs to `process.env.PEPR_STORE_REDACT_VALUES` knows how to iterate the store & understand stuff like what `redactedCache[targetKey] = redactedOperation;` means.

While it's not necessarily out of control (cuz all that stuff is hidden inside function bodies) but it does feels like the `redactedPatch()` and 'redactedStore()` methods don't really belong to a `logger` module.  I wouldn't really think to look there personally, and if I was cold reading the source seeing that import would surprise me.  Maybe I'm the oddball here?

Cuz the real question is, "If this `logger` module should know how to situationally and _specifically_ hide Store outputs, what_ are the limits to what this `logger` module should know how to log?".  Is it reasonable to have a `logger` module that knows the ins & outs of how every other object in the project needs it's information hidden, or not?  Feels like that logic would be better placed closer to the domain objects that will already have an idea of how to distinguish between secret vs. not.

All that said, since these already _are_ functions that can be easily moved later, I'm not going to hold up this PR (since it technically works).  To prevent a low-cohesion, "uber-logger" from arising out of this in the future though, we should probably have a team convo about this at some point.

btlghrants (dismissed):
Seems reasonable. :+1:

src/lib/logger.ts
samayer12: Yep, I agree 100%. Since our logging is pretty basic at-present and the redaction functions are easy to move, let's leave it for now. I'll put it on the schedule for us to discuss sometime.

btlghrants (approved):
LGTM

PR #1321, opened 2024-10-23
feat: add ability to exit Finalize() callback WITHOUT removing the finalizer

### LINKED ISSUE #1316: feat: add ability to exit Finalize() callback WITHOUT removing the finalizer
### Def o' Done ###

```
Given an update to a resource that already has a `deletionTimestamp` field
When the `Finalize()` callback is called
Then a Module Author should be able to opt-out of removing that resource's finalizer by returning _false_
And a Module Author should be able to allow the existing finalizer-removing behavior by returning _nothing or true_.
```

### Background ###

The current implementation of Pepr Finalizer's has the Finalize callback being invoked _any time there is an update to the finalized resource after a deletionTimestamp is added_.  Also, immediately after that callback returns (whether by error or natural completion) the finalizer metadata property is removed... and the resource is deleted from the cluster.

This means that there is no way for a Module Author to 1) prevent the Finalizer callback from being invoked (because that's written into Pepr), or 2) write guard logic inside the Finalize callback to effectively "postpone" execution / removal of the metadata finalizer.

This leaves them in the position of not being able to handle updates to the resource once deletion has started (i.e. a deletionTimestamp exists) -- like, to add a `Status: Removing` field -- and no way to opt-out of finalizer removal, nor _opt-in_ to re-running Finalizer callback in the case of error (etc.).

btlghrants:
## Description

Adds ability for Module Author to "opt out" of removing Pepr's metadata.finalizers entry.

End to End Test:    
[pepr-excellent-examples #172](https://github.com/defenseunicorns/pepr-excellent-examples/pull/172)

## Related Issue

Fixes #1316 

## Type of change

## Checklist before merging

### DIFF
--- a/src/lib/watch-processor.ts
+++ b/src/lib/watch-processor.ts
@@ -120,7 +120,7 @@ async function runBinding(binding: Binding, capabilityNamespaces: string[], igno
               // [ true, void, undefined ] SHOULD remove finalizer
               // [ false ] should NOT remove finalizer
               shouldRemoveFinalizer === false
-                ? Log.debug({ obj }, `Skip removal of finalizer '${peprFinal}' from '${resource}'`)
+                ? Log.debug({ obj }, `Skipping removal of finalizer '${peprFinal}' from '${resource}'`)
                 : await removeFinalizer(binding, obj);
             }
           } else {
--- a/src/lib/watch-processor.ts
+++ b/src/lib/watch-processor.ts
@@ -107,6 +107,7 @@ async function runBinding(binding: Binding, capabilityNamespaces: string[], igno
               return;
             }
             try {
+              // TODO
               await binding.finalizeCallback?.(obj);
 
               // irrespective of callback success / failure, remove pepr finalizer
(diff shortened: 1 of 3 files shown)


src/lib/watch-processor.ts
```diff
@@ -106,12 +106,22 @@ async function runBinding(binding: Binding, capabilityNamespaces: string[], igno
...
             } finally {
-              await removeFinalizer(binding, obj);
+              const peprFinal = "pepr.dev/finalizer";
+              const meta = obj.metadata!;
+              const resource = `${meta.namespace || "ClusterScoped"}/${meta.name}`;
+
+              // [ true, void, undefined ] SHOULD remove finalizer
+              // [ false ] should NOT remove finalizer
+              shouldRemoveFinalizer === false
+                ? Log.debug({ obj }, `Skip removal of finalizer '${peprFinal}' from '${resource}'`)
```
schaeferka:
```suggestion
                ? Log.debug({ obj }, `Skipping removal of finalizer '${peprFinal}' from '${resource}'`)
```

schaeferka: Don't most of our log lines use a -ing verb? If so, we may want to make this "Skipping" instead of "Skip"

cmwylie19 (approved):

PR #1326, opened 2024-10-23
chore: fix circular dependency between types and mutate-request

### LINKED ISSUE #1244: Remove circular dependencies

Pepr has 18 circular dependencies. These can be the source of runtime errors, initialization order issues, and it complicates code structure. Furthermore, this makes things difficult to debug and can be the source of memory leaks or inconsistent behavior across environments.

We should restructure the code so that these circular dependencies are removed.

### Definition of Done

* There are less than 18 circular dependencies in the project.
* The tolerance threshold in CI/CD's `Circular Dependency Check` step is lowered.

Circular dependencies can be addressed over the course of several PRs if necessary, you do not need to resolve all of them at once.

### Additional context
Calculate the current amount of circular dependencies with: `npx madge -c --extensions ts src/`

Related issue to this broader issue are: 

```[tasklist]
### Tasks
- [ ] #1623
- [ ] #1624
- [ ] #1625
- [ ] #1626
- [ ] #1627
- [ ] #1628
- [ ] #1629
- [ ] #1630
- [ ] #1631
- [ ] #1632
- [ ] #1633
```

schaeferka:
## Description
We would like to remove the circular dependency between types.ts and mutate-request.ts
...

End to End Test:    

## Related Issue

Fixes #

Relates to #

## Type of change

## Checklist before merging


.github/workflows/dependency-review.yml
```diff
@@ -35,11 +35,11 @@ jobs:
...
           tail -n +4 tmp.log > circular-deps.log
-          if [ $(wc -l < circular-deps.log) -gt 18 ]; then
-            echo "circular-deps.log has more than 18 circular dependencies."
+          if [ $(wc -l < circular-deps.log) -gt 17 ]; then
+            echo "circular-deps.log has more than 17 circular dependencies."
             wc -l circular-deps.log
             exit 1
           else
-            echo "circular-deps.log has 18 or fewer circular dependencies."
+            echo "circular-deps.log has 17 or fewer circular dependencies."
```
samayer12, line 43: Lowering the tolerance threshold is something we need to do in addition to resolving the circular dep, as shown here.

src/lib/mutate-request.ts
```diff
@@ -1,151 +1,10 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
```
samayer12: We shouldn't disable linter rules unless we have a really good reason. It looks like we can leave `PeprMutateRequest` and type `Mutate Action` in `mutate-request.ts` and extract the rest to `mutate-types.ts` as already shown. Let's see if that works before turning off that check.

src/lib/mutate-types.ts
```diff
@@ -0,0 +1,137 @@
...
+};
+
+// MutateAction type for handling mutation callbacks
+export type MutateAction<T extends GenericClass, K extends KubernetesObject = InstanceType<T>> = (
+  req: PeprMutateRequest<K>,
+  logger?: Logger,
+) => Promise<void> | void | Promise<PeprMutateRequest<K>> | PeprMutateRequest<K>;
+
+// PeprMutateRequest class for mutation request handling
+export class PeprMutateRequest<T extends KubernetesObject> {
```
samayer12: See other comment about leaving these in `mutate-request.ts`.

src/lib/types.ts
```diff
@@ -2,20 +2,21 @@
...
+import { Operation } from "./mutate-types";
 import { WatchPhase } from "kubernetes-fluent-client/dist/fluent/types";
+import { Logger } from "pino";
 
-import { PeprMutateRequest } from "./mutate-request";
+import { PeprMutateRequest } from "./mutate-types";
 import { PeprValidateRequest } from "./validate-request";
 
-import { Logger } from "pino";
+//import { Logger } from "pino";
```
samayer12: Delete commented blocks

samayer12 (changes requested):

.github/workflows/dependency-review.yml
schaeferka: I'll make sure to lower the tolerance on future fixes

src/lib/types.ts
schaeferka: Done

src/lib/mutate-request.ts
schaeferka: Fixed with the changes you suggested

src/lib/mutate-types.ts
schaeferka: Fixed with the changes you suggested

schaeferka:
Closed and replaced with new PR to fix CI issue

PR #1337, opened 2024-10-25
chore: circular dependency between types.ts and validate-request.ts

### LINKED ISSUE #1336: chore: fix circular dependency between types and validate-request
We need to fix the circular dependency between `types.ts` and `validate-request.ts`

schaeferka:
## Description
We need to fix the circular dependency between types.ts and validate-request.ts
...

End to End Test:    

## Related Issue

Fixes #

Relates to #

## Type of change

## Checklist before merging


src/lib/validate-types.ts
```diff
@@ -0,0 +1,100 @@
...
+
+/**
+ * GroupVersionResource unambiguously identifies a resource. It doesn't anonymously include GroupVersion
+ * to avoid automatic coercion. It doesn't use a GroupVersion to avoid custom marshalling
+ */
+export interface GroupVersionResource {
+  readonly group: string;
+  readonly version: string;
+  readonly resource: string;
+}
```
cmwylie19: Do generic kinds needed by all the app need to go to a validate types? This is a little confusing to me. I see mutate processor pulling from validate-types too

schaeferka: I had moved them initially to fix the circular dependency. But now that I've made some other changes, I'll go back to revisit the initial changes to see if I can move things back to types.ts so that both mutate and validate can import them from there.

cmwylie19: bloating our softwares size is going to cause even more memory problems, this is not the approach we should take when fixing circular dependencies because we are not introducing more problems. We should fix the code that was already merged in that duplicated types

cmwylie19 (changes requested):
code is duplicated across types, leading to increase in memory and storage size. Increasing bloat of Pepr is going to cause more problems and actually introduce _more_ dependencies since the code is now dependent on the duplicate type. This is an anti-pattern

schaeferka:
Replaced by:

chore: fix circular dependencies related to types.ts #1339

PR #1255, opened 2024-10-10
fix: eslint max depth warnings

tamirazrab:
## Description
Fixes eslint `max-depth` warnings in different files. 

Left out` helper.ts` purposely not sure how to remove nesting there.

## Related Issue

Fixes #1249 

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Other (security config, docs update, etc)

## Checklist before merging
- [X] Unit, [Journey](https://github.com/defenseunicorns/pepr/tree/main/journey), [E2E Tests](https://github.com/defenseunicorns/pepr-excellent-examples), [docs](https://github.com/defenseunicorns/pepr/tree/main/docs), [adr](https://github.com/defenseunicorns/pepr/tree/main/adr) added or updated as needed
- [X] [Contributor Guide Steps](https://docs.pepr.dev/main/contribute/#submitting-a-pull-request) followed

PR #1365, opened 2024-10-31
Revert "fix: eslint max depth warnings"

samayer12:
Reverts defenseunicorns/pepr#1255


samayer12:
There were two failing E2E tests that we need to handled before #1255 is actually good to go.

cmwylie19 (approved):

Issue #1248, opened 2024-10-09
Resolve eslint `complexity` warnings


samayer12:
### Describe what should be investigated or refactored

ESlint has several warnings for [complexity](https://eslint.org/docs/latest/rules/complexity) violations across the codebase. Resolve these warnings.

This is a large issue that will require multiple PRs to address. Contributors are welcome to open up a smaller, targeted issue and resolve a single complexity warning to keep the size of work to a smaller chunk. We'll use cross-references to this issue to correlate changes.

### Definition of Done

* There are zero eslint warnings for `complexity`
* `complexity` is set to `error` from `warn` in `.eslintrc.json`

### Additional context

Run the linter with `npm format:check`

PR #1497, opened 2024-12-03
refactor: resolve eslint warnings (max-statements, complexity) - `src/lib/assets/index.ts`

btlghrants:

Refactor of `src/lib/assets/index.ts` to reduce complexity warnings:
```
/pepr/src/lib/assets/index.ts
  82:23  warning  Async method 'generateHelmChart' has a complexity of 12. Maximum allowed is 10        complexity
  82:23  warning  Async method 'generateHelmChart' has too many statements (58). Maximum allowed is 20  max-statements
```

### Additional context
In support of #1248

### DIFF
--- a/src/lib/assets/index.ts
+++ b/src/lib/assets/index.ts
@@ -35,11 +35,11 @@ function toYaml(obj: any): string {
   return dumpYaml(obj, { noRefs: true });
 }
 
-function webhookYaml(
+function createWebhookYaml(
   assets: Assets,
-  whc: kind.MutatingWebhookConfiguration | kind.ValidatingWebhookConfiguration,
+  webhookConfiguration: kind.MutatingWebhookConfiguration | kind.ValidatingWebhookConfiguration,
 ): string {
-  const yaml = toYaml(whc);
+  const yaml = toYaml(webhookConfiguration);
   return replaceString(
     replaceString(
       replaceString(yaml, assets.name, "{{ .Values.uuid }}"),
@@ -184,11 +184,11 @@ export class Assets {
       }
 
       if (mutateWebhook) {
-        await fs.writeFile(helm.files.mutationWebhookYaml, webhookYaml(this, mutateWebhook));
+        await fs.writeFile(helm.files.mutationWebhookYaml, createWebhookYaml(this, mutateWebhook));
       }
 
       if (validateWebhook) {
-        await fs.writeFile(helm.files.validationWebhookYaml, webhookYaml(this, validateWebhook));
+        await fs.writeFile(helm.files.validationWebhookYaml, createWebhookYaml(this, validateWebhook));
       }
 
       const watchDeployment = watcher(this, this.hash, this.buildTimestamp);
--- a/src/lib/assets/index.ts
+++ b/src/lib/assets/index.ts
@@ -80,69 +80,79 @@ export class Assets {
   };
 
   generateHelmChart = async (basePath: string) => {
-    //TODO
-    const CHART_DIR = `${basePath}/${this.config.uuid}-chart`;
-    const CHAR_TEMPLATES_DIR = `${CHART_DIR}/templates`;
-    const valuesPath = resolve(CHART_DIR, `values.yaml`);
-    const chartPath = resolve(CHART_DIR, `Chart.yaml`);
-    const nsPath = resolve(CHAR_TEMPLATES_DIR, `namespace.yaml`);
-    const watcherSVCPath = resolve(CHAR_TEMPLATES_DIR, `watcher-service.yaml`);
-    const admissionSVCPath = resolve(CHAR_TEMPLATES_DIR, `admission-service.yaml`);
-    const mutationWebhookPath = resolve(CHAR_TEMPLATES_DIR, `mutation-webhook.yaml`);
-    const validationWebhookPath = resolve(CHAR_TEMPLATES_DIR, `validation-webhook.yaml`);
-    const admissionDeployPath = resolve(CHAR_TEMPLATES_DIR, `admission-deployment.yaml`);
-    const admissionServiceMonitorPath = resolve(CHAR_TEMPLATES_DIR, `admission-service-monitor.yaml`);
-    const watcherDeployPath = resolve(CHAR_TEMPLATES_DIR, `watcher-deployment.yaml`);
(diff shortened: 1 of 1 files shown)


btlghrants:
Got all the complexity warnings fixed and shrunk number of statements to _just one more than the limit_ where I then stopped because any further statement squashing would have made the function more difficult to grok.

This method has the (rather large) responsibility for laying out files on disk in a format that it doesn't own -- a helm chart's expected structure -- and that feels rather larger in scope than the `max-statements` eslint rule was intended to apply to, so... after I shrank the function as much as I thought was advisable I added a comment-based config to tell eslint that this method is (likely) as minimal as it's going to get.

src/lib/assets/index.ts
```diff
@@ -28,6 +29,68 @@ import { watcher, moduleSecret } from "./pods";
 
 import { clusterRoleBinding, serviceAccount, storeRole, storeRoleBinding } from "./rbac";
 import { createDirectoryIfNotExists } from "../filesystemService";
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
samayer12, line 33: Was this added intentionally or as a stopgap while working on the complexity warnings?

src/lib/assets/index.ts
```diff
@@ -28,6 +29,68 @@ import { watcher, moduleSecret } from "./pods";
...
+function toYaml(obj: any): string {
+  return dumpYaml(obj, { noRefs: true });
+}
+
+function webhookYaml(
```
samayer12:
```suggestion
function createWebhookYaml(
```

This function reads less like a property name if we change the function name, which aids clarity in my opinion

src/lib/assets/index.ts
```diff
@@ -28,6 +29,68 @@ import { watcher, moduleSecret } from "./pods";
...
+  assets: Assets,
+  whc: kind.MutatingWebhookConfiguration | kind.ValidatingWebhookConfiguration,
```
samayer12:
```suggestion
  webhookConfiguration: kind.MutatingWebhookConfiguration | kind.ValidatingWebhookConfiguration,
```

Abbreviated parameter names are worth avoiding, even though one could reasonably infer the meaning of `whc` thanks to the typing provided.

src/lib/assets/index.ts
```diff
@@ -28,6 +29,68 @@ import { watcher, moduleSecret } from "./pods";
...
+): string {
+  const yaml = toYaml(whc);
```
samayer12:
```suggestion
  const yaml = toYaml(webhookConfiguration);
```

See earlier comment about abbreviated variable names

src/lib/assets/index.ts
```diff
@@ -28,6 +29,68 @@ import { watcher, moduleSecret } from "./pods";
...
+  return replaceString(
+    replaceString(
+      replaceString(yaml, assets.name, "{{ .Values.uuid }}"),
+      assets.config.onError === "reject" ? "Fail" : "Ignore",
+      "{{ .Values.admission.failurePolicy }}",
+    ),
+    `${assets.config.webhookTimeout}` || "10",
+    "{{ .Values.admission.webhookTimeout }}",
+  );
```
samayer12, line 51: We call `replaceString(replaceString(replaceString()))`. However, this was the original implementation, just wanted to call it out here since it looks odd at first-glance

src/lib/assets/index.ts
```diff
@@ -79,102 +142,59 @@ export class Assets {
...
-          replaceString(
-            replaceString(yamlMutateWebhook, this.name, "{{ .Values.uuid }}"),
-            this.config.onError === "reject" ? "Fail" : "Ignore",
-            "{{ .Values.admission.failurePolicy }}",
-          ),
-          `${this.config.webhookTimeout}` || "10",
-          "{{ .Values.admission.webhookTimeout }}",
-        );
-        await fs.writeFile(mutationWebhookPath, mutateWebhookTemplate);
+        await fs.writeFile(helm.files.mutationWebhookYaml, webhookYaml(this, mutateWebhook));
```
samayer12:
```suggestion
        await fs.writeFile(helm.files.mutationWebhookYaml, createWebhookYaml(this, mutateWebhook));
```

src/lib/assets/index.ts
```diff
@@ -79,102 +142,59 @@ export class Assets {
...
-          replaceString(
-            replaceString(yamlValidateWebhook, this.name, "{{ .Values.uuid }}"),
-            this.config.onError === "reject" ? "Fail" : "Ignore",
-            "{{ .Values.admission.failurePolicy }}",
-          ),
-          `${this.config.webhookTimeout}` || "10",
-          "{{ .Values.admission.webhookTimeout }}",
-        );
-        await fs.writeFile(validationWebhookPath, validateWebhookTemplate);
+        await fs.writeFile(helm.files.validationWebhookYaml, webhookYaml(this, validateWebhook));
```
samayer12:
```suggestion
        await fs.writeFile(helm.files.validationWebhookYaml, createWebhookYaml(this, validateWebhook));
```

src/lib/assets/index.ts
btlghrants: This was intentional.

This function has the responsibility to take **any** (legit!) JS object -- regardless of type -- and pass it through to a yaml serializer for stringifying; doesn't do any validations / property accesses and so shouldn't really have anything to say about the types passed in, IMO.

src/lib/assets/index.ts
btlghrants: Yep, I'm not crazy about the layout here either, really -- I personally prefer L-to-R readability (i.e. ramda.pipe() style) -- but there's nothing technically _wrong_ with the nested functional calls here (which have to be read inside-out rather than L-to-R) so I opted not to touch it. :man_shrugging:

src/lib/assets/index.ts
btlghrants: `Abbreviated parameter names are worth avoiding`

Inside small-ish functions (<100 lines) with well-defined types, I haven't found this to be particularly valuable -- I actually find expansive variable names that simply restate types to be rather _harder_ to scan through, honestly.

But willing to go on an adventure with you on this one. :man_shrugging:

samayer12 (approved):
Discussion points have been resolved, good work on addressing some complex code!

PR #1528, opened 2024-12-06
chore: returns on utils,queue,cosign

### LINKED ISSUE #1526: missing return statements on queue, utils, cosign

As we create uniforn typing standards across our projects, we need to return types on queue, cosign, and utils

cmwylie19:
## Description

As we continue to strictly type, this is the next map of files that need return types to improve our typing system

## Related Issue

Fixes #1526 

Relates to #

## Type of change

## Checklist before merging

### DIFF
--- a/src/lib/queue.ts
+++ b/src/lib/queue.ts
@@ -29,11 +29,19 @@ export class Queue<K extends KubernetesObject> {
     this.#uid = `${Date.now()}-${randomBytes(2).toString("hex")}`;
   }
 
-  label() {
+  label(): { name: string; uid: string } {
     return { name: this.#name, uid: this.#uid };
   }
 
-  stats() {
+  stats(): {
+    queue: {
+      name: string;
+      uid: string;
+    };
+    stats: {
+      length: number;
+    };
+  } {
     return {
       queue: this.label(),
       stats: {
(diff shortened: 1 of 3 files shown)


src/lib/queue.ts
```diff
@@ -73,7 +81,7 @@ export class Queue<K extends KubernetesObject> {
    *
    * @returns A promise that resolves when the webapp is reconciled
    */
-  async #dequeue() {
+  async #dequeue(): Promise<false | undefined> {
```
samayer12, line 84: returning `false | undefined` makes me wonder about where this is consumed. There's the possibility for some unexpected behavior in conditional statements.
> `false`: Is falsy but has a defined type and value.
> `undefined`: Is also falsy but indicates the absence of a value.

Do you think this justifies a small follow-up issue to dig into how this is consumed? It's definitely out-of-scope on this PR focused on just _adding_ return types.

samayer12 (approved):
typing++, thanks!

src/lib/queue.ts
cmwylie19: as the queue is such a critical piece i definitely think it deserves a close look

cmwylie19: https://github.com/defenseunicorns/pepr/issues/1530

PR #1555, opened 2024-12-09
chore: return types on store and capability

### LINKED ISSUE #1550: return types on store.ts, capability.ts
Turn on eslint settings to keep codebase clean, maintainable and standardized.

cmwylie19:
## Description

return types on Store and Capability

## Related Issue

Fixes #1550 

Relates to #

## Type of change

## Checklist before merging

### DIFF
--- a/src/lib/capability.ts
+++ b/src/lib/capability.ts
@@ -71,7 +71,7 @@ export class Capability implements CapabilityExport {
     }
   };
 
-  public getScheduleStore() {
+  public getScheduleStore(): Storage {
     return this.#scheduleStore;
   }
 
@@ -111,19 +111,19 @@ export class Capability implements CapabilityExport {
     onReady: this.#scheduleStore.onReady,
   };
 
-  get bindings() {
+  get bindings(): Binding[] {
     return this.#bindings;
   }
 
-  get name() {
+  get name(): string {
     return this.#name;
   }
 
-  get description() {
+  get description(): string {
     return this.#description;
   }
 
-  get namespaces() {
+  get namespaces(): string[] {
     return this.#namespaces || [];
   }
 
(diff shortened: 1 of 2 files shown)


src/lib/capability.ts
```diff
@@ -207,8 +207,19 @@ export class Capability implements CapabilityExport {
...
-    const log = (message: string, cbString: string) => {
+
+    type CommonChainType = typeof commonChain;
+    type ExtendedCommonChainType = CommonChainType & {
+      Alias: (alias: string) => CommonChainType;
+      InNamespace: (...namespaces: string[]) => BindingWithName<T>;
+      InNamespaceRegex: (...namespaces: RegExp[]) => BindingWithName<T>;
+      WithName: (name: string) => BindingFilter<T>;
+      WithNameRegex: (regexName: RegExp) => BindingFilter<T>;
+      WithDeletionTimestamp: () => BindingFilter<T>;
```
samayer12, line 218: For `With*` We use Generics here because we have no guarantees for what's passed in as a function, right?

Looking at [`types.ts`](https://github.com/defenseunicorns/pepr/blob/main/src/lib/types.ts), I think this is an example where we can stop using `<T extends GenericClass>` now that Pepr has stabilized a bit. 

I think we can specify an actual type for `In*`. Since `CommonChainType` is not exported and is constrained to just `capability.ts`, I'm open to making that a separate issue if we feel it's appropriate.

samayer12 (approved):
Good PR, consider comment about use of Generics in `ExtendedCommonChainType`

src/lib/capability.ts
cmwylie19: Yep, lets look at a second issue for this to keep focused. I like the idea but changing types would be out of scope for this.

PR #1563, opened 2024-12-11
chore: different periods between send and receive

### LINKED ISSUE #1561: Investigate Store Failures that Recently Occurred in UDS Core
```plaintext
Failed to set token in store for client 'uds-core-admin-neuvector', package neuvector/neuvector
```

That's basically our catch-all for when the setItemAndWait [throws an error here](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/controllers/keycloak/client-sync.ts#L174-L182) - probably a feature/QOL request here but I think the error that we catch is essentially nothing so we don't have any more insight into what went wrong.

At least provide more insight but do a deep dive to see why the promise is rejecting. Analyze `src/lib/storage.ts` and `src/lib/controller/store.ts`

cmwylie19:
## Description

We lowered the debounce period (to one second) in which our Store Subscribers **receive** updates from the Watch on the `PeprStore` resource, meaning they can react to changes faster.

The same debounce period was responsible for **sending** JSON patches to the `PeprStore` resource. The cache that holds these updates is updated and the items that were sent are deleted after the JSON patch succeeds. This meant that if we are sending events every second, and the kube-apiserver does not process the patch in a second, then the same events will be sent over multiple intervals. It is unclear if this would make the data incorrect as it is idempotent but it certainly introduces pressure on the network which could create a fallout of other network created problems. 

This PR:
- Creates two different debounce periods, one for send and one for receive
- Sends a reason for the reject in `SetItemAndWait/RemoveItemAndWait`
- Cleans up the timeouts in `SetItemAndWait/RemoveItemAndWait`
- Adds an array item to the JSON patch to the Store to ensure a watch event occurs to resolve promises in Set/RemoveItemAndWait

e2e: https://github.com/defenseunicorns/pepr-excellent-examples/pull/204
## Related Issue

Fixes #1561

Relates to #

## Type of change

## Checklist before merging

### DIFF
--- a/src/lib/controller/store.ts
+++ b/src/lib/controller/store.ts
@@ -166,6 +166,9 @@ export class StoreController {
         metadata: {
           name: this.#name,
           namespace,
+          labels: {
+            "pepr.dev-cacheID": `${Date.now()}`,
+          },
         },
         data: {
           // JSON Patch will die if the data is empty, so we need to add a placeholder
--- a/src/lib/controller/store.ts
+++ b/src/lib/controller/store.ts
@@ -12,7 +12,8 @@ import { DataOp, DataSender, DataStore, Storage } from "../storage";
 import { fillStoreCache, sendUpdatesAndFlushCache } from "./storeCache";
 
 const namespace = "pepr-system";
-export const debounceBackoff = 1000;
+const debounceBackoffReceive = 1000;
+const debounceBackoffSend = 4000;
 
 export class StoreController {
   #name: string;
@@ -134,7 +135,7 @@ export class StoreController {
 
     // Debounce the update to 1 second to avoid multiple rapid calls
     clearTimeout(this.#sendDebounce);
-    this.#sendDebounce = setTimeout(debounced, this.#onReady ? 0 : debounceBackoff);
+    this.#sendDebounce = setTimeout(debounced, this.#onReady ? 0 : debounceBackoffReceive);
   };
 
   #send = (capabilityName: string): DataSender => {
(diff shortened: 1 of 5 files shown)


src/lib/controller/store.ts
```diff
@@ -134,7 +135,7 @@ export class StoreController {
 
     // Debounce the update to 1 second to avoid multiple rapid calls
     clearTimeout(this.#sendDebounce);
-    this.#sendDebounce = setTimeout(debounced, this.#onReady ? 0 : debounceBackoff);
+    this.#sendDebounce = setTimeout(debounced, this.#onReady ? 0 : debounceBackoffReceive);
```
btlghrants, line 147: `this.#sendDebounce` --> `debounceBackoffReceive`..?  Zat right?

cmwylie19: yep! it is addressed!

cmwylie19: thanks for the callout

samayer12 (changes requested):
Given the troubleshooting we performed yesterday, this seems like it should also come with a test in pepr-excellent-examples to verify that the store behaves as intended with many sequential writes. I think we can turn @cmwylie19's troubleshooting steps into something scripted so we can catch this issue without manual testing.

cmwylie19:
> Given the troubleshooting we performed yesterday, this seems like it should also come with a test in pepr-excellent-examples to verify that the store behaves as intended with many sequential writes. I think we can turn @cmwylie19's troubleshooting steps into something scripted so we can catch this issue without manual testing.

Here is one -- https://github.com/defenseunicorns/pepr-excellent-examples/pull/204 

It does a double `SetItemAndWait` and `RemoveItemAndWait` to ensure the promise is not rejected and the store stays idempotent

samayer12 (approved):
E2E test added, good to go!

PR #1577, opened 2024-12-13
refactor: resolve eslint warnings (max-depth, complexity) - src/cli/deploy.ts

### LINKED ISSUE #1538: complexity of deploy.ts
```bash
  27:13  warning  Async arrow function has a complexity of 17. Maximum allowed is 10        complexity
```

btlghrants:

Refactor of `src/cli/deploy.ts` to reduce complexity warnings:
```
/pepr/src/cli/deploy.ts
  28:13  warning  Async arrow function has too many statements (27). Maximum allowed is 20  max-statements
  28:13  warning  Async arrow function has a complexity of 17. Maximum allowed is 10        complexity
```

### Additional context
Fixes #1538 
In support of #1248


src/cli/deploy.ts
```diff
@@ -12,6 +12,87 @@ import { sanitizeName } from "./init/utils";
...
+  const confirm = await prompt({
+    type: "confirm",
+    name: "confirm",
+    message: "This will remove and redeploy the module. Continue?",
+  });
+
+  return confirm.confirm ? true : false;
+}
+
 export default function (program: RootCmd) {
```
cmwylie19:
```suggestion
export default function (program: RootCmd): void {
```

cmwylie19: if you would like

src/cli/deploy.ts
```diff
@@ -12,7 +12,88 @@ import { sanitizeName } from "./init/utils";
...
+  }
+  if (!details.dockerServer) {
+    missing.push("--docker-server");
+  }
+  if (!details.dockerUsername) {
+    missing.push("--docker-username");
+  }
+  if (!details.dockerPassword) {
+    missing.push("--docker-password");
+  }
```
samayer12, line 51: This addresses the complexity requirements, so it's good to go. Although, I wonder if there's a way we could rephrase this block as "IF X" instead of "IF NOT X." Not a big deal, and shouldn't hold up the PR.

samayer12 (approved):
Good refactor, breaks this gnarly block of deployment code into something much more manageable.

src/cli/deploy.ts
btlghrants: Yeah, I don't usually prefer to not use non-negations too (teehee) but I didn't have anything new to say about those values since this was a refactor and checking anything other than existence -- a.k.a. the "don't be undefined / null / empty string" of the pre-existing behavior -- felt like it'd be 1) longer, and 2) changing externally-visible behavior. :man_shrugging:

PR #1364, opened 2024-10-31
chore: encourage return types in function definitions

samayer12:
## Description

Encourage non-test code to specify return types.

## Related Issue

Hotifx

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Other (security config, docs update, etc)

## Checklist before merging
- [x] Unit, [Journey](https://github.com/defenseunicorns/pepr/tree/main/journey), [E2E Tests](https://github.com/defenseunicorns/pepr-excellent-examples), [docs](https://github.com/defenseunicorns/pepr/tree/main/docs), [adr](https://github.com/defenseunicorns/pepr/tree/main/adr) added or updated as needed
- [x] [Contributor Guide Steps](https://docs.pepr.dev/main/contribute/#submitting-a-pull-request) followed

PR #1602, opened 2024-12-18
chore: add typing to templates.ts

samayer12:
## Description

Adds typing to `templates.ts`

End to End Test:    

## Related Issue

Relates to #1364 

## Type of change

## Checklist before merging

### DIFF
--- a/src/cli/init/enums.ts
+++ b/src/cli/init/enums.ts
@@ -1,0 +1,9 @@
+export enum RbacMode {
+  SCOPED = "scoped",
+  ADMIN = "admin",
+}
+export enum OnError {
+  AUDIT = "audit",
+  IGNORE = "ignore",
+  REJECT = "reject",
+}
--- a/src/cli/init/templates.ts
+++ b/src/cli/init/templates.ts
@@ -14,6 +14,8 @@ import settingsJSON from "../../templates/settings.json";
 import tsConfigJSON from "../../templates/tsconfig.module.json";
 import { sanitizeName } from "./utils";
 import { InitOptions } from "../types";
+import { V1PolicyRule as PolicyRule } from "@kubernetes/client-node";
+import { OnError, RbacMode } from "./enums";
 
 export const { dependencies, devDependencies, peerDependencies, scripts, version } = packageJSON;
 
@@ -26,12 +28,14 @@ type peprPackageJSON = {
     engines: { node: string };
     pepr: {
       uuid: string;
-      onError: string;
+      onError: OnError;
       webhookTimeout: number;
-      customLabels: { namespace: { "pepr.dev": string } };
-      alwaysIgnore: { namespaces: never[] };
-      includedFiles: never[];
+      customLabels: { namespace: Record<string, string> };
+      alwaysIgnore: { namespaces: string[] };
+      includedFiles: string[];
       env: object;
+      rbac: PolicyRule[];
+      rbacMode: RbacMode;
     };
     scripts: { "k3d-setup": string };
     dependencies: { pepr: string; undici: string };
@@ -76,6 +80,8 @@ export function genPkgJSON(opts: InitOptions, pgkVerOverride?: string): peprPack
       },
       includedFiles: [],
       env: pgkVerOverride ? testEnv : {},
+      rbac: [],
+      rbacMode: RbacMode.SCOPED,
     },
     scripts: {
       "k3d-setup": scripts["test:journey:k3d"],
--- a/src/cli/init/templates.ts
+++ b/src/cli/init/templates.ts
@@ -17,7 +17,31 @@ import { InitOptions } from "../types";
 
 export const { dependencies, devDependencies, peerDependencies, scripts, version } = packageJSON;
 
-export function genPkgJSON(opts: InitOptions, pgkVerOverride?: string) {
+type peprPackageJSON = {
+  data: {
+    name: string;
+    version: string;
+    description: string;
+    keywords: string[];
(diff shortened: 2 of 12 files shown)


src/cli/init/templates.ts
```diff
@@ -17,7 +17,31 @@ import { InitOptions } from "../types";
...
 
-export function genPkgJSON(opts: InitOptions, pgkVerOverride?: string) {
+type peprPackageJSON = {
+  data: {
+    name: string;
+    version: string;
+    description: string;
+    keywords: string[];
+    engines: { node: string };
+    pepr: {
```
cmwylie19, line 29: Good work, just to let you know here be dragons on this story, the implementation is good though. I am glad we are getting it strongly typed.

We are missing a few config options that can be in the pepr section of `package.json`.

check out [customization](https://docs.pepr.dev/main/user-guide/customization/). Looks like we need `rbac` (type `PolicyRule[]`) and `rbacMode`.

src/cli/init/templates.ts
```diff
@@ -17,7 +17,31 @@ import { InitOptions } from "../types";
...
+      uuid: string;
+      onError: string;
```
cmwylie19: To safeguard the user from typing anything, should this be an enum of [Errors](https://github.com/defenseunicorns/pepr/blob/dd866e5d0b6558c4720c5e82485bd62e53a6b366/src/lib/errors.ts#L4)?

src/cli/init/templates.ts
```diff
@@ -17,7 +17,31 @@ import { InitOptions } from "../types";
...
+      webhookTimeout: number;
+      customLabels: { namespace: { "pepr.dev": string } };
```
cmwylie19: I could be mistaken, but it looks like you are assigning a default value here. Should this be[`Record<string, string>`](https://github.com/defenseunicorns/pepr/blob/dd866e5d0b6558c4720c5e82485bd62e53a6b366/src/lib/core/module.ts#L14)?

src/cli/init/templates.ts
```diff
@@ -17,7 +17,31 @@ import { InitOptions } from "../types";
...
+      alwaysIgnore: { namespaces: never[] };
```
cmwylie19:
```suggestion
      alwaysIgnore: { namespaces: string[] };
```

Shouldn't this be a string array?

src/cli/init/templates.ts
```diff
@@ -17,7 +17,31 @@ import { InitOptions } from "../types";
...
+      includedFiles: never[];
```
cmwylie19:
```suggestion
      includedFiles: never[];
```

Should this be a string array? This is so that the package will be built with a custom file, like a wasm file so the module author can use it in the module

src/cli/init/templates.ts
samayer12: Yep!

src/cli/init/templates.ts
samayer12: Indeed it should

src/cli/init/templates.ts
samayer12: Correct, that's also consistent with the docs

src/cli/init/templates.ts
samayer12: It should be, yep!

src/cli/init/templates.ts
samayer12: Agreed, added

src/cli/init/enums.ts
samayer12, line 1: This would create two files called `enums.ts`, one in `cli` and one in `lib`. They share separate "domains", so I think thats fine?

src/cli/init/templates.ts
```diff
@@ -52,6 +80,8 @@ export function genPkgJSON(opts: InitOptions, pgkVerOverride?: string) {
       },
       includedFiles: [],
       env: pgkVerOverride ? testEnv : {},
+      rbac: [],
+      rbacMode: RbacMode.SCOPED,
```
samayer12: I lack context here, but it seems like `rbacMode` is the equivalent to setting a default value here. In that case, I'm assuming we ought to default to the use of `scoped` instead of `admin`.

samayer12:
Generally, I noticed references to `ErrorList` which feel like a complicated way to just use an `Enum`. If it doesn't take _too_ long (less than a day of effort), I'll handle that work as part of this typing refactor. That's probably something for early Jan 2025.

cmwylie19 (approved):
nice job!

src/cli/init/templates.ts
cmwylie19: ah sorry, I missed this. 

So, in short, in prod, without a doubt you are right, no reason for anyone to use cluster-admin, no way someone is going to need to control every single object in a cluster. 

There is a more philosophical conversation to be had around starting a project (and Pepr's origin), figuring out how to do the Kubernetes calls that you need, then adding the appropriate RBAC after you figure everything out.

The intent of the project is that **anyone** can build an operator or controller. We increase the barrier of entry if we expect them to know RBAC at the beginning.

Currently, after initializing a new Pepr Module, the ServiceAccount is bound to a ClusterRole that has cluster-admin privs.

If we were going to change a default behavior there would need to be a longer convo, and warnings to potential users.

cmwylie19 (changes requested):
For now lets keep RBAC mode wide open since this would be a major change. Other than that it is all good.

cmwylie19 (approved):
lgtm! Nice

PR #1617, opened 2025-01-03
feat: provides a way to override ignored namespaces at deployment time

### LINKED ISSUE #1610: Allow deploy-time configurable `alwaysIgnore.namespaces`

Currently on uds-core we set 2 namespaces to ignore by default. In some environments these namespaces should not really be ignored (one is a dev namespace) and in other namespaces additional namespaces may need to be ignored (i.e. additional "system namespaces"). Currently these are configured at build time and we have not found an obvious way to allow this to be configured at deploy time.

The current setting is templated into the webhook configuration under:
```yaml
        - key: kubernetes.io/metadata.name
          operator: NotIn
          values:
            - kube-system
            - pepr-system
```

It would be a relatively easy change to allow this to be configured in the helm chart to append to/modify the list here. In the case of uds-core we could default this list to including the dev-stack, but via overrides at deploy time it could be modified.

1. Make this the module author's responsibility - rather than filtering out with the webhook config, the module code could filter based on namespace.
2. Identify an alternative path with zarf/etc that would be viable to modify this.

cmwylie19:
## Description

In scenarios where build-time configuration is not feasible, you can set the namespaces to ignore at deploy time by using the `PEPR_IGNORED_NAMESPACES` environment variable in your deployments. However, this method bypasses the build-time safety checks and is not recommended except as a last resort. 

Needs: 1618
## Related Issue

Fixes #1610 

Relates to #1618 

## Type of change

## Checklist before merging

### DIFF
--- a/docs/030_user-guide/120_customization.md
+++ b/docs/030_user-guide/120_customization.md
@@ -66,9 +66,9 @@ In scenarios where build-time configuration is not feasible, you can set the nam
 ```
 
 ```bash
-kubectl set env deployment/pepr-admission PEPR_IGNORED_NAMESPACES=zarf,istio-system -n pepr-system
+kubectl set env deployment/pepr-admission PEPR_IGNORED_NAMESPACES="zarf,istio-system" -n pepr-system
 
-kubectl set env deployment/pepr-watcher PEPR_IGNORED_NAMESPACES=zarf,istio-system -n pepr-system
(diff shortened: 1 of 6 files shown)


cmwylie19:
`ON-HOLD` label applied as we are going through some finalizing details with our downstream stakeholder. This implementation is likely to slightly change a bit after having chatted through some specifics on how they want to use it

cmwylie19:
covered in https://github.com/defenseunicorns/pepr/pull/1641

PR #1619, opened 2025-01-03
chore: carriedNamespace/carriesIgnoredNamespace account for Namespace object

### LINKED ISSUE #1618: carriedNamespace does not account for the `Namespace` Kubernetes Object
### Environment

[carriedNamespace](https://github.com/defenseunicorns/pepr/blob/6a5abbea971e985d7a9cb67b4bab7482f67370d9/src/lib/filter/adjudicators/adjudicators.ts#L77) looks for `metadata.namespace` which does breaks when your binding is based on a Namespace 

### Steps to reproduce
```ts
When(a.Namespace)
  .IsCreated()
  .WithName("pepr-demo-2")
  .Watch(async ns => {
    Log.info("Namespace pepr-demo-2 was created.");
```

My ignored namespace is pepr-demo-2

### Expected result
The callback not to be fired.

### Actual Result
The callback is fired. 

### Additional Context

cmwylie19:
## Description

CarriedNamespace did not account for Namespace objects which caused some strange behaviors in #1610 

After Fix:

```bash
[14:20:13.943] DEBUG (44989): Ignoring Watch Callback: Object carries namespace 'pepr-demo-2' but ignored namespaces include '["pepr-demo-2"]'.
```

## Related Issue

Fixes #1618 

Relates to #

## Type of change

## Checklist before merging

### DIFF
--- a/src/lib/filter/adjudicators/adjudicators.ts
+++ b/src/lib/filter/adjudicators/adjudicators.ts
@@ -73,10 +73,14 @@ export const carriedName = pipe(
 export const carriesName = pipe(carriedName, equals(""), not);
 export const missingName = complement(carriesName);
 
-export const carriedNamespace = pipe(
-  (kubernetesObject: KubernetesObject): string | undefined => kubernetesObject?.metadata?.namespace,
-  defaultTo(""),
-);
+export const carriedNamespace = pipe((kubernetesObject: KubernetesObject): string | undefined => {
+  if (kubernetesObject?.kind === "Namespace") {
+    return kubernetesObject?.metadata?.name;
+  } else {
+    return kubernetesObject?.metadata?.namespace;
+  }
+}, defaultTo(""));
+
 export const carriesNamespace = pipe(carriedNamespace, equals(""), not);
 
 export const carriedAnnotations = pipe(
--- a/src/lib/filter/adjudicators/adjudicators.ts
+++ b/src/lib/filter/adjudicators/adjudicators.ts
@@ -73,12 +73,10 @@ export const carriedName = pipe(
 export const carriesName = pipe(carriedName, equals(""), not);
 export const missingName = complement(carriesName);
 
-export const carriedNamespace = pipe((kubernetesObject: KubernetesObject): string | undefined => {
-  if (kubernetesObject?.kind === "Namespace") {
-    return kubernetesObject?.metadata?.name;
-  }
-  return kubernetesObject?.metadata?.namespace;
-}, defaultTo(""));
+export const carriedNamespace = pipe(
+  (kubernetesObject: KubernetesObject): string | undefined => kubernetesObject?.metadata?.namespace,
+  defaultTo(""),
+);
 
 export const carriesNamespace = pipe(carriedNamespace, equals(""), not);
 
--- a/src/lib/filter/adjudicators/adjudicators.ts
+++ b/src/lib/filter/adjudicators/adjudicators.ts
@@ -76,9 +76,8 @@ export const missingName = complement(carriesName);
 export const carriedNamespace = pipe((kubernetesObject: KubernetesObject): string | undefined => {
   if (kubernetesObject?.kind === "Namespace") {
     return kubernetesObject?.metadata?.name;
-  } else {
-    return kubernetesObject?.metadata?.namespace;
   }
+  return kubernetesObject?.metadata?.namespace;
 }, defaultTo(""));
 
 export const carriesNamespace = pipe(carriedNamespace, equals(""), not);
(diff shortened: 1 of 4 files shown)


btlghrants:
Wait, but... Namespace objects _don't_ "carry" namespaces -- they have no `metadata.namespace` prop, nor should they --  because they "are" namespaces.  Namespaces _shouldn't_ carry `namespace` props, right?   So why would we want this method to present like they do?

Also, this edit makes the `carriedNamespace()` give back what `carriedName()` gives back _but only for Namespaces_ which is confusing, IMO.  Why was this update needed, exactly?

btlghrants (commented):
Wait, but... Namespace objects _don't_ "carry" namespaces -- they have no `metadata.namespace` prop, nor should they --  because they "are" namespaces.  Namespaces _shouldn't_ carry `namespace` props, right?   So why would we want this method to present like they do?

Also, this edit makes the `carriedNamespace()` give back what `carriedName()` already gives back _but only for Namespaces_ which is confusing, IMO.  Why was this update needed, exactly?

cmwylie19:
> Wait, but... Namespace objects _don't_ "carry" namespaces -- they have no `metadata.namespace` prop, nor should they -- because they "are" namespaces. Namespaces _shouldn't_ carry `namespace` props, right? So why would we want this method to present like they do?
> 
> Also, this edit makes the `carriedNamespace()` give back what `carriedName()` already gives back _but only for Namespaces_ which is confusing, IMO. Why was this update needed, exactly?

The reason this is need is due to the fact that if you are using `alwaysIgnore.namespaces`, then you should not execute any callback on anything in the namespace, including the namespace object. For instance, In the below binding I am ignoring the pepr-demo-2 `namespace`, however, the callback gets executed. Due to the currently implementation we are not ignoring the namespace that we should.

```ts
When(a.Namespace)
  .IsCreated()
  .WithName("pepr-demo-2")
  .Watch(async ns => {
    Log.info("Namespace pepr-demo-2 was created.");
```

RE- `carriedName()` - Yes- good callout - maybe we need to make carriedName check if the object is a namespace too and return the metadata.name

btlghrants (commented):
Seems like the [uncarryableNamespace()](https://github.com/defenseunicorns/pepr/blob/cd03899ee92909a0f6386ba2b572219198375825/src/lib/filter/adjudicators/adjudicators.ts#L251) func would be a better place to put the new check -- this method lives one-level above the `carriesNamespace` and `carriedNamespace` methods & is the actual method responsible for using available information to give a final determination on whether a forbidden namespace is in use or not.

If we want to update our definition of what it means to "use a forbidden namespace" from _only_ "has a forbidden metadata.namespace defined" and have it _also_ include "or IS a Namespace with a forbidden name" then I think that would be the place.

src/lib/filter/adjudicators/adjudicators.ts
```diff
@@ -73,10 +73,13 @@ export const carriedName = pipe(
 export const carriesName = pipe(carriedName, equals(""), not);
 export const missingName = complement(carriesName);
 
-export const carriedNamespace = pipe(
-  (kubernetesObject: KubernetesObject): string | undefined => kubernetesObject?.metadata?.namespace,
-  defaultTo(""),
-);
+export const carriedNamespace = pipe((kubernetesObject: KubernetesObject): string | undefined => {
```
cmwylie19: This is required to be able to update the message for `Namespace` objects

```plaintext
Ignoring Watch Callback: Object carries namespace '' but ignored namespaces include '["pepr-demo-2"]'.
```

src/lib/filter/adjudicators/adjudicators.ts
```diff
@@ -73,10 +73,13 @@ export const carriedName = pipe(
...
+  if (kubernetesObject?.kind === "Namespace") {
```
btlghrants: This looks like a hold-over from the first attempt, which... shouldn't be needed any longer, right?

Since we decided that the `carried[...]()` methods were just data accessors that _shouldn't_ behave in a conditional way & you've already moved the "is it a Namespace" checks over into `uncarryableNamespace()` and `carriesIgnoredNamespace()`..?

btlghrants (changes requested):

src/lib/filter/adjudicators/adjudicators.ts
cmwylie19: If we do not include this, then we will not have an accurate filter message

```plaintext
Ignoring Watch Callback: Object carries namespace '' but ignored namespaces include '["pepr-demo-2"]'.
```

```ts
export function adjudicateUncarryableNamespace(
  capabilityNamespaces: string[],
  obj: KubernetesObject,
): AdjudicationResult {
  return uncarryableNamespace(capabilityNamespaces, obj)
    ? `Object carries namespace '${carriedNamespace(obj)}' but namespaces allowed by Capability are '${JSON.stringify(capabilityNamespaces)}'.`
    : null;
}
```

cmwylie19: unless of course we have a 1-off check inside of `adjudicateUncarryableNamespace`, is that what you prefer here?

A check inside of that function?

cmwylie19: Okay, i accounted for it in the `adjudicate**` funcs and changed it back

btlghrants (approved):
Yep, seems like that should do it.

PR #1642, opened 2025-01-09
test: add cli test to validate multiple manifest generation paths result in comparable output

### LINKED ISSUE #1640: Ensure helm templates manifests correspond to Kubernetes manifests 1:1

We need a way to ensure that our Kubernetes and Helm templated manifests stay in synchronization. 

The task is to:
- generate a Pepr Module with all possible values in the `pepr` section of `package.json`
- Do a `helm template .` in the `dist/static-test-chart > something`
- Compare each manifest to make sure they correspond 1 to 1

### LINKED ISSUE #1654: Convert cli-test workflow into locally-runnable test harness
## Description

Introduce a new type of testing to Pepr: integration testing.  Create a place in the code base for integration testing to live & set it up such that it allows for tests that:

1. are larger in scope than unit tests but smaller in-scope than "full system" (i.e. the excellent-examples (e2es)),
2. are developer-centric (as opposed to user-centric, like the journey tests),
3. are intended to cover large-ish features comprehensively (e.g. all the pepr CLI flags),
4. are runnable independently and/or in parallel, and
5. can be run on a developer workstation.

Testing for the Pepr CLI is a great example of something that meets those conditions, so once a place for integration tests exists covert the CI-only tests in `.github/workflows/cli-tests.yml` over & then delete that workflow.   That'll demonstrate / position the project with a nice place to put many of the CLI-focused (and various other) tests that don't-quite-fit within  journey test suite as we move toward a post-#1597 world.

## Background

Pepr currently has 3 suites of tests: unit tests, "journey" tests & in-cluster end-to-end tests (i.e. the [excellent-examples](https://github.com/defenseunicorns/pepr-excellent-examples)).  The team has long recognized that, of the three, the journey test suite has (by _far_) the least well defined scope / boundaries as it tests a lot of things, like: the pepr CLI, the `hello-pepr.ts` feature-demo Module, multiple deployment mechanisms, and more.

While not "wrong" per se -- nor surprising given the organic nature of it's growth since inception -- it also isn't right: it touches too many parts of the pepr code base not-quite-comprehensively, it takes a long time to run, it's hard to run pieces independently, etc.  Issue #1597 -- which this Issue helps (partially / technically) address -- is a direct consequence of many of those not-so-ideal qualities & demonstrates the need for this work.

## Related Issue

Relates to #1597 - in support of 
Relates to #1640 - in preparation for

## Type of change

## Checklist before merging

btlghrants:
## Description

Adds a new `npm run test:int` (int == "integration") script to invoke a new style of test -- integration tests -- where "integration" means something like "wider in scope than unit tests" but "not requiring a full K8s cluster" (general idea is to have a nice place for non-k8s-required tests to land (e.g. client-side `pepr` CLI commands, like init / build).

## Related Issue

Fixes #1654 

## Type of change

## Checklist before merging


integration/cli/build.defaults.test.ts
```diff
@@ -0,0 +1,52 @@
...
+describe("build", () => {
+  const workdir = new Workdir(`${FILE}`, `${HERE}/../testroot/cli`);
+
+  beforeAll(async () => {
+    await workdir.recreate();
+  });
+
+  describe("builds a module", () => {
+    const id = FILE.split(".").at(1);
+    const mod = `${workdir.path()}/${id}`;
```
samayer12:
```suggestion
    const module = `${workdir.path()}/${id}`;
```

Avoiding abbreviations for short words is a personal style preference of mine

integration/cli/build.defaults.test.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    beforeAll(async () => {
+      await fs.rm(mod, { recursive: true, force: true });
```
samayer12:
```suggestion
      await fs.rm(module, { recursive: true, force: true });
```

integration/cli/build.defaults.test.ts
```diff
@@ -0,0 +1,52 @@
...
+      const argz = [
+        `--name ${id}`,
+        `--description ${id}`,
+        `--errorBehavior reject`,
+        "--confirm",
+        "--skip-post-init",
+      ].join(" ");
+      await pepr.cli(workdir.path(), { cmd: `pepr init ${argz}` });
+      await pepr.tgzifyModule(mod);
```
samayer12:
```suggestion
      await pepr.tgzifyModule(module);
```

integration/cli/build.defaults.test.ts
```diff
@@ -0,0 +1,52 @@
...
+      await pepr.cli(mod, { cmd: `npm install` });
```
samayer12:
```suggestion
      await pepr.cli(module, { cmd: `npm install` });
```

integration/cli/build.defaults.test.ts
```diff
@@ -0,0 +1,52 @@
...
+    }, time.toMs("2m"));
+
+    describe("using default build options", () => {
+      it(
+        "builds",
+        async () => {
+          const build = await pepr.cli(mod, { cmd: `pepr build` });
```
samayer12:
```suggestion
          const build = await pepr.cli(module, { cmd: `pepr build` });
```

integration/cli/build.help.test.ts
```diff
@@ -0,0 +1,31 @@
...
+  });
+
+  it(
+    "gives command line help",
+    async () => {
+      const argz = "--help";
+      const res = await pepr.cli(workdir.path(), { cmd: `pepr build ${argz}` });
+      expect(res.exitcode).toBe(0);
+      expect(res.stderr.join("").trim()).toBe("");
+      expect(res.stdout.at(0)).toMatch("Usage: pepr build");
```
samayer12:
```suggestion
      const result = await pepr.cli(workdir.path(), { cmd: `pepr build ${argz}` });
      expect(result.exitcode).toBe(0);
      expect(result.stderr.join("").trim()).toBe("");
      expect(result.stdout.at(0)).toMatch("Usage: pepr build");
```

Avoiding abbreviations for short works is a personal-preference of mine

integration/cli/build.noembed.test.ts
```diff
@@ -0,0 +1,93 @@
...
+        "builds",
+        async () => {
+          const argz = [`--no-embed`].join(" ");
+          const build = await pepr.cli(mod, { cmd: `pepr build ${argz}` });
+          expect(build.exitcode).toBe(0);
+
+          // TODO: team talk
+          // Should this be writing to stderr? Even with a 0 exit code..?
+          expect(build.stderr.join("").trim()).toContain("Error: Cannot find module");
+          // TODO: end
```
samayer12: Let's remove the `TODO` comment and show this as an issue in the backlog

integration/cli/build.noembed.test.ts
```diff
@@ -0,0 +1,93 @@
...
+      const argz = [
+        `--name ${id}`,
+        `--description ${id}`,
+        `--errorBehavior reject`,
+        "--confirm",
+        "--skip-post-init",
+      ].join(" ");
+      await pepr.cli(workdir.path(), { cmd: `pepr init ${argz}` });
+      await pepr.tgzifyModule(mod);
+      await pepr.cli(mod, { cmd: `npm install` });
```
samayer12:
```suggestion
      await pepr.tgzifyModule(module);
      await pepr.cli(module, { cmd: `npm install` });
```

integration/cli/build.noembed.test.ts
```diff
@@ -0,0 +1,93 @@
...
+  beforeAll(async () => {
+    await workdir.recreate();
+  });
+
+  describe("builds a module", () => {
+    const id = FILE.split(".").at(1);
+    const mod = `${workdir.path()}/${id}`;
+
+    beforeAll(async () => {
+      await fs.rm(mod, { recursive: true, force: true });
```
samayer12:
```suggestion
      await fs.rm(module, { recursive: true, force: true });
```

integration/cli/build.noembed.test.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    describe("for use as a library", () => {
+      let packageJson;
+      let uuid: string;
+
+      it(
+        "builds",
+        async () => {
+          const argz = [`--no-embed`].join(" ");
+          const build = await pepr.cli(mod, { cmd: `pepr build ${argz}` });
```
samayer12:
```suggestion
          const build = await pepr.cli(module, { cmd: `pepr build ${argz}` });
```

integration/cli/build.noembed.test.ts
```diff
@@ -0,0 +1,93 @@
...
+          expect(build.exitcode).toBe(0);
+
+          // TODO: team talk
+          // Should this be writing to stderr? Even with a 0 exit code..?
+          expect(build.stderr.join("").trim()).toContain("Error: Cannot find module");
+          // TODO: end
+
+          expect(build.stdout.join("").trim()).toContain("");
+
+          packageJson = await resource.oneFromFile(`${mod}/package.json`);
```
samayer12:
```suggestion
          packageJson = await resource.oneFromFile(`${module}/package.json`);
```

integration/cli/build.noembed.test.ts
```diff
@@ -0,0 +1,93 @@
...
+      it(
+        "outputs appropriate configuration",
+        async () => {
+          const missing = [
+            `${mod}/dist/pepr-${uuid}.js`,
+            `${mod}/dist/pepr-${uuid}.js.map`,
+            `${mod}/dist/pepr-${uuid}.js.LEGAL.txt`,
+            `${mod}/dist/pepr-module-${uuid}.yaml`,
+            `${mod}/dist/zarf.yaml`,
+            `${mod}/dist/${uuid}-chart/`,
```
samayer12:
```suggestion
            `${module}/dist/pepr-${uuid}.js`,
            `${module}/dist/pepr-${uuid}.js.map`,
            `${module}/dist/pepr-${uuid}.js.LEGAL.txt`,
            `${module}/dist/pepr-module-${uuid}.yaml`,
            `${module}/dist/zarf.yaml`,
            `${module}/dist/${uuid}-chart/`,
```

integration/cli/build.noembed.test.ts
```diff
@@ -0,0 +1,93 @@
...
+          ];
+          for (const path of missing) {
+            expect(await file.exists(path)).toBe(false);
+          }
+
+          const found = [
+            `${mod}/dist/pepr.js`,
+            `${mod}/dist/pepr.js.map`,
+            `${mod}/dist/pepr.js.LEGAL.txt`,
```
samayer12:
```suggestion
            `${module}/dist/pepr.js`,
            `${module}/dist/pepr.js.map`,
            `${module}/dist/pepr.js.LEGAL.txt`,
```

integration/cli/build.nonconflict.test.ts
```diff
@@ -0,0 +1,187 @@
...
+describe("build", () => {
+  const workdir = new Workdir(`${FILE}`, `${HERE}/../testroot/cli`);
+
+  beforeAll(async () => {
+    await workdir.recreate();
+  });
+
+  describe("builds a module", () => {
+    const id = FILE.split(".").at(1);
+    const mod = `${workdir.path()}/${id}`;
```
samayer12:
```suggestion
    const module = `${workdir.path()}/${id}`;
```

integration/cli/build.nonconflict.test.ts
```diff
@@ -0,0 +1,187 @@
...
+
+    beforeAll(async () => {
+      await fs.rm(mod, { recursive: true, force: true });
```
samayer12:
```suggestion
      await fs.rm(module, { recursive: true, force: true });
```

integration/cli/build.nonconflict.test.ts
```diff
@@ -0,0 +1,187 @@
...
+      const argz = [
+        `--name ${id}`,
+        `--description ${id}`,
+        `--errorBehavior reject`,
+        "--confirm",
+        "--skip-post-init",
+      ].join(" ");
+      await pepr.cli(workdir.path(), { cmd: `pepr init ${argz}` });
+      await pepr.tgzifyModule(mod);
```
samayer12:
```suggestion
      await pepr.tgzifyModule(module);
```

integration/cli/build.nonconflict.test.ts
```diff
@@ -0,0 +1,187 @@
...
+      await pepr.cli(mod, { cmd: `npm install` });
```
samayer12:
```suggestion
      await pepr.cli(module, { cmd: `npm install` });
```

integration/cli/build.nonconflict.test.ts
```diff
@@ -0,0 +1,187 @@
...
+    }, time.toMs("2m"));
+
+    describe("using non-conflicting build override options", () => {
+      const entryPoint = "pepr2.ts";
+      const customImage = "pepr:override";
+      const outputDir = `${mod}/out`;
```
samayer12:
```suggestion
      const outputDir = `${module}/out`;
```

integration/cli/build.nonconflict.test.ts
```diff
@@ -0,0 +1,187 @@
...
+      const withPullSecret = "shhh";
+      const zarf = "chart";
+
+      let packageJson;
+      let uuid: string;
+
+      it(
+        "builds",
+        async () => {
+          await fs.rename(`${mod}/pepr.ts`, `${mod}/${entryPoint}`);
```
samayer12:
```suggestion
          await fs.rename(`${module}/pepr.ts`, `${module}/${entryPoint}`);
```

integration/cli/build.nonconflict.test.ts
```diff
@@ -0,0 +1,187 @@
...
+
+          const argz = [
+            `--entry-point ${entryPoint}`,
+            `--custom-image ${customImage}`,
+            `--output-dir ${outputDir}`,
+            `--timeout ${timeout}`,
+            `--withPullSecret ${withPullSecret}`,
+            `--zarf ${zarf}`,
+          ].join(" ");
+          const build = await pepr.cli(mod, { cmd: `pepr build ${argz}` });
```
samayer12:
```suggestion
          const build = await pepr.cli(module, { cmd: `pepr build ${argz}` });
```

integration/cli/build.nonconflict.test.ts
```diff
@@ -0,0 +1,187 @@
...
+
+          expect(build.exitcode).toBe(0);
+          expect(build.stderr.join("").trim()).toBe("");
+          expect(build.stdout.join("").trim()).toContain("K8s resource for the module saved");
+
+          packageJson = await resource.oneFromFile(`${mod}/package.json`);
```
samayer12:
```suggestion
          packageJson = await resource.oneFromFile(`${module}/package.json`);
```

integration/cli/build.nonconflict.test.ts
```diff
@@ -0,0 +1,187 @@
...
+          uuid = packageJson.pepr.uuid;
+        },
+        time.toMs("1m"),
+      );
+
+      const getDepConImg = (deploy: kind.Deployment, container: string): string => {
+        return deploy!.spec!.template!.spec!.containers.filter(f => f.name === container).at(0)!
```
samayer12:
```suggestion
        return deploy!.spec!.template!.spec!.containers.filter(file => file.name === container).at(0)!
```

integration/cli/build.nonconflict.test.ts
```diff
@@ -0,0 +1,187 @@
...
+          .image!;
+      };
+
+      it("--entry-point, works", async () => {
+        // build would fail if given entrypoint didn't exist, so... no-op test!
+      });
```
samayer12, line 86: Is this no-op test here just to enumerate ways to use the CLI tool? Could we have it actually do something?

integration/cli/build.nonconflict.test.ts
```diff
@@ -0,0 +1,187 @@
...
+
+      it("--output-dir, works", async () => {
+        const dist = `${mod}/dist`;
```
samayer12:
```suggestion
        const dist = `${module}/dist`;
```

integration/cli/build.nonconflict.test.ts
```diff
@@ -0,0 +1,187 @@
...
+        expect(await file.exists(dist)).toBe(false);
+
+        const out = outputDir;
+        expect(await file.exists(out)).toBe(true);
```
samayer12:
```suggestion
        expect(await file.exists(outputDir)).toBe(true);
```

Seems like we don't need to assign the higher-scoped variable to a local const since we use it for an assertion.

integration/cli/build.nonconflict.test.ts
```diff
@@ -0,0 +1,187 @@
...
+          expect(build.stderr.join("").trim()).toBe("");
+          expect(build.stdout.join("").trim()).toContain("K8s resource for the module saved");
+
+          packageJson = await resource.oneFromFile(`${mod}/package.json`);
+          uuid = packageJson.pepr.uuid;
+        },
+        time.toMs("1m"),
+      );
+
+      const getDepConImg = (deploy: kind.Deployment, container: string): string => {
```
samayer12, line 78: `DepCon` means DeploymentContainer?

integration/cli/build.nonconflict.test.ts
```diff
@@ -0,0 +1,187 @@
...
+
+          const watcherImage = valuesYaml.watcher.image;
+          expect(watcherImage).toBe(customImage);
+        }
+      });
+
+      it("--timeout, works", async () => {
+        const moduleYaml = await resource.manyFromFile(`${outputDir}/pepr-module-${uuid}.yaml`);
+        {
+          const mwc = resource.select(
```
samayer12, line 126: It's clear that `mwc` means `MutatingWebhookConfiguration` and spelling the whole thing out is a bit much. No issue with this abbreviation

integration/cli/build.nonconflict.test.ts
```diff
@@ -0,0 +1,187 @@
...
+            moduleYaml,
+            kind.MutatingWebhookConfiguration,
+            `pepr-${uuid}`,
+          );
+          const webhook = mwc.webhooks!.filter(f => f.name === `pepr-${uuid}.pepr.dev`).at(0)!;
```
samayer12:
```suggestion
          const webhook = mwc.webhooks!.filter(file => file.name === `pepr-${uuid}.pepr.dev`).at(0)!;
```

integration/cli/build.nonconflict.test.ts
```diff
@@ -0,0 +1,187 @@
...
+          expect(webhook.timeoutSeconds).toBe(timeout);
+        }
+        {
+          const mwc = resource.select(
+            moduleYaml,
+            kind.ValidatingWebhookConfiguration,
+            `pepr-${uuid}`,
+          );
+          const webhook = mwc.webhooks!.filter(f => f.name === `pepr-${uuid}.pepr.dev`).at(0)!;
```
samayer12:
```suggestion
          const webhook = mwc.webhooks!.filter(file => file.name === `pepr-${uuid}.pepr.dev`).at(0)!;
```

integration/cli/build.nonconflict.test.ts
```diff
@@ -0,0 +1,187 @@
...
+          expect(webhook.timeoutSeconds).toBe(timeout);
+        }
+
+        const valuesYaml = await resource.oneFromFile(`${outputDir}/${uuid}-chart/values.yaml`);
+        expect(valuesYaml.admission.webhookTimeout).toBe(timeout);
+      });
+
+      it("--withPullSecret, works", async () => {
+        const getDepImgPull = (deploy: kind.Deployment): string[] => {
```
samayer12, line 153: `getDepImgPull` means `getDeploymentImagePull`?

integration/cli/build.nonconflict.test.ts
```diff
@@ -0,0 +1,187 @@
...
+          return deploy!.spec!.template!.spec!.imagePullSecrets!.map(m => m.name!);
```
samayer12: What does `m` stand for in this filter? `module`?

integration/cli/build.nonconflict.test.ts
```diff
@@ -0,0 +1,187 @@
...
+        const admissionSecrets = getDepImgPull(admission);
+        expect(admissionSecrets).toEqual([withPullSecret]);
+
+        const watcher = resource.select(moduleYaml, kind.Deployment, `pepr-${uuid}-watcher`);
+        const watcherSecrets = getDepImgPull(watcher);
+        expect(watcherSecrets).toEqual([withPullSecret]);
+
+        // TODO: team talk
+        // Image pull secrets don't seem to map into the Helm chart anywhere..? Should they?
+        // TODO: end
```
samayer12: Let's write this up as an issue and remove the TODO

integration/cli/build.nonconflict.test.ts
```diff
@@ -0,0 +1,187 @@
...
+        const zarfYaml = await resource.oneFromFile(`${outputDir}/zarf.yaml`);
+        const component = zarfYaml.components
+          .filter((f: { name: string }) => f.name === "module")
+          .at(0);
+        expect(component.charts).toContainEqual(chart);
+      });
+
+      // it("--rbac-mode, ...todo!", async () => {
+      // --rbac-mode scoped
+      // }
```
samayer12:
```suggestion
```

Dangling comment, let's pull it

integration/cli/build.nonconflict.test.ts
```diff
@@ -0,0 +1,187 @@
...
+        const chart = {
+          name: "module",
+          namespace: "pepr-system",
+          version: "0.0.1",
+          localPath: `${uuid}-chart`,
+        };
+
+        const zarfYaml = await resource.oneFromFile(`${outputDir}/zarf.yaml`);
+        const component = zarfYaml.components
+          .filter((f: { name: string }) => f.name === "module")
```
samayer12:
```suggestion
          .filter((file: { name: string }) => file.name === "module")
```

Does `f` represent a `file` here?

integration/cli/build.registryinfo.test.ts
```diff
@@ -0,0 +1,97 @@
...
+describe("build", () => {
+  const workdir = new Workdir(`${FILE}`, `${HERE}/../testroot/cli`);
+
+  beforeAll(async () => {
+    await workdir.recreate();
+  });
+
+  describe("builds a module", () => {
+    const id = FILE.split(".").at(1);
+    const mod = `${workdir.path()}/${id}`;
```
samayer12:
```suggestion
    const module = `${workdir.path()}/${id}`;
```

integration/cli/build.registryinfo.test.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    let packageJson;
+    let uuid: string;
+
+    beforeAll(async () => {
+      await fs.rm(mod, { recursive: true, force: true });
```
samayer12:
```suggestion
      await fs.rm(module, { recursive: true, force: true });
```

integration/cli/build.registryinfo.test.ts
```diff
@@ -0,0 +1,97 @@
...
+      const argz = [
+        `--name ${id}`,
+        `--description ${id}`,
+        `--errorBehavior reject`,
+        "--confirm",
+        "--skip-post-init",
+      ].join(" ");
+      await pepr.cli(workdir.path(), { cmd: `pepr init ${argz}` });
+      await pepr.tgzifyModule(mod);
+      await pepr.cli(mod, { cmd: `npm install` });
```
samayer12:
```suggestion
      await pepr.tgzifyModule(module);
      await pepr.cli(module, { cmd: `npm install` });
```

integration/cli/build.registryinfo.test.ts
```diff
@@ -0,0 +1,97 @@
...
+    }, time.toMs("2m"));
+
+    describe("using a custom registry", () => {
+      const registryInfo = "registry.io/username";
+
+      const getDepConImg = (deploy: kind.Deployment, container: string): string => {
+        return deploy!.spec!.template!.spec!.containers.filter(f => f.name === container).at(0)!
```
samayer12:
```suggestion
        return deploy!.spec!.template!.spec!.containers.filter(file => file.name === container).at(0)!
```

integration/cli/build.registryinfo.test.ts
```diff
@@ -0,0 +1,97 @@
...
+      ].join(" ");
+      await pepr.cli(workdir.path(), { cmd: `pepr init ${argz}` });
+      await pepr.tgzifyModule(mod);
+      await pepr.cli(mod, { cmd: `npm install` });
+    }, time.toMs("2m"));
+
+    describe("using a custom registry", () => {
+      const registryInfo = "registry.io/username";
+
+      const getDepConImg = (deploy: kind.Deployment, container: string): string => {
```
samayer12, line 47: This stands for `getDeploymentContainerImage`, yeah?

integration/cli/build.registryinfo.test.ts
```diff
@@ -0,0 +1,97 @@
...
+        return deploy!.spec!.template!.spec!.containers.filter(f => f.name === container).at(0)!
+          .image!;
+      };
+
+      it(
+        "builds",
+        async () => {
+          const argz = [`--registry-info ${registryInfo}`].join(" ");
+          const build = await pepr.cli(mod, { cmd: `pepr build ${argz}` });
```
samayer12:
```suggestion
          const build = await pepr.cli(module, { cmd: `pepr build ${argz}` });
```

integration/cli/build.registryinfo.test.ts
```diff
@@ -0,0 +1,97 @@
...
+          expect(build.stdout.join("").trim()).toContain("K8s resource for the module saved");
+
+          packageJson = await resource.oneFromFile(`${mod}/package.json`);
+          uuid = packageJson.pepr.uuid;
+        },
+        time.toMs("1m"),
+      );
+
+      it("outputs appropriate configuration", async () => {
+        // overwrites --custom-image..?
```
samayer12:
```suggestion
```

Can we remove this comment? Was it implemented?

integration/cli/build.registryinfo.test.ts
```diff
@@ -0,0 +1,97 @@
...
+          const admission = resource.select(moduleYaml, kind.Deployment, `pepr-${uuid}`);
+          const admissionImage = getDepConImg(admission, "server");
+          expect(admissionImage).toBe(image);
+
+          const watcher = resource.select(moduleYaml, kind.Deployment, `pepr-${uuid}-watcher`);
+          const watcherImage = getDepConImg(watcher, "watcher");
+          expect(watcherImage).toBe(image);
+        }
+        {
+          const zarfYaml = await resource.oneFromFile(`${mod}/dist/zarf.yaml`);
```
samayer12:
```suggestion
          const zarfYaml = await resource.oneFromFile(`${module}/dist/zarf.yaml`);
```

integration/cli/build.registryinfo.test.ts
```diff
@@ -0,0 +1,97 @@
...
+          const componentImage = zarfYaml.components.at(0).images.at(0);
+          expect(componentImage).toBe(image);
+        }
+        {
+          const valuesYaml = await resource.oneFromFile(`${mod}/dist/${uuid}-chart/values.yaml`);
```
samayer12:
```suggestion
          const valuesYaml = await resource.oneFromFile(`${module}/dist/${uuid}-chart/values.yaml`);
```

integration/cli/build.registryinfo.test.ts
```diff
@@ -0,0 +1,97 @@
...
+          const admissionImage = valuesYaml.admission.image;
+          expect(admissionImage).toBe(image);
+
+          const watcherImage = valuesYaml.watcher.image;
+          expect(watcherImage).toBe(image);
```
samayer12, line 94: We expect both images to match `image`. Does that have to do with just how we set up the test scaffolding? Does this match real use?

integration/cli/build.version.test.ts
```diff
@@ -0,0 +1,68 @@
...
+describe("build", () => {
+  const workdir = new Workdir(`${FILE}`, `${HERE}/../testroot/cli`);
+
+  beforeAll(async () => {
+    await workdir.recreate();
+  });
+
+  describe("builds a module", () => {
+    const id = FILE.split(".").at(1);
+    const mod = `${workdir.path()}/${id}`;
```
samayer12:
```suggestion
    const module = `${workdir.path()}/${id}`;
```

integration/cli/build.version.test.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    beforeAll(async () => {
+      await fs.rm(mod, { recursive: true, force: true });
```
samayer12:
```suggestion
      await fs.rm(module, { recursive: true, force: true });
```

integration/cli/build.version.test.ts
```diff
@@ -0,0 +1,68 @@
...
+      const argz = [
+        `--name ${id}`,
+        `--description ${id}`,
+        `--errorBehavior reject`,
+        "--confirm",
+        "--skip-post-init",
+      ].join(" ");
+      await pepr.cli(workdir.path(), { cmd: `pepr init ${argz}` });
+      await pepr.tgzifyModule(mod);
+      await pepr.cli(mod, { cmd: `npm install` });
```
samayer12:
```suggestion
      await pepr.tgzifyModule(module);
      await pepr.cli(module, { cmd: `npm install` });
```

integration/cli/build.version.test.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    describe("using a specified pepr version", () => {
+      const cliVersion = "0.0.0-development";
+      const version = "1.2.3";
+
+      it(
+        "builds",
+        async () => {
+          const argz = [`--version ${version}`].join(" ");
+          const build = await pepr.cli(mod, { cmd: `pepr build ${argz}` });
```
samayer12:
```suggestion
          const build = await pepr.cli(module, { cmd: `pepr build ${argz}` });
```

integration/cli/build.version.test.ts
```diff
@@ -0,0 +1,68 @@
...
+          expect(build.exitcode).toBe(0);
+          expect(build.stderr.join("").trim()).toBe("");
+          expect(build.stdout.join("").trim()).toContain(cliVersion);
+
+          // TODO: team talk
+          // looks like it's just giving back the `pepr --version` then exiting,
+          //  rather than buidling/affecting the output files at all..?
+          expect(await file.exists(`${mod}/dist`)).toBe(false);
+          // TODO: end
```
samayer12: Let's tag up on this sync and see what the deal is here

integration/cli/build.version.test.ts
```diff
@@ -0,0 +1,68 @@
...
+        },
+        time.toMs("1m"),
+      );
+
+      // it(
+      //   "outputs appropriate configuration",
+      //   async () => { /* TOOD? */ }
+      // );
```
samayer12:
```suggestion
```

Dangling comments

integration/cli/init.test.ts
```diff
@@ -0,0 +1,60 @@
...
+  });
+
+  it(
+    "gives command line help",
+    async () => {
+      const argz = "--help";
+      const res = await pepr.cli(workdir.path(), { cmd: `pepr init ${argz}` });
+      expect(res.exitcode).toBe(0);
+      expect(res.stderr.join("").trim()).toBe("");
+      expect(res.stdout.at(0)).toMatch("Usage: pepr init");
```
samayer12:
```suggestion
      const result = await pepr.cli(workdir.path(), { cmd: `pepr init ${argz}` });
      expect(result.exitcode).toBe(0);
      expect(result.stderr.join("").trim()).toBe("");
      expect(result.stdout.at(0)).toMatch("Usage: pepr init");
```

integration/cli/init.test.ts
```diff
@@ -0,0 +1,60 @@
...
+    },
+    time.toMs("2m"),
+  );
+
+  it(
+    "creates new module using input args",
+    async () => {
+      const name = "flags-name";
+      const desc = "flags-desc";
+      const errb = "reject";
```
samayer12:
```suggestion
      const errorBehavior = "reject";
```
`errb` means....? ErrorBehavior?

integration/cli/init.test.ts
```diff
@@ -0,0 +1,60 @@
...
+      const argz = [
+        `--name ${name}`,
+        `--description ${desc}`,
+        `--errorBehavior ${errb}`,
```
samayer12:
```suggestion
        `--errorBehavior ${errorBehavior}`,
```

integration/cli/init.test.ts
```diff
@@ -0,0 +1,60 @@
...
+        "--confirm",
+        "--skip-post-init",
+      ].join(" ");
+      const res = await pepr.cli(workdir.path(), { cmd: `pepr init ${argz}` });
+      expect(res.exitcode).toBe(0);
+      expect(res.stderr.join("").trim()).toBe("");
+      expect(res.stdout.join("").trim()).toContain("New Pepr module created");
```
samayer12:
```suggestion
      const result = await pepr.cli(workdir.path(), { cmd: `pepr init ${argz}` });
      expect(result.exitcode).toBe(0);
      expect(result.stderr.join("").trim()).toBe("");
      expect(result.stdout.join("").trim()).toContain("New Pepr module created");
```

integration/cli/init.test.ts
```diff
@@ -0,0 +1,60 @@
...
+
+      const packageJson = JSON.parse(
+        await fs.readFile(`${workdir.path()}/${name}/package.json`, { encoding: "utf8" }),
+      );
+      expect(packageJson.name).toBe(name);
+      expect(packageJson.description).toBe(desc);
+      expect(packageJson.pepr.onError).toBe(errb);
```
samayer12:
```suggestion
      expect(packageJson.pepr.onError).toBe(errorBehavior);
```

integration/helpers/cmd.test.ts
```diff
@@ -0,0 +1,56 @@
+import { describe, expect, it } from "@jest/globals";
+import { Cmd } from "./cmd";
+
+describe("runRaw()", () => {
```
samayer12, line 4: See comments in `cmd.test.ts` for if we actually need this or can use a `DevDependency` instead.

integration/helpers/cmd.ts
```diff
@@ -0,0 +1,74 @@
...
+  env?: object; // object containing key-value pairs
+}
+
+export interface Result {
+  stdout: string[];
+  stderr: string[];
+  exitcode: number;
+}
+
+export class Cmd {
```
samayer12, line 16: Why do we prefer our own implementation instead of using something like exec or execSync?

integration/helpers/file.ts
```diff
@@ -0,0 +1,14 @@
+import { access } from "node:fs/promises";
+
+export async function exists(path: string): Promise<boolean> {
```
samayer12: For tests, it seems like we can afford do these checks synchronously. Can we use [`fs.existsSync()`](https://nodejs.org/api/fs.html#fsexistssyncpath) instead and not make this a concern that we need to maintain and test?

integration/helpers/file.test.ts
```diff
@@ -0,0 +1,28 @@
+import { beforeAll, describe, expect, it } from "@jest/globals";
+import * as path from "node:path";
+import * as fs from "node:fs/promises";
+import { Workdir } from "../helpers/workdir";
+import * as sut from "./file";
+
+const FILE = path.basename(__filename);
+const HERE = __dirname;
+
+describe("exists", () => {
```
samayer12: See comments for `file.test.ts`.

integration/helpers/workdir.ts
```diff
@@ -0,0 +1,50 @@
...
+  path() {
+    return path.join(this.root, this.leaf);
+  }
+
+  async create(): Promise<string> {
+    await fs.mkdir(this.path(), { recursive: true });
+    return this.path();
+  }
+
+  async exists() {
```
samayer12, line 25: See comment about using `fs.existsSync()` instead. Even if we can't use this, I think it's a duplicate implementation from `file.ts`.

integration/prep.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+ME="$(readlink -f "$0")"
```
samayer12, line 4: What does `ME` mean here?

integration/prep.sh
```diff
@@ -0,0 +1,10 @@
...
+HERE="$(dirname "$ME")"
```
samayer12, line 5: Is this similar to the current working-directory?

package.json
```diff
@@ -27,6 +27,9 @@
     "build:image": "npm run build && docker buildx build --output type=docker --tag pepr:dev .",
     "test": "npm run test:unit && npm run test:journey",
     "test:unit": "npm run gen-data-json && jest src --coverage --detectOpenHandles --coverageDirectory=./coverage --testPathIgnorePatterns='cosign.e2e.test.ts'",
+    "test:int": "npm run test:int:prep && npm run test:int:run",
+    "test:int:prep": "./integration/prep.sh",
+    "test:int:run": "jest --maxWorkers=4 integration",
```
samayer12:
```suggestion
    "test:integration": "npm run test:int:prep && npm run test:int:run",
    "test:integration:prep": "./integration/prep.sh",
    "test:integration:run": "jest --maxWorkers=4 integration",
```

Unless we see ourselves running the command `npm run test:integration:*` a lot locally, we might as well spell it out if it's mostly used in CI

samayer12:
```suggestion
    "test:integration": "./integration/prep.sh && jest --maxWorkers=4 integration",
```

Alternatively, based on the PR we really only use `test:int` in CI and the nested commands of `prep` and `run` only contain one command each.

So, in interest of avoiding a situation where `A` calls `B` & `C` where `B` calls `1` and `C` calls `2` we can have `A` call `1` and `2` directly. Less bloat for an already chunky `scripts` section.

.github/workflows/node.js.yml
```diff
@@ -44,6 +44,25 @@ jobs:
...
+        with:
+          node-version: 22
+          cache: "npm"
+      - name: Setup Helm
+        uses: azure/setup-helm@fe7b79cd5ee1e45176fcad797de68ecaf3ca4814 # v4.2.0
+        with:
+          version: v3.3.4
+
+      - run: npm ci
+      - run: npm run test:int
```
samayer12:
```suggestion
      - run: npm run test:integration
```

Personal style preference to avoid abbreviations for small words

integration/prep.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
```
samayer12, line 1:
```suggestion
#!/bin/sh
# This script makes a version of the npm cache for local use to avoid installing test artifacts into the global npm cache.
# This isn't an issue in CI where environments are ephemeral, but is useful for local testing.
```

We can add a quick blurb about why we need to prep to get around the npm cache problem.

integration/helpers/time.ts
samayer12, line 1: We should avoid taking on time-related utility functions as a concern for Pepr. We should investigate alternatives before concluding this is the best way to represent an arbitrary amount of milliseconds in human-readable formats.

If we accept that adding `devDependencies` for testing is a thing we can do without as much concern as real project dependencies, then we should research some alternatives such as:
* [millis-js](https://www.npmjs.com/package/millis-js)
* [convert-ms](https://www.npmjs.com/package/convert-ms)
* [milliseconds](https://www.npmjs.com/package/milliseconds)
* [convert-hrtime](https://www.npmjs.com/package/convert-hrtime)

integration/helpers/time.test.ts
```diff
@@ -0,0 +1,64 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2023-Present The Pepr Authors
+
+import { describe, it, expect } from "@jest/globals";
+import * as sut from "./time";
+
+describe("toHuman()", () => {
```
samayer12, line 7: See comments on `time.ts`. We should avoid taking responsibility for time converting functions unless we have a really good reason to do so.

integration/helpers/resource.ts
samayer12, line 1: Need to review this later, didn't have a chance to look at the details.

integration/helpers/resource.test.ts
samayer12, line 1: Need to review this later, didn't have a chance to look at the details.

integration/helpers/pepr.ts
samayer12, line 1: Need to review this later, didn't have a chance to look at the details.

integration/helpers/pepr.test.ts
samayer12, line 1: Need to review this later, didn't have a chance to look at the details.

samayer12 (changes requested):

integration/cli/build.defaults.test.ts
btlghrants: A reasonable preference -- and one I agree with, generally -- when the variable is going to be exported / made part of a public API / used widely / live a long time, but... none of that really applies here.

_Not_ using reasonable shortenings just adds length to the code & makes it harder to scan, IMO, especially for the names of a local variables that are instantiated & die no more than 20 lines later.  But I don't feel that strongly about this case so sure, we can change it.  ¯\\_(ツ)_/¯

On this particular item though: using the name "module" would shadow a node built-in that I'd prefer to leave alone, so... I'll use "testModule" instead.  That work?

integration/cli/build.noembed.test.ts
btlghrants: The plan was to do this _after_ the PR merged and to leave these comments in as a to-be-removed part of those Issues' "Definition of Done".

That would leave _this_ test as fully descriptive of the current behavior _and_ leave a searchable pin for where the new tests should live once work starts on follow-up issues but if I'm the only one that finds that sort of thing valuable then I can remove it.

To be clear though, what's your objection to leaving this nugget in a "TODO" comment?

integration/helpers/resource.ts
samayer12, line 1: This file appears to let the tester grab data from input that's _either_ JSON or YAML. So, having it as a helper seems reasonable. `manyFromFile()` and `oneFromFile()` seem very similar, I wonder if we can combine them instead of having two implementations that are almost the same.

My mindset is that operating on a collection with `1` item could be the same as operating on a set of `N` items with the right use of calls to things like filter/map/reduce. Maybe.

integration/helpers/pepr.ts
```diff
@@ -0,0 +1,50 @@
+import { resolve } from "node:path";
+import { cp, readFile, writeFile } from "node:fs/promises";
+import { basename } from "node:path";
+import { Spec, Cmd, Result } from "./cmd";
+import { clone } from "ramda";
+
+const HERE = __dirname;
+
+export async function projectRoot(): Promise<string> {
```
samayer12, line 8: I don't think we need to actually export this function since it's only called by `pepr.ts` and `pepr.test.ts`. That probably came about during initial implementation, which makes sense. Based off the rest of the unit tests in `pepr.test.ts`, I think this looks like exposing an implementation-detail of `tgzifyModule()` and `cli()` when do don't need to.

integration/helpers/pepr.ts
```diff
@@ -0,0 +1,50 @@
...
+export async function tgzifyModule(modulePath: string): Promise<void> {
+  const packagePath = `${modulePath}/package.json`;
+  const packageJson = JSON.parse(await readFile(packagePath, { encoding: "utf8" }));
+
+  const root = await projectRoot();
+  packageJson.dependencies.pepr = `file://${root}/pepr-0.0.0-development.tgz`;
+  await writeFile(packagePath, JSON.stringify(packageJson, null, 2));
+}
+
+export async function copyModule(src: string, dst: string): Promise<void> {
```
samayer12: `copyModule()` is only called once by `pepr.test.ts` and not used elsewhere. Do we really need it?

samayer12 (changes requested):
Had a chance to review `pepr.ts` and `resource.ts`, a few questions about implementation details. The overall purpose of the files makes sense.

integration/cli/build.nonconflict.test.ts
btlghrants: That no-op test is there just so that a reader of the tests can see from the test output that the `--entry-point` flag _is not_ being forgotten -- simply "successfully building" is the success bar here.

I guess there probably _are_ things that could be inserted there but I don't think they'd be anything more than busywork / test theater because:
- all entrypoint flag does is change the name of a file being looked for during build, and
- we establish / setup / name those files ourselves as part of test setup (i.e. moving `pepr.ts` to `pepr2.ts` and then telling the entrypoint to look for `pepr2.ts`).

Everything else from that point forward is bog standard.  The output module produced is the same, whatever the entrypoint (so long as it works to build).

Open to suggestions if you've got 'em, but I couldn't think of anything to test there that'd add anything more than extra runtime.

integration/cli/build.noembed.test.ts
samayer12: Sure, that works for me. At present we don't have an automated check that warns on lingering-TODOs (though, we could...). In my experience, they have a nasty habit of accumulating over time.

As long as we don't forget to revisit this topic, I'm cool with leaving it in 👍

integration/cli/build.nonconflict.test.ts
btlghrants: getDepConImg == "get Deployment Container Image" and is just a throw-away name for a throw-away lambda that we use here to avoid having to type out this YAML-object accessor-nightmare out the couple/few times we need it:
```
deploy!.spec!.template!.spec!.containers.filter(f => f.name === container).at(0)!.image!;
```

integration/cli/build.nonconflict.test.ts
btlghrants: "m" stands for "mapped item" (or whatever you'd like) at it is, admittedly, a throwaway identifier (a lot like "x" in any random algebraic function).

The thing that's being mapped there are `.imagePullSecrets` (right there on the same line, immediately before it), so if you were looking for a fully expanded name it'd be something like "imagePullSecret".

integration/cli/build.nonconflict.test.ts
btlghrants: More or less, yes.

integration/cli/build.registryinfo.test.ts
btlghrants: It's a pepr CLI design question -- one that I still don't know the answer to.  If the answer is yes then this should result in an update to the Pepr docs explaining that the `--registry-info` flag overwrites the `custom-image` contribution, because it appears to & surprise me here.

btlghrants: That said, I'll remove the comment and add it to the list of follow-on issues.

integration/cli/build.registryinfo.test.ts
btlghrants: It does indeed match real use -- that's how that CLI flag functions.

When we pass in a custom image / registry info flag to the `pepr build` command, we have to distribute the single resulting custom image string into 2 separate places within each manifest: one copy for running the Pepr admission controller, and the other for running the Pepr watch controller.

Same image is used for both.

integration/helpers/cmd.ts
btlghrants: We _definitely_ prefer this, since:
1.  this uses the node `spawn` method to interface with the system (just like we would / did if we chose to call into it more directly
1. this  is already in wide use over in the excellent examples -- [where we had to learn the shortcomings of exec/execSync the hard way](https://github.com/defenseunicorns/pepr-excellent-examples/pull/187) -- and
1. this has clear input/output object semantics which are _way_ easier to read / think through than looking at the buffers directly, having everyone throw together their own wrappers around exec/execSync/spawn wherever the need 'em, etc.

integration/helpers/file.ts
btlghrants: Yes, we can.  I like to stick with the async API when it comes to fs activities (since file access is slow & async APIs mean I don't have to hold up the JS loop to wait) but if we like bouncing between async & sync then the `fs.existsSync` call will work just file.

integration/helpers/pepr.ts
btlghrants: I mean, you _could_ look at it that way, but... I actually intended it to be exposed as it's a totally reasonable thing for other helpers to want to access to / to need to know.

If there's no _problem_ with exposing it, I think I'd prefer to leave it as-is.  Rather than thinking of it as an implementation detail, think of `tgzifyModule` and `cli` as higher-level helpers that build off this as a basic, pepr-module-scoped `projectRoot` helper instead.

integration/helpers/pepr.ts
btlghrants: Nope -- you're right.  This one was an oversight on my part when I did the conversion from serialized-to-parallelizable test layout I just fogot to remove it.

Consider it gone.

integration/helpers/time.test.ts
btlghrants: Don't think of this as "taking responsibility for time",  rather think of this as more of a test helper for human readable durations within tests.

Reason this is here is because `time.toMs("2m30s")` is _infinitely_ more readable than seeing random numbers (i.e. 150000) or math (i.e. `2.5 * 60 * 1000`) sprinkled around everywhere.

I agree about not owning time -- cuz I don't want to have to account for leap years / seconds / time-zones / etc. either -- but this isn't really that; this is more of a reversible, human-readable-duration-to-milliseconds helper function (which we already have in various styles, in various places... just that this is cleaner, and tested).

Alternatives to this are 1) pull in a whole, non-trivial time lib just to do simple formatting (which I don't like for minimizing-dependencies-is-good reasons), 2) remove this and go back to everybody calculating test timeout durations on the fly, or 3) defining local conversion functions wherever we need them (none of which are ideal, IMO).

If, at some point in the future, we find a good reason to pull in a "date/time library" (which I don't really think we have ATM), then I wouldn't object to ripping these out and relying on that instead, but... I'm not convinced that we have grounds for that just yet.

integration/prep.sh
btlghrants: Good suggestion.

integration/prep.sh
btlghrants: It means "me", literally.  As in, "me, myself, and I".

It contains a fully resolved, absolute path to the script being executed.

integration/prep.sh
btlghrants: Nah, not really.  Instead it refers to the directory containing the script being executed.

So, if you're executing the script from _within_ the directory the script sits in, then yes: this value and $(pwd) would return similar paths (though there might still be relative vs absolute issues).

But in the case where you execute this script from a location _outside_ the script directory (say, by running something like `./../this/is/not/in/pwd/script.sh`) then $(pwd) and "$HERE" will be very different.

integration/helpers/cmd.test.ts
btlghrants: It's what we're already using for running the majority of the e2es assertsion, so... if you want to swap it out with something else, we should definitely talk about that as a separate stream of work (cuz the change won't just be here).

Since this is just the thinnest wrapper around node's native `child_process.spawn` though... until we find we can't make it work / find we need significantly more complex CLI interaction facilities my vote is probably going to be to just leave this as-is.

integration/helpers/resource.ts
btlghrants: Yep, good eye.  I thought so too originally.  I tried to combine them early-on but just couldn't get the return types to work out in a way that wouldn't require a caller to have to do a runtime "is this a single or a collection"-style check every time it was used though, so I settled for a two-func api.

Admittedly, it forces the caller state up-front whether they want one or many items back, but... that feels cleaner to me than requiring a caller to do / possibly forget to do an after-call type check.

Definitely open to refinements here if your TS is better than mine.  Minor caveat to watch for -- should you decide to try your hand -- is that the yaml lib we're using has a separate read function for reading from singular vs  multi-doc files, so... there be dagrons.

samayer12: I'll take a crack at it and report back with suggested changes if I come up with any

integration/cli/build.defaults.test.ts
samayer12: `mod` -> `testModule` sounds good, thanks!

integration/helpers/resource.ts
samayer12: @btlghrants, #1663 is a PR to this PR that refactors the parsing logic. The tests in `resource.test.ts` pass locally, but as we noted today there are issues running this workflow in a Mac environment. So, I'm not certain that the integration tests are unaffected but the unit test are g2g.

Merge #1663 into this PR if you're cool with that refactor.

btlghrants: Love it.  Merged it in.

integration/helpers/time.ts
btlghrants: I think... I definitely have concerns about pulling dependencies _even if_ they're devDeps (since hacks can happen in CI too).

Open to have a deeper conversation about what position we should take as a project here, but for now this seems a reasonably simple test shim for avoiding random math being done to convert human readable durations into ms-based timeout values.  Shouldn't be much of a problem to rip out & replace later should we find the need to pull in a new lib for dates/times either.

package.json
btlghrants: Nah, would rather not collapse these for dev-flow purposes.

The benefit of the separation is that I can just run the "test:integration:prep" step _once_ to get the pepr-dev-version-.tgz installed and then I can run the "test:integration:run" command a _bunch_ of times as I iterate on the tests.  Saves me 10s-ish every run by not having to run the prep step over and over.

integration/helpers/file.test.ts
btlghrants: Fixed!

integration/cli/build.version.test.ts
btlghrants: OBE -- comment removed.  Issue added to address: https://github.com/defenseunicorns/pepr/issues/1661

Happy to still sync on if you wanna, of course, as part of a different PR though. :+1:

integration/helpers/time.test.ts
samayer12: Sure, easy enough to rip out internal calls with a library if we prefer

samayer12 (approved):
Big lift! There are some areas to consolidate and discussions around which helpers we should own vs. list as devdeps, but that's for later.

PR #1688, opened 2025-01-17
test: demonstrate that 'pepr build' outputs comparable config-derived values

### LINKED ISSUE #1640: Ensure helm templates manifests correspond to Kubernetes manifests 1:1

btlghrants:
## Description

Automatically validtes that pepr-native output resource manifest and helm charted manifest contain comparable config-derived values.

## Related Issue

Fixes #1640 

## Type of change

## Checklist before merging


integration/cli/build.configequality.test.ts
samayer12, line 1: The `it()` test names should follow a BDD pattern along the lines `should $RESULT [when $CONDITON]` or some similar version instead of referencing just the property name. Here's [a reference](https://dev.to/smyrman/test-with-expect-a-bdd-style-go-naming-pattern-5eh5) to BDD naming in Go that I find useful.

integration/cli/build.configequality.test.ts
```diff
@@ -0,0 +1,424 @@
...
+      ].join(" ");
+      await pepr.cli(workdir.path(), { cmd: `pepr init ${argz}` });
+      await pepr.tgzifyModule(testModule);
+      await pepr.cli(testModule, { cmd: `npm install` });
+    }, time.toMs("2m"));
+
+    describe("maps config options from package.json into pepr + helm manifests", () => {
+      let moduleConfig: Required<ModuleConfig>;
+      let peprResources: any; // eslint-disable-line @typescript-eslint/no-explicit-any
+      let helmResources: any; // eslint-disable-line @typescript-eslint/no-explicit-any
```
samayer12, line 46: Can we avoid `any` and specify a type instead? If not, does that warrant an issue to add better typing later?

integration/cli/build.configequality.test.ts
```diff
@@ -0,0 +1,424 @@
...
+
+      // eslint-disable-next-line max-statements, complexity
+      it("uuid", async () => {
+        for (const clusterRole of [
+          resource.select(peprResources, kind.ClusterRole, peprUuid),
+          resource.select(helmResources, kind.ClusterRole, peprUuid),
+        ]) {
+          expect(clusterRole).toBeDefined();
+        }
+
```
samayer12, line 134: Tests after this line are unrelated to `uuid` but still part of the `uuid` test.

integration/cli/build.configequality.test.ts
samayer12, line 1: Each of the `for(const X of [Y,Z])` present a few areas for improvement that either provide more atomic feedback or reduce maintenance overhead:

* _Serial nature of assertions_: Checks after the first failing `expect` will break. We can rewrite each `for` loop as its own test case. `peprResources` and `helmResources` are declared once in the `beforeAll()`.
~~* _Functional alternative to for-loops_: We can use `flatMap()`, `map()`, or `forEach()` for situations where we're iterating over collections. See example rewrite later for what I have in mind. Seems like most of these tests are along the lines of "Does A equal B"~~
    ~~* My opinion is that `forEach()` lets the reader focus on the operations on the data instead of the looping mechanism.~~
   ~~* It's also easier to to additional transformations on the data~~

integration/cli/build.configequality.test.ts
```diff
@@ -0,0 +1,424 @@
...
+        for (const clusterRoleBinding of [
+          resource.select(peprResources, kind.ClusterRoleBinding, peprUuid),
+          resource.select(helmResources, kind.ClusterRoleBinding, peprUuid),
+        ]) {
+          expect(clusterRoleBinding).toBeDefined();
+          expect(clusterRoleBinding.roleRef.name).toBe(peprUuid);
+          expect(clusterRoleBinding!.subjects!.at(0)!.name).toBe(peprUuid);
+        }
```
samayer12, line 142: Here's an example of what an alternate to the for-loop pattern may look like: 
```suggestion
        [
          resource.select(peprResources, kind.ClusterRoleBinding, peprUuid),
          resource.select(helmResources, kind.ClusterRoleBinding, peprUuid),
        ].flat().forEach(clusterRoleBinding => {
          expect(clusterRoleBinding).toBeDefined();
          expect(clusterRoleBinding.roleRef.name).toBe(peprUuid);
          expect(clusterRoleBinding!.subjects!.at(0)!.name).toBe(peprUuid);
        });
```

integration/cli/build.configequality.test.ts
samayer12, line 1: How do calls to `resource.select()` handle `null` or `undefined` fields? I checked `resource.test.ts` and didn't see any test cases for what the behavior is in those situations. If we have to allow `any` as the type for `peprResources` and `helmResources`, then `resource.select()` should have defined behavior for null/undefined backed up by test coverage.

samayer12 (changes requested):

integration/cli/build.configequality.test.ts
btlghrants: Don't think so, since those variables are _supposed_ to hold `any` value pulled in from parsing free-form .yaml files.

Before we use anything out of those variables for testing though, we're sub-selecting items from within them -- using resource.select() calls a bit further down -- and doing _that_ strongly types them so that they're not `any` by the time we use them.

integration/cli/build.configequality.test.ts
btlghrants: ...wha?  I don't think I follow.  All the assertions after that are for properties / values in the generated files that _use_ the config'd UUID (and there are a _lot_ of them).

integration/cli/build.configequality.test.ts
btlghrants: Mmm, I don't think `resource.select()` _has_ an opinion about null or undefined fields -- it just returns the data you gave it as the type you told it it was supposed to be.

I'm not sure there _is_ a way for `resource.select()` to have an informed opinion here really -- at least, not without some _serious_ thinking / time & energy spent on it.  Feel free to put an issue in for it if you feel strongly about it, I guess?

As for this PR, that's just a helper that's already in the test suite, that we're already using elsewhere, and that meets needs for this test case, so I don't think I want to address it here.

integration/cli/build.configequality.test.ts
btlghrants:
```
We can use flatMap(), map(), or forEach() for situations where we're iterating over collections.
```
That's true.  But we could also just use plain-ole loops, right?  In this case, there's nothing _wrong_ with the classic loops, and I chose it intentionally since I _liked_ that the loop variable was in the top line.  IMO,  it increases scannability of the file & helps set context for how a dev is intended to read the contents of the loop.

For example:
```js
for (const deployAdmission of [ 
  resource.select(peprResources, kind.Deployment, peprUuid),
  resource.select(helmResources, kind.Deployment, peprUuid),
]) {
  expect(deployAdmission.metadata!.annotations!["pepr.dev/description"]).toBe(
    moduleConfig.description,
  );
}
```
That top line:
```
for (const deployAdmission of [ 
```
helps hint to the reader that what's coming next are Deployments of the _admission_ flavor, for example.  It's a way to use the loop variable naming to help top-down readability.

Burying that variable between the loop definition and the functional function body just loses out on an opportunity & increases reading friction, IMO.  If I read the array of values first, I think, "What are these Deployments of?" and then I have to go digging for an answer.

Also, just more generally, I find plain ole' loops tend to be more resilient against oft-unintended consequences of mixing .forEach-like iteration with the async/await-style programming.  Simple-ish S.O explanation here: https://stackoverflow.com/a/37576787 (in short, there be dragons).  Not really an issue here, but who can say what the future holds?  A feather in the cap of `for..of`s, if nothing else.

Given all that, I think I'd prefer to keep the loops as they are.

integration/cli/build.configequality.test.ts
btlghrants: While suggestions like that can be useful -- and are certainly helpful for learning / framing thinking  -- I generally find them to be incredibly verbose.  What I value is being able to grok the intention behind the tests I'm reading / seeing the result of, and so I try to optimize for that rather than adherence to a pattern.

To that end, I wrote the test names the way I did so that I could get a test run output like this:
```
builds a module
      maps config options from package.json into pepr + helm manifests
        ✓ peprVersion (8 ms)
        ✓ appVersion (4 ms)
        ✓ uuid (27 ms)
        ✓ description (3 ms)
        ✓ webhookTimeout (2 ms)
        ✓ onError (2 ms)
        ✓ alwaysIgnore.namespaces (6 ms)
        ✓ logLevel (4 ms)
        ✓ env (4 ms)
        ✓ customLabels.namespace (2 ms)
        ✓ rbacMode: scoped + rbac: [...] (3 ms)
```

This lets me know that the tests have accounted for each of the config options we expose _and_ makes it easier to review "at a glance" / more likely I'll spot something wrong as I review the test results.  I don't think adding "given / when / then"s or additional "should"s into the test naming would make the output any clearer, so... what kind of thing were you thinking?

integration/cli/build.configequality.test.ts
samayer12: Sure thing, that's certainly more of a stylistic point.  Appreciate the POV

integration/cli/build.configequality.test.ts
samayer12: Functional alternatives discussed [here](https://github.com/defenseunicorns/pepr/pull/1688#discussion_r1927583874), choosing to stick with the for-loop

integration/cli/build.configequality.test.ts
samayer12: Ahh gotcha, the purpose of the test was lost among the 140-line implementation & same-y looking for loops to check properties. I wonder if there's an approach to this test along the lines of a parameterized test. Will spend a few cycles on the idea so we comply with linter rules.

integration/cli/build.configequality.test.ts
btlghrants:
```
Serial nature of assertions: Checks after the first failing expect will break.
```
Yep, I think it's an admirable goal... in spirit -- but much like trying to attain 100% code coverage, I think there are limits to when it is advisable to chase it.  In this case, I opted _against_ spreading those assertions out since 1) it _massively_ expands the test result output text, and 2) that output text would mostly be talking about _implementation detail_ rather than desired behavior.

For example:
```
builds a module
  maps config options from package.json into pepr + helm manifests
    ✓ uuid (27 ms)
```
^-- this says to me that `package.json`'s `uuid` option gets mapped into both the pepr & helm manifests "appropriately" (which is about the right level of detail for an integration test, IMO).  That's it because that's all I want a reader to have to care about.  It doesn't say how specific properties it gets mapped in to specific resources, how many times a config item gets used, etc. because all that is just "how it works" rather than "that it does work".

Possible reworkings could look like:
```
builds a module
  maps config options from package.json into pepr + helm manifests
    ✓ uuid should be used in ClusterRole (X ms)
    ✓ uuid should be used in ClusterRoleBinding (X ms)
    ✓ uuid should be used in ServiceAccount (X ms)
    ✓ uuid should be used in API Token Secret (X ms)
    ✓ uuid should be used in TLS Secret (X ms)
    [... omitting 10 more ...]
```
Or to take it even further:
```
builds a module
  maps config options from package.json into pepr + helm manifests
    ✓ uuid should be used to name ClusterRole (X ms)
    ✓ uuid should be used to name ClusterRoleBinding (X ms)
    ✓ uuid should be used as roleRef name within ClusterRoleBinding (X ms)
    ✓ uuid should be used as subject name within ClusterRoleBinding (X ms)
    ✓ uuid should be used to name ServiceAccount (X ms)
    ✓ uuid should be used to name API Token Secret (X ms)
    ✓ uuid should be used to name TLS Secret (X ms)
    [... omitting LOTS more ...]
```
^-- neither of which seem preferable to me in terms of "cost of reading" when weighed against "value of knowing that nothing failed" + "value of not having to reorganize test output when manifest implementations change".

samayer12: Standard test names across the project are a much broader topic than this PR. I ought to survey the current pattern on Pepr and see if there's anything silly that has emerged in our test suite over time. At present, that test output is workable

samayer12: Naming things (one of the two hard problems in computer science) is distinct from the value provided to Pepr by the tests in this PR.  I'm willing to flex on test naming if we create an issue in the backlog to document testing standards and/or spend some time at a Pepr Cruising session doing some team norming. 

We can see that existing tests in Pepr generally are named along the lines of `it("should...")` or `it("returns...")`. One can check that for themselves with a command like `grep --no-filename  -e "^\s*test(" -e "^\s*it(" src/**/*.test.ts integration/**/*.test.ts journey/**/*.test.ts | awk '{$1=$1};1' | sort`.

samayer12: Had a sync discussion with @btlghrants around this idea, identified some good topics related to how we verify helm manifests generated from pepr manifests that we let users modify.

I plan to take some time to examine the test implementation a little more and see if an `it.each` type of approach is appropriate or a bridge too far.

integration/cli/build.configequality.test.ts
```diff
@@ -0,0 +1,424 @@
...
+
+      it("appVersion", async () => {
+        const peprStringified = JSON.stringify(peprResources);
+        const helmStringified = JSON.stringify(helmResources);
+
+        expect(peprStringified.includes(moduleConfig.appVersion)).toBe(false);
+        expect(helmStringified.includes(moduleConfig.appVersion)).toBe(false);
+      });
+
+      // eslint-disable-next-line max-statements, complexity
```
samayer12:
```suggestion
```

Refactoring this function adds more PR review overhead than it's worth at-present. I'd like to remove this linter suppression since `max-statements` and `complexity` are `warn` and not `error`.

The team has scheduled some time to talk about testing strategies next week. I'd prefer to not ignore the linter warnings and accept some warning output when working in this file (which we don't foresee changing very much anyways).

integration/cli/build.configequality.test.ts
samayer12: The team scheduled a session to norm around testing strategies, so we can consider this one resolved since we've had some discussions around the topic.

samayer12 (approved):
Tests are good, implementation details around the tests are for a different conversation once the team norms around testing practices.

PR #1753, opened 2025-01-31
chore: add more testing around asset class

### LINKED ISSUE #1655: Improve test coverage in `src/lib/assets/yaml`

While removing circular dependencies to resolve #1632, some additional untested code was discovered. We should backfill missing tests for this code. Before starting this issue, ensure that #1632 is closed & #1652 has been merged.

### Additional context

The decrease in code-coverage is noted in #1652.

cmwylie19:
## Description

The assets class does not have unit tests. The purpose of this PR is to add coverage to ensure the class is working as expected. 

Due to the use of dependency injection we are loosely helper functions with the method and functions on the assets class that rely on members of the assets class. Due to that, the intent of this PR is to ensure those injected functions are being with the correct members of the assets class and to hedge against accidentally change by ensuring certain functions are called a certain number of times. 

## Related Issue

Fixes #1655 

Relates to #

## Type of change

## Checklist before merging

### DIFF
--- a/src/lib/assets/assets.test.ts
+++ b/src/lib/assets/assets.test.ts
@@ -264,7 +264,7 @@ describe("Assets", () => {
     expect(consoleErrorSpy).not.toHaveBeenCalled();
   });
 
- it("should call generateHelmChart without an error when asset class instance is correct", async () => {
+  it("should call generateHelmChart without an error when asset class instance is correct", async () => {
     const consoleErrorSpy = jest.spyOn(console, "error").mockImplementation(() => {});
 
     const webhookGeneratorFunction = jest
--- a/src/lib/assets/assets.test.ts
+++ b/src/lib/assets/assets.test.ts
@@ -2,19 +2,164 @@
 // SPDX-FileCopyrightText: 2023-Present The Pepr Authors
 import { ModuleConfig } from "../core/module";
 import { Assets } from "./assets";
-import { expect, describe, it, jest } from "@jest/globals";
+import { expect, describe, it, jest, afterAll } from "@jest/globals";
+import { CapabilityExport } from "../types";
+import { kind } from "kubernetes-fluent-client";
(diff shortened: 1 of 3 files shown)


src/lib/assets/assets.ts
samayer12, line 1: I noticed that the `catch` condition of `generateHelmChart()` isn't covered under this block of tests. 

As part of this PR, should we add in a test-case for defined error behavior? Maybe something along the lines of having one of the statements within the `try`-block (e.g., `Promise.all()`, `overridesFile()`, or even `this.writeWebhookFiles`) get mocked and throw an exception when called?

cmwylie19: Added an additional test leaving the catch clause covered just incase we make an error in our configuration somewhere so this is caught. Good suggestion 👍

cmwylie19: I believe this comment is addressed because I have addressed the error cause in a singular case but if an error  is thrown in another place the same clause would catch it. lmk if you want to see changes

src/lib/assets/assets.test.ts
```diff
@@ -261,4 +263,39 @@ describe("Assets", () => {
     await assets.generateHelmChart(webhookGeneratorFunction, "/tmp");
     expect(consoleErrorSpy).not.toHaveBeenCalled();
   });
+
+  it("should call generateHelmChart and get and not get an error when asset class instance is correct", async () => {
```
samayer12:
```suggestion
 it("should call generateHelmChart without an error when asset class instance is correct", async () => {
```

Rewording test name for clarity.

samayer12 (approved):
There's room for improvement around how we generate test data, and some of the tests may run the risk of being tied _too_ closely to implementations. However, this is an improvement over previously untested code. Any future refactors of the test suite will have a good starting point 👍

Issue #1380, opened 2024-11-04
Remove calls to `process.exit()`


samayer12:
### Describe what should be investigated or refactored

There are several spots where Pepr calls `process.exit()`. This complicates the ability to do global logging, auditing, or cleanup. We should avoid this pattern to handle failure cases in a more graceful manner. Consider this explanation for why it's a [bad idea](https://www.appmarq.com/public/tqi,1020876,Avoid-using-nodejs-processexit-TypeScript#:~:text=exit()%20(TypeScript),-Rule%20Definition&text=Calling%20process.,I%2FO%20operations%20to%20process.).

We'll know we're done with this issue when each of those calls to `process.exit()` handle shutdown in a more graceful manner (e.g., throwing exceptions, passing control to some other function, etc.)

### Links to any relevant code

```
src/cli/build.ts:          process.exit(1);
src/cli/build.ts:            process.exit(1);
src/cli/build.ts:            process.exit(1);
src/cli/build.ts:          process.exit(1);
src/cli/build.ts:    process.exit(1);
src/cli/build.ts:    if (!e.stdout) process.exit(1); // Exit with a non-zero exit code on any other error
src/cli/deploy.ts:        process.exit(1);
src/cli/deploy.ts:        process.exit(1);
src/cli/deploy.ts:          process.exit(0);
src/cli/deploy.ts:          process.exit(1);
src/cli/dev.ts:          process.exit(0);
src/cli/dev.ts:            process.exit(1);
src/cli/dev.ts:        process.exit(1);
src/cli/format.ts:        process.exit(1);
src/cli/init/index.ts:          process.exit(1);
src/cli/kfc.ts:        process.exit(1);
src/cli/monitor.ts:        process.exit(1);
src/cli/update.ts:        process.exit(1);
src/cli/update.ts:        process.exit(1);
src/lib/assets/index.ts:      process.exit(1);
src/lib/assets/pods.ts:    process.exit(1);
src/lib/controller/index.ts:        process.exit(0);
src/lib/module.ts:          process.exit(1);
src/lib/watch-processor.ts:    process.exit(1);
src/lib/watch-processor.ts:    process.exit(1);
src/runtime/controller.ts:    process.exit(1);
```

### Additional context

Generate the above list with the command `grep "process.exit(" src/**/*.*s`

cmwylie19:
Lets break this up to stories by file to minimize risk and ease of rollback

PR #1757, opened 2025-02-03
chore: update peer deps

### LINKED ISSUE #1756: Update peer dependencies
Peer dependencies are manually updated due to limitations with dependabot.

Resolve this issue by updating peer dependencies in Pepr.

samayer12:
## Description

Regularly-scheduled update of dependencies.

End to End Test:    

## Related Issue

Fixes #1756 

## Type of change

## Checklist before merging

### DIFF
--- a/src/cli/deploy.ts
+++ b/src/cli/deploy.ts
@@ -93,6 +93,35 @@ export async function getUserConfirmation(opts: { confirm: boolean }): Promise<b
   return confirm.confirm ? true : false;
 }
 
+async function buildAndDeployModule(image: string, force: boolean): Promise<void> {
+  const builtModule = await buildModule();
+  if (!builtModule) {
+    return;
+  }
+
+  // Generate a secret for the module
+  const webhook = new Assets(
+    { ...builtModule.cfg.pepr, description: builtModule.cfg.description },
+    builtModule.path,
+    [],
+  );
+  webhook.image = image ?? webhook.image;
+
+  try {
+    await webhook.deploy(deployWebhook, force, builtModule.cfg.pepr.webhookTimeout ?? 10);
+
+    // wait for capabilities to be loaded and test names
+    validateCapabilityNames(webhook.capabilities);
+
+    // Wait for the pepr-system resources to be fully up
+    await namespaceDeploymentsReady();
+    console.info(`✅ Module deployed successfully`);
+  } catch (e) {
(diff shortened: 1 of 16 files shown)


samayer12:
#1758 addresses the drop in code coverage while updating peer dependencies.

src/cli/deploy.ts
```diff
@@ -117,33 +146,10 @@ export default function (program: RootCmd): void {
         return;
       }
 
-      (await getUserConfirmation(opts)) || process.exit(0);
-
-      const builtModule = await buildModule();
-      if (!builtModule) {
-        return;
+      if (!(await getUserConfirmation(opts))) {
+        process.exit(0);
```
cmwylie19, line 150:
```suggestion
        return;
```

This will not stop the PR from going through (technically it is a different story where we talk about this), but do we want to return instead instead of process exit when opts.confirm is negative?

src/cli/deploy.ts
```diff
@@ -93,6 +93,35 @@ export async function getUserConfirmation(opts: { confirm: boolean }): Promise<b
...
+
+    // wait for capabilities to be loaded and test names
+    validateCapabilityNames(webhook.capabilities);
+
+    // Wait for the pepr-system resources to be fully up
+    await namespaceDeploymentsReady();
+    console.info(`✅ Module deployed successfully`);
+  } catch (e) {
+    console.error(`Error deploying module:`, e);
+    process.exit(1);
```
cmwylie19, line 121: Same question if we want to handle this differently than process.exit. Does not really have to do with updating peer deps though and it was already here. Could wait until that story comes up since we would have to handle the return on this function differently

cmwylie19 (approved):
LGTM. Left a few non-blocking questons, they are probably better for another story.

I have manually tested build and deploy (confirm) although we already have good coverage around these parts. We also have good coverage for the hooks in the e2e tests.

Nice job!

src/cli/deploy.ts
samayer12: I agree that we should prefer to avoid `process.exit(0)` calls in favor of `return`. This PR focuses on just the peerDeps so I wanted to stay focused on that one task. Good callout for #1380!

src/cli/deploy.ts
samayer12: Agreed, we have #1380 for this specific goal so we'll leave it as-written on this PR.

Issue #1776, opened 2025-02-07
Enhance our Pepr Debugging docs


cmwylie19:
### Describe what should be investigated or refactored

https://docs.pepr.dev/main/best-practices/#debugging Needs to include how to tell if a watch event was caught, and if a resource was mutated.

cmwylie19:
@samayer12 This ticket came out of retro, do you think there needs to be more docs on top of the ones that exist now? https://docs.pepr.dev/main/best-practices/#debugging Is there anything specific that you would want to see?

PR #318, opened 2023-10-13
docs: fix typo dom intellisense discussion

cmwylie19:
## Description

The WebAssembly function is called `concats`, not `peprWASM`. I changed the name of the function to describe what it does and format to update the docs.

Additionally, this addresses comments regarding the verbosity in the intellisense while enabling WebAssembly

## Related Issue

Fixes #

Relates to #

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Other (security config, docs update, etc)

## Checklist before merging

- [x] Test, docs, adr added or updated as needed
- [x] [Contributor Guide Steps](https://github.com/defenseunicorns/pepr/blob/main/CONTRIBUTING.md#submitting-a-pull-request) followed

PR #1803, opened 2025-02-10
chore: check .md files with markdownlint-cli

### LINKED ISSUE #2120: Ensure quality of markdown docs with a linter

#1803 was created as a hotfix PR to add markdown linting. This has spent a lot of time on hold, partly because it was not tied to an issue in the backlog.

This issue serves to track that line of effort and can be added to an iteration when appropriate.

### Additional context

See also, #2080

samayer12:
## Description

While working on #1776, I noticed that we do not use a markdown linter. We should lint our markdown code since it's used to create user-facing documentation. This PR adds `markdownlint-cli` as a project dependency and configures `.md` linting upon commit.

## Related Issue

Fixes #2120 

Relates to #1776 

## Type of change

## Checklist before merging


src/templates/tsconfig.module.json
```diff
@@ -5,7 +5,7 @@
     "declarationMap": true,
     "emitDeclarationOnly": true,
     "esModuleInterop": true,
-    "lib": ["ES2022"],
+    "lib": ["ES2022", "DOM"],
```
cmwylie19: I thought we did not want the dom type definitions? 

https://www.typescriptlang.org/tsconfig/#lib

```
You may want to change these for a few reasons:

Your program doesn’t run in a browser, so you don’t want the "dom" type definitions
```

samayer12: Yeah, gotta think a bit on this one. I noticed that in Pepr's `.tsconfig.json` file we added `DOM` but not in the templates. That seems reasonable, but not including `DOM` causes issues with running a `pepr build` once `markdownlint-cli` was added as a dependency due to a typing issue with `katex` (a transitive).

Why that causes the build of a new module to [fail](https://github.com/defenseunicorns/pepr/actions/runs/13249499731/job/36983848870) in the first place seems weird, since they don't share devDeps.

cmwylie19: Got it, im not sure that this is bad either. I saw that this one time had "DOM" included.

I vaguely believe I remember that this _may_ have messed up intellisense on the module.. We have to check bc I do not remember anymore

cmwylie19: I finally found the convo. I did this to allow intellisense around WASM files in October of 2023. Jeff made a point that this explodes the amount of types for the end user. 

https://github.com/defenseunicorns/pepr/pull/318#issuecomment-1764611965

samayer12: Looks like we were able to do this without including DOM after all!

samayer12:
Use of `markdownlint-cli` requires `DOM` to be set in the `tsconfig.json` because of `katex`. See `npm list katex`:

```
pepr@0.0.0-development /Users/sam.mayer/code/work/pepr
└─┬ markdownlint-cli@0.45.0
  └─┬ markdownlint@0.38.0
    └─┬ micromark-extension-math@3.1.0
      └── katex@0.16.22
```

As a workaround, we could remove `markdownlint-cli` as part of the build command. That approach needs a little more investigation.

samayer12:
Since this is only a dev-tool that we do not need to distribute, we could use `npx` to just run the command as-needed in the hooks.

cmwylie19 (approved):

Issue #1791, opened 2025-02-10
remove calls to process.exit in `src/cli/format.ts`


cmwylie19:
### Describe what should be investigated or refactored

There are several spots where Pepr calls process.exit(). This complicates the ability to do global logging, auditing, or cleanup. We should avoid this pattern to handle failure cases in a more graceful manner. Consider this explanation for why it's a [bad idea](https://www.appmarq.com/public/tqi,1020876,Avoid-using-nodejs-processexit-TypeScript#:~:text=exit()%20(TypeScript),-Rule%20Definition&text=Calling%20process.,I%2FO%20operations%20to%20process.).

We'll know we're done with this issue when each of those calls to process.exit() handle shutdown in a more graceful manner (e.g., throwing exceptions, passing control to some other function, etc.)

```
src/cli/format.ts:        process.exit(1);
```

PR #1805, opened 2025-02-10
chore: early returns and console errors instead of process.exit

### LINKED ISSUE #1794: remove calls to process.exit in `src/cli/monitor.ts`

There are several spots where Pepr calls process.exit(). This complicates the ability to do global logging, auditing, or cleanup. We should avoid this pattern to handle failure cases in a more graceful manner. Consider this explanation for why it's a [bad idea](https://www.appmarq.com/public/tqi,1020876,Avoid-using-nodejs-processexit-TypeScript#:~:text=exit()%20(TypeScript),-Rule%20Definition&text=Calling%20process.,I%2FO%20operations%20to%20process.).

We'll know we're done with this issue when each of those calls to process.exit() handle shutdown in a more graceful manner (e.g., throwing exceptions, passing control to some other function, etc.)

```
src/cli/monitor.ts:        process.exit(1);
```

cmwylie19:
## Description

Relies on console.error to report errors and returns early when there are errors instead of calling `process.exit`

## Related Issue

Fixes #1794 #1792 #1791 

Relates to #

## Type of change

## Checklist before merging


src/cli/format.ts
samayer12, line 1: This seems like an okay place to do `process.exit(1)` since `pepr format` is a command that the user would run from CLI. However, we don't have any test cases around defined behavior if `success` is false.

We should add a test case that we can use to verify failure behavior (which could be exit `1` or some other option.

src/cli/monitor.ts
samayer12, line 1: This is also a command we'd expect the user to invoke from CLI, so maybe a `process.exit(1)` is acceptable here. We lack testing around what to do in an error case and should back it up with an example CLI test since it represents user behavior.

src/cli/init/index.ts
samayer12, line 1: We should match this to a test case in `init.test.ts` that experiences an error. Not sure if the status code would be non-zero here. Since `pepr init` is a user-facing command, it may be okay to leave the exit code. Hard to tell without a corresponding test case

samayer12 (changes requested):
The three examples addressed by this PR are what I'd consider "user-facing" and it may make sense to leave `process.exit(1)` calls in those instances.

What I _did_ notice is that the existing tests cover the "happy path" and do not consider the experience provided to the user when those commands fail. I think it makes sense to match those branches with test cases.

cmwylie19:
It is a good idea to inform the user during CLI commands that the command failed

PR #1807, opened 2025-02-11
chore: bump esbuild version

samayer12:
## Description

This PR mitigates [GHSA-67mh-4wv8-2f99](https://vat.dso.mil/vat/finding?name=GHSA-67mh-4wv8-2f99). We use `esbuild` to bundle the Pepr module, which is during the build process in the CLI. This code is not accessible during runtime. There is no attack vector in Pepr vulnerable to this.

Doing this upgrade will avoid us flagging on security scans so that we do not need to justify a security finding that does not impact us.

This PR also uses parameterized tests to assert on the presence of expected files after `pepr build --no-embed`.

### Update Consideration

`esbuild:0.25.0` omits legal comment output files when empty, see [esbuild/#3670](https://github.com/evanw/esbuild/issues/3670).

As a result, pepr now mirrors this behavior with output to `dist/`. It is unlikely that this will cause issues but is something users should be aware of, depending upon what they care about in build artifacts.

## Related Issue

Security hotfix.

## Type of change

## Checklist before merging

### DIFF
--- a/integration/cli/build.noembed.test.ts
+++ b/integration/cli/build.noembed.test.ts
@@ -8,7 +8,6 @@ import { existsSync } from "node:fs";
 import { Workdir } from "../helpers/workdir";
 import * as time from "../helpers/time";
 import * as pepr from "../helpers/pepr";
-import * as resource from "../helpers/resource";
 import { Result } from "../helpers/cmd";
 
 const FILE = path.basename(__filename);
@@ -25,8 +24,6 @@ describe("build", () => {
     const id = FILE.split(".").at(1);
     const testModule = `${workdir.path()}/${id}`;
     let buildOutput: Result;
-    let packageJson;
-    let uuid: string;
 
     beforeAll(async () => {
       await fs.rm(testModule, { recursive: true, force: true });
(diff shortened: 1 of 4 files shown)


samayer12:
Integration tests read as shown in the below snippet. `UUID` is a placeholder that's swapped with the appropriate regex inside the test body. This makes for shorter test titles for each case, and does not explicitly print the long UUID regex to the terminal. For example:

```
  build
    when building a module
      ✓ should execute 'pepr build' (5 ms)
      for use as a library
        ✓ should create: 'pepr.d.ts.map' (1 ms)
        ✓ should create: 'pepr.d.ts' (1 ms)
        ✓ should create: 'pepr.js.map' (1 ms)
        ✓ should create: 'pepr.js'
        ✕ should not create: '^UUID-chart/$'
        ✓ should not create: '^pepr-UUID\.js\.map$' (1 ms)
        ✓ should not create: '^pepr-UUID\.js$'
        ✓ should not create: '^pepr-module-UUID\.yaml$' (2 ms)
        ✓ should not create: '^zarf\.yaml$' (1 ms)
        ✕ should not create: '^pepr-UUID\.js\.LEGAL\.txt$'
        ✓ should not create: '^pepr\.js\.LEGAL\.txt$' (1 ms)
```

The hypothetical failing cases are marked with `✕` and provide the assertion string that was used in the failure output. When investigating a failing test case, the specific regex in use will be apparent in the test body.

cmwylie19 (approved):
looks good, passes tests, mitigates an esbuild CVE. Also, nice regex replacements and testing. Manual testing seems good too

PR #1830, opened 2025-02-13
chore: rename api-token to api-path

### LINKED ISSUE #1785: Remove or rename `apiToken` from webhooks

A [community question](https://kubernetes.slack.com/archives/C06DGH40UCB/p1738954927904269) asked if `apiToken` is something that needs to be sanitized from the helm chart that is produced after running `pepr build`.

This might be a case of a lingering feature that causes confusion for users of Pepr. We could consider renaming it to reflect the non-sensitive nature of the `apiToken` field or removing it altogether. First, we'll have to research why it might have been introduced into the project.

cmwylie19:
## Description

We noticed users in the open source being confused about the rhetoric around `api-token`. The `token` in this sense is a random path to get to the endpoints for mutate/validate

## Related Issue

Fixes #1785 

Relates to #

## Type of change

## Checklist before merging


samayer12 (changes requested):
Would it be more appropriate to make `api-path` a `ConfigMap`?

I noticed that we're still storing `api-path` as a k8s `secret`.  The k8s docs seem to suggest that may be the better approach here.

> Secrets are similar to [ConfigMaps](https://kubernetes.io/docs/concepts/configuration/configmap/) but are specifically intended to hold confidential data.

and 

> ConfigMap does not provide secrecy or encryption. If the data you want to store are confidential, use a [Secret](https://kubernetes.io/docs/concepts/configuration/secret/) rather than a ConfigMap, or use additional (third party) tools to keep your data private.

Alternatively, I could see this as _just_ a user-facing rename and then we create another issue to convert `api-path` to a `ConfigMap` as a separate PR.

cmwylie19:
> Would it be more appropriate to make `api-path` a `ConfigMap`?
> 
> I noticed that we're still storing `api-path` as a k8s `secret`. The k8s docs seem to suggest that may be the better approach here.
> 
> > Secrets are similar to [ConfigMaps](https://kubernetes.io/docs/concepts/configuration/configmap/) but are specifically intended to hold confidential data.
> 
> and
> 
> > ConfigMap does not provide secrecy or encryption. If the data you want to store are confidential, use a [Secret](https://kubernetes.io/docs/concepts/configuration/secret/) rather than a ConfigMap, or use additional (third party) tools to keep your data private.
> 
> Alternatively, I could see this as _just_ a user-facing rename and then we create another issue to convert `api-path` to a `ConfigMap` as a separate PR.

Yeah if you don't mind i would say _how_ we are holding the data is out of scope for this renaming story but I agree that is doesn't really need to be base64 encoded.

samayer12:
> Yeah if you don't mind i would say _how_ we are holding the data is out of scope for this renaming story but I agree that is doesn't really need to be base64 encoded.

Concur, I'll write up another issue. See #1843

samayer12 (approved):

PR #1845, opened 2025-02-19
chore: ensure generateAllYaml generates all kube manifests

### LINKED ISSUE #1764: Provide test coverage for generateAllYaml in `src/lib/assets/yaml`

After doing some work on the assets class it became obvious that we are lacking coverage in some of the functions. We need to increase coverage in `generateAllYaml` to ensure correctness

cmwylie19:
## Description

This test ensures that generateAllYaml calls the correct functions in order to generate Pepr's Kubernetes manifests.

## Related Issue

Fixes #1764 

Relates to #

## Type of change

## Checklist before merging

### DIFF
--- a/src/lib/assets/yaml/generateAllYaml.test.ts
+++ b/src/lib/assets/yaml/generateAllYaml.test.ts
@@ -1,0 +1,127 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2023-Present The Pepr Authors
+
+import { expect, describe, it, jest } from "@jest/globals";
+import { generateAllYaml } from "./generateAllYaml";
+import { Assets } from "../assets";
+import { ModuleConfig } from "../../types";
+import { V1Deployment, KubernetesObject } from "@kubernetes/client-node";
+import { promises as fs } from "fs";
+import { webhookConfigGenerator } from "../webhooks";
+import { WebhookType } from "../../enums";
+import { getModuleSecret, getNamespace } from "../pods";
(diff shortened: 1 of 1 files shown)


samayer12 (changes requested):
We have a _lot_ of mocking here. The tests are mostly what I'd refer to as "contract tests" that just make sure some function was called. #1853 shows how we can actually set `mockReturnValue()` to `{}` for most cases and have the tests still pass.

Could we instead make assertions on the `Promise<string>` returned by `generateAllYaml()`?

cmwylie19:
> We have a _lot_ of mocking here. The tests are mostly what I'd refer to as "contract tests" that just make sure some function was called. #1853 shows how we can actually set `mockReturnValue()` to `{}` for most cases and have the tests still pass.
> 
> Could we instead make assertions on the `Promise<string>` returned by `generateAllYaml()`?

I think I am confused by that PR, the test is dedicated to ensuring that the 

> We have a _lot_ of mocking here. The tests are mostly what I'd refer to as "contract tests" that just make sure some function was called. #1853 shows how we can actually set `mockReturnValue()` to `{}` for most cases and have the tests still pass.
> 
> Could we instead make assertions on the `Promise<string>` returned by `generateAllYaml()`?

Is in case, the mocked functions are already well tested, `generateAllYaml` is a function that relies on these contracts to hold up their part of the job. This test is to make sure the correct functions and all the right functions are called with the right data from the config object.

cmwylie19:
i understand what you are saying now, no point in returning things from mocks. I will try this. makes sense

samayer12 (approved):
Had a sync discussion, we're good to go!

PR #1873, opened 2025-02-24
chore: watch-processor support watching generic kind

### LINKED ISSUE #1838: GenericKind cannot be watched

```ts
When(a.GenericKind, {
  group: "pepr.dev",
  version: "v1",
  kind: "Unicorn",
})
  .IsCreated()
  .WithName("example-1")
  .Watch(() => Log.info("Unicorn Created"));
```

Seems like the watch URL is being called with GenericKind, if we do the GVK map and it is undefined than there is more work to do in order to know what to watch.

```bash
[08:50:49.696] DEBUG (95315): Watch event network_error received. watch connect failed: 404 Not Found.
```

cmwylie19:
## Description

We support doing mutates/validates on GenericKinds by allowing the user to add in the GVK manually as the second argument to the `When()`. The problem occurs when the user then tries to watch a resource through `.Watch() or .Reconcile()` then the KFC does not know how to lookup the generic kind because the second argument is not passed to the [k8s function](https://github.com/defenseunicorns/kubernetes-fluent-client/blob/00a7be6c71fb39855e1e2329850c5fa868979576/src/fluent/index.ts#L26). 

The second point here is that the same GVK look up happening twice, [once](https://github.com/defenseunicorns/pepr/blob/fe7f40df5e062d6b64fd65418a310ce35e2eb975/src/lib/core/capability.ts#L184) in the Capability class when the modules are being transformed into capability exports, and then again when Pepr calls the KFC watch (and KFC watch does a second [lookup](https://github.com/defenseunicorns/kubernetes-fluent-client/blob/00a7be6c71fb39855e1e2329850c5fa868979576/src/fluent/index.ts#L29)). This is an important backstop for KFC since it is its own project and does not rely on Pepr, but in this case, we already have all of the information.

The K8s KFC function takes filters that account for `kindOverride`, by adding kindOverride as the kind, then we pass the information from the first GVK lookup OR the GVK from the binding directly to the K8s function. There is no need to consider the kind of the Kubernetes object, whether it is a GenericKind or well known, since the kindOverrides filter will always have the appropriate information, whether from the GVK map or from the GVK argument in `When()`.  

(Capability adds correct types to binding kind [here](https://github.com/defenseunicorns/pepr/blob/fe7f40df5e062d6b64fd65418a310ce35e2eb975/src/lib/core/capability.ts#L194))

```ts
When(a.GenericKind, {
  group: "pepr.dev",
  version: "v1",
  kind: "Unicorn",
})
```

- [x] **Needs excellent example coverage** [PEXEX](https://github.com/defenseunicorns/pepr-excellent-examples/pull/237)

```bash
 PASS  capabilities/genericKind.e2e.test.ts (46.82 s)
  genericKind.ts
    ✓ watches resource creates (542 ms)
    ✓ watches resource create-or-updates (544 ms)
    ✓ watches resource updates (568 ms)
    ✓ watches resource deletes (527 ms)
 ```

image
image

## Related Issue

Fixes #1838 

Relates to #

## Type of change

## Checklist before merging


samayer12 (approved):
Ensures Pepr interacts with KFC and resolves a watch issue, good work!

PR #1891, opened 2025-03-03
chore: increase coverage in validate processor

### LINKED ISSUE #1890: increase test coverage in validate-processor

Validate processor is missing tests.

```bash
validate-processor.ts      |   53.48 |    35.71 |      50 |   53.48 | 63-101 
```

We need to increase coverage to ensure the right functions are being called during calls to `validateProcessor` to ensure the the requests will be skipped if the criteria is wrong, the requests are measured in terms of the webhookTimeout, and secrets are converted from base64.

cmwylie19:
## Description

We need to increase coverage in validateProcessor.
Ensure that:
- the requests will be skipped if the filter criteria is wrong
- the requests are measured in terms of the webhookTimeout
-  secrets are converted from base64.

## Related Issue

Fixes #1890 

Relates to #

## Type of change

## Checklist before merging

### DIFF
--- a/src/lib/processors/validate-processor.test.ts
+++ b/src/lib/processors/validate-processor.test.ts
@@ -7,7 +7,34 @@ import { AdmissionRequest, Binding, Filters } from "../types";
 import { Event, Operation } from "../enums";
 import { PeprValidateRequest } from "../validate-request";
 import { clone } from "ramda";
-import * as sut from "./validate-processor";
+import { processRequest, validateProcessor } from "./validate-processor";
+import { ModuleConfig } from "../types";
+import { Capability } from "../core/capability";
+import { MeasureWebhookTimeout } from "../telemetry/webhookTimeouts";
+import * as utils from "../utils";
+import { shouldSkipRequest } from "../filter/filter";
+
+jest.mock("../telemetry/logger", () => ({
+  info: jest.fn(),
+  debug: jest.fn(),
+  error: jest.fn(),
+}));
+
+jest.mock("../telemetry/metrics", () => ({
+  metricsCollector: {
+    addCounter: jest.fn(),
+    incCounter: jest.fn(),
+  },
+  MeasureWebhookTimeout: jest.fn(),
+}));
+
+jest.mock("../telemetry/timeUtils", () => ({
+  getNow: jest.fn(() => 1000),
+}));
+
+jest.mock("../filter/filter", () => ({
+  shouldSkipRequest: jest.fn(),
+}));
 
 const testFilters: Filters = {
   annotations: {},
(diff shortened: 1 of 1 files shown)


src/lib/processors/validate-processor.test.ts
samayer12, line 1: Since the mocks used in tests are scoped to the test-level instead, we don't need to call `mockRestore()`. Let's remove that call after our assertions since it doesn't really do anything

samayer12, line 1: `validate-processor.ts`'s `processRequest()` function is only ever called by `validateProcessor()`, which means it's internal to the behavior of the `validateProcessor()` function.

We should avoid testing function implementations in favor of testing the behavior of the publicly available function (i.e., `validateProcessor()`). It looks like there are the following execution paths in `processRequest()` that we can test by changing inputs to `validateProcessor()`:

1. callback receives a status code from `binding.validateCallback()`
2. callback receives no status code from `binding.validateCallback()`
3. an exception is thrown by `binding.validateCallback()` (e.g., the non-null assertion when we set `callbackResp` could fail)

`binding` comes from `capabilities`, which is an input to `validateProcessor()`, so I think the right inputs are exposed by that function so that we can get branch coverage without being tied to `processRequest()`'s implementation details.

samayer12 (changes requested):
`validate-processor.ts`'s `processRequest()` function is only ever called by `validateProcessor()`, which means it's internal to the behavior of the `validateProcessor()` function.

We should avoid testing function implementations in favor of testing the behavior of the publicly available function (i.e., `validateProcessor()`). It looks like there are the following execution paths in `processRequest()` that we can test by changing inputs to `validateProcessor()`:

1. callback receives a status code from `binding.validateCallback()`
2. callback receives no status code from `binding.validateCallback()`
3. an exception is thrown by `binding.validateCallback()` (e.g., the non-null assertion when we set `callbackResp` could fail)

`binding` comes from `capabilities`, which is an input to `validateProcessor()`, so I think the right inputs are exposed by that function so that we can get branch coverage without being tied to `processRequest()`'s implementation details.

cmwylie19:
> `validate-processor.ts`'s `processRequest()` function is only ever called by `validateProcessor()`, which means it's internal to the behavior of the `validateProcessor()` function.
> 
> We should avoid testing function implementations in favor of testing the behavior of the publicly available function (i.e., `validateProcessor()`). It looks like there are the following execution paths in `processRequest()` that we can test by changing inputs to `validateProcessor()`:
> 
> 1. callback receives a status code from `binding.validateCallback()`
> 2. callback receives no status code from `binding.validateCallback()`
> 3. an exception is thrown by `binding.validateCallback()` (e.g., the non-null assertion when we set `callbackResp` could fail)
> 
> `binding` comes from `capabilities`, which is an input to `validateProcessor()`, so I think the right inputs are exposed by that function so that we can get branch coverage without being tied to `processRequest()`'s implementation details.

Okay sounds good, will look to make assertions on the result of validateProcessor as well and add different inputs.

samayer12 (approved):
Had a sync discussion about problems around test-data generation that for `Capability` objects that contain `Bindings` that make it hard to test in the way requested. Current implementation is good enough for now!

PR #1895, opened 2025-03-05
chore: account for early exit and stop timer

### LINKED ISSUE #1894: Mutate Processor `mutation_timeouts` metrics does not account for early exit
### Environment
Device and OS:
App version:
Kubernetes distro being used:
Other:

### Steps to reproduce
There are several places in the mutate process where the request will exit early before the final return of the response. This leads to metricCollector incrementing the count for mutation timeouts which is inaccurate. We need to stop the timer to ensure we have accurate metrics.

### Expected result
If an early exit occurs, we need to stop the timer.

### Actual Result
It is not stopped and the mutation_timeout metric is incremented.

### Additional Context

cmwylie19:
## Description

There are several places in the mutate process where the request will exit early before the final return of the response. This leads to metricCollector incrementing the count for mutation timeouts which is inaccurate. We need to stop the timer to ensure we have accurate metrics.

## Related Issue

Fixes #1894 

Relates to #

## Type of change

## Checklist before merging

### DIFF
--- a/src/lib/processors/mutate-processor.ts
+++ b/src/lib/processors/mutate-processor.ts
@@ -186,6 +186,7 @@ export async function mutateProcessor(
   for (const bindable of bindables) {
     ({ wrapped, response } = await processRequest(bindable, wrapped, response));
     if (config.onError === OnError.REJECT && response?.warnings!.length > 0) {
+      webhookTimer.stop();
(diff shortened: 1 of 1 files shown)


samayer12 (approved):
Relates to #1897

PR #1897, opened 2025-03-05
chore: webhook timer coverage in mutate-processor test

### LINKED ISSUE #1892: increase coverage in mutate-processor

mutate processor is missing tests.

```bash
  mutate-processor.ts        |   64.89 |       75 |      60 |   65.59 | 143-217  
```

We need to increase coverage to ensure the right functions are being called during calls to `mutateProcessor` to ensure the the requests will be skipped if the criteria is wrong, the requests are measured in terms of the webhookTimeout, and secrets are converted from base64.

cmwylie19:
## Description

This is very hard to test because the Capability does not have actual bindings. Because of that the test stops [here](https://github.com/defenseunicorns/pepr/blob/4e8549a43bb911072bd0316f7dace7172e5eb1a0/src/lib/processors/mutate-processor.ts#L197) and returns early. The only meaningful tests I could do was to ensure the timer start and stop were called. 

There is already too much mocking in this file I cannot get any assertions to pass for base64 calls here. 

**UPDATE**: was able to validate the first decodeData was called by shuffling files around and mocking the call in `mutate-process.test.ts`. The rest of the test is untestable due to no bindings in the capabilities

Needs https://github.com/defenseunicorns/pepr/pull/1895 to pass.

## Related Issue

Fixes #1892 

Relates to #

## Type of change

## Checklist before merging

### DIFF
--- a/src/lib/decode-utils.test.ts
+++ b/src/lib/decode-utils.test.ts
@@ -1,0 +1,122 @@
+import { beforeEach, describe, expect, it, jest } from "@jest/globals";
+import { AdmissionRequest } from "./types";
+import { convertFromBase64Map, convertToBase64Map } from "./utils";
+import { Operation } from "./enums";
+import { PeprMutateRequest } from "./mutate-request";
+import { decodeData, reencodeData } from "./decode-utils";
+
+jest.mock("./utils");
+
+const defaultPeprMutateRequest = (admissionRequest = defaultAdmissionRequest) =>
+  new PeprMutateRequest(admissionRequest);
+
+const mockConvertToBase64Map = jest.mocked(convertToBase64Map);
+const mockConvertFromBase64Map = jest.mocked(convertFromBase64Map);
+
+const defaultAdmissionRequest: AdmissionRequest = {
+  uid: "uid",
+  kind: {
+    kind: "kind",
+    group: "group",
+    version: "version",
+  },
+  resource: {
+    group: "group",
+    version: "version",
+    resource: "resource",
+  },
+  name: "",
+  object: {
+    metadata: {
+      name: "create-me",
+    },
+  },
+  operation: Operation.CREATE,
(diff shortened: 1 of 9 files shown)


src/lib/processors/mutate-processor.ts
samayer12, line 1: What's the deal with the commented out chunks here? Can we remove them?

src/lib/decode-utils.test.ts
```diff
@@ -0,0 +1,122 @@
...
+  },
+  name: "",
+  object: {
+    metadata: {
+      name: "create-me",
+    },
+  },
+  operation: Operation.CREATE,
+  userInfo: {},
+};
```
samayer12: I see that we have several test files that use a `defaultAdmissionRequest` and we define a constant in `src/lib/filter/adjudicators/defaultTestObjects.ts`.

Do you think it's appropriate to switch to using a generator function (as seen in #1898)  as part of this PR since that's our preferred pattern or should we write that up as a follow-on issue?

src/lib/decode-utils.ts
samayer12, line 1: This file is only used by `processors/mutate-processor.ts` so I think we should also store this file in `processors/`

src/lib/decode-utils.test.ts
cmwylie19: Yep that will work

src/lib/processors/mutate-processor.ts
cmwylie19: oversight, yes

src/lib/decode-utils.test.ts
cmwylie19: This is actually not new code, it is copied out of mutate-processor but we can go ahead and update it in this case.

samayer12 (approved):

PR #1934, opened 2025-03-13
chore: new org username and password

cmwylie19:
## Description

We have an org username and password for chainguard that I was not aware of. This should allow us to authenticate to their registry like they do in [renovate](https://github.com/defenseunicorns/uds-common/blob/6a3348a1d747ff49c69b5edf9fb56be24bc4840c/config/renovate.json5#L49)


soltysh (approved):
lgtm

PR #1952, opened 2025-03-19
chore: consolidate prettier config to one file for consistent behavior

### LINKED ISSUE #1900: Investigate presence of two `.pretterrc` files
We have two different `.prettierrc` files in this project. We should understand why we maintain two configuration files for the same tool. If unnecessary, then we should consolidate down to just one `.prettierrc` file.

See:
* `./.prettierrc`
* `./src/lib/.prettierrc`

### Additional context

We sometimes use different formatting/linting standards between "source code" and "test code", but it seems odd to have a top-level config and a specific config for `src/`. This is because we prefer all code to follow the top-level config unless otherwise justified (e.g., a piece of test code that warrants a different rule set).

samayer12:
## Description

We had two different `.prettierrc` files which overrode each other on the `bracketSameLine` rule that was in `src/lib/.prettierrc`. This PR uses one `.prettierrc` file for the entire project and applies a consistent ruleset.

It appears that `src/lib/` preferred the use of `bracketSameLine` but the rule was not being applied due to `bracketSameLine` being `false` in the top-level config. We may as well be consistent across the entire project, so this PR enables that rule.

Note that this also required changes to the `src/templates/prettierrc.json` template that new pepr modules use. This is not expected to create any breaking changes to existing users of Pepr since the template is used at module creation; new pepr modules will use a different `.prettierrc` file. It may be worth noting this change to users, tbd.

## Related Issue

Fixes #1900 

## Type of change

## Checklist before merging


cmwylie19 (approved):
nice job consolidating our linters

PR #1963, opened 2025-03-20
feat: override uuid during init

cmwylie19:
## Description

This allows users to override the randomly generated UUID created during `npx pepr init`. Overriding this field helps clearly identify which module Pepr pods belong to, especially in environments with multiple modules.

Missing:
- [x] Update Tests
- [x] Explicit Documentation that UUID is a unique identifier 

## Related Issue

Fixes #1962 

Relates to #

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Other (security config, docs update, etc)

## Checklist before merging
- [ ] Unit, [Journey](https://github.com/defenseunicorns/pepr/tree/main/journey), [E2E Tests](https://github.com/defenseunicorns/pepr-excellent-examples), [docs](https://github.com/defenseunicorns/pepr/tree/main/docs), [adr](https://github.com/defenseunicorns/pepr/tree/main/adr) added or updated as needed
- [ ] [Contributor Guide Steps](https://docs.pepr.dev/main/contribute/#submitting-a-pull-request) followed

PR #1979, opened 2025-03-27
chore: use uuidv4() when user does not provide a module UUID

### LINKED ISSUE #1973: Generate UUIDs with `uuidv4()` instead of a mixture of calls

This was part of the old implementation, but it looks odd to use `uuidv5()` and `uuidv4()` at the same time. See [UUID versions explained](https://www.sohamkamani.com/uuid-versions-explained/). That article has some suggestions on which UUID to use when:

> If you don’t know what to go with, go with v4. It’s good enough, and the chances of collision are practically none.
> 
> If you actually want your UUID to give some indication of the date and computer in which it was created, then UUID v1 may be for you (although it is).
> 
> UUID v5 is normally used only for very specific use cases, when you want to derive a UUID from another piece of information on the fly.

`src/cli/init/templates.ts` calls `const uuid = uuidv5(opts.name, uuidv4());` which seems like a circuitous way to generate a UUID.

Instead, we should use: `const uuid = !opts.uuid ? uuidv4() : opts.uuid;` to create the Pepr UUID when a user does not specify it.

None. 

### Additional context

See also, #1963

samayer12:
## Description

This module resolves some convoluted uuid generation. See also, #1963 

## Related Issue

Fixes #1973

## Type of change

## Checklist before merging

### DIFF
--- a/src/cli/init/templates.ts
+++ b/src/cli/init/templates.ts
@@ -3,7 +3,7 @@
 
 import { dumpYaml } from "@kubernetes/client-node";
 import { inspect } from "util";
-import { v4 as uuidv4, v5 as uuidv5 } from "uuid";
+import { v4 as uuidv4 } from "uuid";
 
 import eslintJSON from "../../templates/.eslintrc.template.json";
 import peprSnippetsJSON from "../../templates/pepr.code-snippets.json";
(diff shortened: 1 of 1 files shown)


soltysh (approved):
lgtm

PR #2027, opened 2025-04-09
chore: removes the `migrateAndSetupWatch` out of `store.ts` for improved testability

### LINKED ISSUE #1324: Extract cache migration code out of `store.ts`

Related to #959 and #1323. `store.ts` has some hard-to-test code for migrating values when updating the store cache. Find a way to extract the migration logic that's part of `PeprControllerStore.#migrateAndSetupWatch()` into an isolated function that can be maintained independently.

Start this issue _after_ #1259 is merged.

Reference branch [1324-migration-refactor-reference](https://github.com/defenseunicorns/pepr/tree/1324-migration-refactor-reference) for a starting point. Examine `storeCache.ts` and `storeCache.test.ts` to start out.

### Additional context

You'll need to examine the type definitions and manage differences between cache entries of type `Record<string, Operation>` and `DataStore` entries which are defined as `Record<string, string>`.

cmwylie19:
## Description

removes the `migrateAndSetupWatch` out of `store.ts` for improved testability

## Related Issue

Fixes #1324 

Relates to #

## Type of change

## Checklist before merging

### DIFF
--- a/src/lib/controller/migrateStore.test.ts
+++ b/src/lib/controller/migrateStore.test.ts
@@ -1,0 +1,103 @@
+import { migrateAndSetupWatch, StoreMigration } from "./migrateStore";
+import Log from "../telemetry/logger";
+import * as storeCache from "./storeCache";
+import { describe, it, expect, jest, beforeEach } from "@jest/globals";
+import type { Operation } from "fast-json-patch";
+import { Storage } from "../core/storage";
+
+const mockPatch = jest.fn();
+jest.mock("../telemetry/logger", () => ({
+  __esModule: true,
+  default: {
+    info: jest.fn(),
+    debug: jest.fn(),
+  },
+  redactedStore: jest.fn(x => x),
+}));
+
+jest.mock("kubernetes-fluent-client", () => {
+  return {
+    ...(jest.requireActual("kubernetes-fluent-client") as object),
+    K8s: jest.fn().mockImplementation(() => ({
+      Patch: mockPatch,
+    })),
+  };
+});
+jest.mock("./storeCache", () => ({
+  ...(jest.requireActual("./storeCache") as object),
+  sendUpdatesAndFlushCache: jest.fn(),
+}));
+
+describe("migrateAndSetupWatch", () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it("should patch the store, transform keys, and call setupWatch", async () => {
+    const setupWatch = jest.fn();
+
+    // non-migrated store v1
+    const storeMigrationData: StoreMigration = {
+      name: "pepr-static-test-store",
+      namespace: "pepr-system",
+      setupWatch,
+      store: {
+        kind: "PeprStore",
+        data: {
+          __pepr_do_not_delete__: "k-thx-bye",
+          "hello-pepr-example-1": "was-here",
+          "hello-pepr-example-1-data": '{"key":"ex-1-val"}',
+          "hello-pepr-watch-data": "This data was stored by a Watch Action.",
+        },
+        metadata: {
+          labels: {
+            "pepr.dev-cacheID": "1744214317282",
+          },
+          name: "pepr-static-test-store",
+          namespace: "pepr-system",
+        },
+      },
+      stores: {
+        "hello-pepr-example-1": new Storage(),
+      },
+    };
+
+    (
+      storeCache.sendUpdatesAndFlushCache as jest.MockedFunction<
+        typeof storeCache.sendUpdatesAndFlushCache
+      >
+    ).mockResolvedValue({} as Record<string, Operation>);
+
+    await migrateAndSetupWatch(storeMigrationData);
+    const [actualCache, actualNamespace, actualName] = (
+      storeCache.sendUpdatesAndFlushCache as jest.Mock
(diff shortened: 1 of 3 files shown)


src/lib/controller/store.ts
```diff
@@ -56,7 +57,16 @@ export class StoreController {
...
           .Get(this.#name)
-          .then(async (store: Store) => await this.#migrateAndSetupWatch(store))
+          .then(
+            async (store: Store) =>
+              await migrateAndSetupWatch({
+                name,
+                namespace,
+                store,
+                stores: this.#stores,
+                setupWatch: this.#setupWatch,
```
samayer12, line 67: We used to pass in a `Store` object, and now we pass in a subset of properties pulled from that object. I assume this was done so that we don't need to create a full mock of a `Store` object in the unit test, right?

cmwylie19: More of a superset. We still pass in store but the other properties that `migrateAndSetupWatch` used were local to the class which it lost access to once it moved.

src/lib/controller/migrateStore.test.ts
```diff
@@ -0,0 +1,103 @@
...
+  ...(jest.requireActual("./storeCache") as object),
+  sendUpdatesAndFlushCache: jest.fn(),
+}));
+
+describe("migrateAndSetupWatch", () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it("should patch the store, transform keys, and call setupWatch", async () => {
```
samayer12: As written, this unit test is tightly-coupled with implementation details. What do you think about splitting it into three test cases? Maybe something along the lines of

* it should patch the store
* it should transform keys
* it should call setupWatch

src/lib/controller/migrateStore.ts
```diff
@@ -0,0 +1,56 @@
...
+  // Add cacheID label to store
+  await K8s(Store, { namespace, name }).Patch([
+    {
+      op: "add",
+      path: "/metadata/labels/pepr.dev-cacheID",
+      value: `${Date.now()}`,
+    },
+  ]);
+
+  const data: DataStore = store.data || {};
```
samayer12: We could add a test case here to capture the behavior when `store.data` is `{}`. Something along the lines of:

* it should not transform keys when no store data is provided

and then assert upon `transformedKeys` in the test?

src/lib/controller/migrateStore.ts
```diff
@@ -0,0 +1,56 @@
...
+          value: data[key],
+        });
+        storeCache = fillStoreCache(storeCache, name, "add", {
+          key: [key.slice(offset)],
+          value: data[key],
+          version: "v2",
+        });
+      }
+    }
+  }
```
samayer12, line 53: I suggest extracting this to its own function so that we can test the conditional logic within independent of the rest of the migration function.

Something along the lines of `let storeCache = transformKeys($PARAMS)` where `transformKeys(): Record<string, Operation>` is the function signature.

samayer12 (changes requested):
This PR extracts the desired function, I think we have some improvements to the unit tests that we can achieve before merging this PR.

src/lib/controller/migrateStore.ts
cmwylie19: This is actually a condition that cannot occur ever because stores is initialized to an empty object. I removed that or.

```ts

export class StoreController {
  #name: string;
  #stores: Record<string, Storage> = {};
```

src/lib/controller/migrateStore.ts
cmwylie19: I'd say this branch of the code is thoroughly tested to the point where the unit test would break since we are asserting on the actual values for the JSON Patch.

```ts
    expect(transformedKeys).toEqual(
      expect.arrayContaining([
        expect.stringMatching("remove:/data/hello-pepr-example-1-"),
        expect.stringMatching("add:/data/hello-pepr-example-1-v2-:was-here"),
        expect.stringMatching("remove:/data/hello-pepr-example-1-data"),
        expect.stringMatching('add:/data/hello-pepr-example-1-v2-data:{"key":"ex-1-val"}'),
      ]),
    );
```

cmwylie19: would break in the event that there were a regression*

samayer12: Cool, as long as we have a way to catch that stuff with detailed feedback it works for me

src/lib/controller/migrateStore.test.ts
```diff
@@ -97,7 +94,41 @@ describe("migrateAndSetupWatch", () => {
...
+      },
+    };
+    await migrateAndSetupWatch(emptyStoreMigrationData);
+    const [actualCache] = (storeCache.sendUpdatesAndFlushCache as jest.Mock).mock.calls[0] as [
+      Record<string, Operation>,
+    ];
+    expect(actualCache).toEqual({});
+  });
+
+  it("should log that the store is being migrated to easier debuggin through the logs", async () => {
```
samayer12:
```suggestion
  it("should log that the store is being migrated for easier debugging through the logs", async () => {
```

typo fix

samayer12 (dismissed):

samayer12 (approved):

Issue #2023, opened 2025-04-08
[SPIKE] - Look into TS-Morph for Generating CRDs from TypeScript types into the `openAPIV3Schema`


cmwylie19:
### Describe what should be investigated or refactored

Currently we are parsing raw files with Regex for the `npx pepr crd generate` function. There is a library called [ts-morph](https://github.com/dsherret/ts-morph) which TypeScript Compiler API wrapper for static analysis and programmatic code changes that may simplify this process.

cmwylie19:
closed in https://github.com/defenseunicorns/pepr/pull/2033

PR #2017, opened 2025-04-05
feat: generate CRDs from TypeScript types

cmwylie19:
## Description

Adds CLI commands for generating CRDs from TypeScript types, Integration and unit tests. This PR is already big so the docs will be part of a separate issue. 

It creates parses raw TypeScript files and generates CRDs from them.

I think we may need to supress some of the eslint stuff and create issues to work on it for CI purposes. This feature is scoped _just_ to the two commands. `npx pepr crd create/generate`

Usage:

```bash
npx ts-node src/cli.ts crd create --group cache --version v1alpha1 --kind Memcache --shortName mc --plural memcaches --scope Namespaced

npx ts-node src/cli.ts crd generate
kubectl apply -f crds

kubectl apply -f -<<EOF
apiVersion: cache.pepr.dev/v1alpha1
kind: Memcache
metadata:
  name: memcached
  namespace: default
spec:
  size: [3,3]
  containerPort: 8080
  config: 
    language: ["en", "fr"]
    timezone: 3
    zone: 
      state: "UTC"
      areaCode: ["ok"]
EOF
```

## Related Issue

Fixes #1062 

Relates to #2023 

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Other (security config, docs update, etc)

## Checklist before merging
- [x] Unit, [Journey](https://github.com/defenseunicorns/pepr/tree/main/journey), [E2E Tests](https://github.com/defenseunicorns/pepr-excellent-examples), [docs](https://github.com/defenseunicorns/pepr/tree/main/docs), [adr](https://github.com/defenseunicorns/pepr/tree/main/adr) added or updated as needed
- [x] [Contributor Guide Steps](https://docs.pepr.dev/main/contribute/#submitting-a-pull-request) followed

PR #2033, opened 2025-04-13
feat: pepr to scaffold and generate crds from typescript types

### LINKED ISSUE #1062: CRD generation

It is frustrating having to create your own CRD. We need a way to scaffold and create a CRD.

- **Given** I create a class and want to generate a CRD
- **When** I issue the generate command
- **Then** a CRD is created

cmwylie19:
## Description

Adds CLI commands for generating CRDs from TypeScript types, Integration and unit tests. This PR is already big so the docs will be part of a separate issue. 

It creates parses raw TypeScript files and generates CRDs from them.

I think we may need to supress some of the eslint stuff and create issues to work on it for CI purposes. This feature is scoped _just_ to the two commands. `npx pepr crd create/generate`

Usage:

```bash
npx ts-node src/cli.ts crd create --group cache --version v1alpha1 --kind Memcache --shortName mc --plural memcaches --scope Namespaced

npx ts-node src/cli.ts crd generate
kubectl apply -f crds

kubectl apply -f -<<EOF
apiVersion: cache.pepr.dev/v1alpha1
kind: Memcache
metadata:
  name: memcached
  namespace: default
spec:
  size: [3,3]
  containerPort: 8080
  config: 
    language: ["en", "fr"]
    timezone: 3
    zone: 
      state: "UTC"
      areaCode: ["ok"]
EOF
```

## Related Issue

Fixes #1062 

Relates to #2023 

## Type of change

## Checklist before merging


cmwylie19:
supersedes #2017

src/cli.ts
```diff
@@ -25,6 +26,7 @@ if (!process.env.PEPR_NODE_WARNINGS) {
   process.removeAllListeners("warning");
 }
 program
+  .enablePositionalOptions()
```
samayer12, line 29: Why do we enable positional options? I have an opinion that flagged options reduce headache, but am open to the idea.

src/cli/crd/create.test.ts
```diff
@@ -0,0 +1,53 @@
...
+    expect(result).toContain(`${kind}Spec`);
+    expect(result).toContain(`${kind}Status`);
+    expect(result).toContain(`// Group: ${group}`);
+    expect(result).toContain(`// Version: ${version}`);
+    expect(result).toContain(`// Domain: ${data.domain}`);
+    expect(result).toContain(`plural: "${data.plural}"`);
+    expect(result).toContain(`scope: "${data.scope}"`);
+    expect(result).toContain(`shortName: "${data.shortName}"`);
+    expect(result).toContain(`${kind}StatusCondition`);
+    expect(result).toContain(`${kind}Status`);
```
samayer12, line 29: This is thorough, I like it. What do you think of making these strings a collection that we iterate over? Perhaps something along these lines:

```suggestion
const expectedSnippets = [
  // Kind-related
  `${kind}Spec`,
  `${kind}Status`,
  `${kind}StatusCondition`,
  `${kind}Status`,

  // Metadata comments
  `// Group: ${group}`,
  `// Version: ${version}`,
  `// Domain: ${data.domain}`,

  // CRD config fields
  `plural: "${data.plural}"`,
  `scope: "${data.scope}"`,
  `shortName: "${data.shortName}"`,
];

expect(expectedSnippets.every(snippet => result.includes(snippet))).toBe(true);  
```

The value here is that we show readers of the test what sorts of things we care about in a CRD definition (e.g., kind-related thing, metadata, and CRD config fields)

src/cli/crd/create.test.ts
```diff
@@ -0,0 +1,53 @@
...
+  it("should generate the correct CRD scaffold", () => {
+    const group = "example";
+    const version = "v1alpha1";
+    const kind = "MyCustomResource";
+    const data = {
+      domain: "example.com",
+      plural: "mycustomresources",
+      scope: "Namespaced",
+      shortName: "mcr",
+    };
```
samayer12, line 17: Thoughts on pre-emptively turning this into a generator function? Something along the lines of `createDefaultCRD()`? We've started to adopt that pattern elsewhere in our test suite.

src/cli/crd/create.test.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  it("should return the scope if it is valid - Cluster", () => {
+    const result = validateScope("Cluster");
+    expect(result).toBe("Cluster");
+  });
+
+  it("should return the scope if it is valid - Namespaced", () => {
+    const result = validateScope("Namespaced");
+    expect(result).toBe("Namespaced");
+  });
```
samayer12: Is it worth it to parameterize these two tests? There are only so many scope-able things in K8s

```suggestion
  it.each([
    ["Cluster"],
    ["Namespaced"],
  ])("should return the scope if it is valid - %s", (input) => {
    const result = validateScope(input);
    expect(result).toBe(input);
  });
```

src/cli/crd/generate.test.ts
```diff
@@ -0,0 +1,202 @@
...
+      const result = emptySchema();
+      expect(result).toEqual({ properties: {}, required: [] });
+    });
+  });
+
+  describe("getAPIVersions", () => {
+    it("should return directories from the api root", () => {
+      (fs.readdirSync as jest.Mock).mockReturnValue(["v1", "v2"]);
+      (fs.statSync as jest.Mock).mockImplementation(p => ({
+        isDirectory: () => typeof p === "string" && (p.endsWith("v1") || p.endsWith("v2")),
```
samayer12:
```suggestion
      (fs.statSync as jest.Mock).mockImplementation(path => ({
        isDirectory: () => typeof path === "string" && (path.endsWith("v1") || p.endsWith("v2")),
```

I assume `p` represents a filepath on the system,

src/cli/crd/generate.test.ts
```diff
@@ -0,0 +1,202 @@
...
+      }));
+
+      const versions = getAPIVersions("/api");
+      expect(versions).toEqual(["v1", "v2"]);
+    });
+
+    it("should ignore non-directory entries", () => {
```
samayer12, line 66: what happens if I have two directories that end in `v1` (e.g., `Av1`, `Bv1`)? I figure this probably _shouldn't_ happen, but could it? Perhaps a user is generating CRDs with Pepr and some other tool or manually copying things around (for some unknown reason).

src/cli/crd/generate.test.ts
```diff
@@ -0,0 +1,202 @@
...
+
+      const versions = getAPIVersions("/api");
+      expect(versions).toEqual(["v1"]);
+    });
+  });
+
+  describe("loadVersionFiles", () => {
+    it("should load only TypeScript files from version directory", () => {
+      const project = new Project();
+      const addSpy = jest.spyOn(project, "addSourceFilesAtPaths").mockReturnValue([]); // don't resolve real files
```
samayer12, line 80:
```suggestion
      const addSpy = jest.spyOn(project, "addSourceFilesAtPaths").mockImplementation(() => {});// don't resolve real files
```

If this is just a contract test, then we don't even need to bother with asserting return types

src/cli/crd/generate.test.ts
```diff
@@ -0,0 +1,202 @@
...
+
+      (fs.readdirSync as jest.Mock).mockReturnValue(["foo.ts", "bar.js", "baz.ts"]);
+
+      const result = loadVersionFiles(project, "/api/v1");
+
+      expect(addSpy).toHaveBeenCalledWith(["/api/v1/foo.ts", "/api/v1/baz.ts"]);
+      expect(result).toEqual([]); // we mocked return value
```
samayer12, line 87:
```suggestion
```

See above comment if this is just a "contract test" with the function that we call.

src/cli/crd/generate.test.ts
```diff
@@ -0,0 +1,202 @@
...
+    });
+  });
+
+  describe("extractDetails", () => {
+    const createProjectWithFile = (name: string, content: string) => {
+      const project = new Project();
+      return project.createSourceFile(name, content);
+    };
```
samayer12, line 95: This looks identical to the one in `processSourceFile`, thoughts on extracting this to a common generator function?

src/cli/crd/generate.test.ts
```diff
@@ -0,0 +1,202 @@
...
+      const file = createProjectWithFile("noDetails.ts", `const somethingElse = {};`);
+      expect(() => extractDetails(file)).toThrow(`Missing 'details' variable declaration.`);
+    });
+  });
+
+  describe("processSourceFile", () => {
+    const createProjectWithFile = (name: string, content: string) => {
+      const project = new Project();
+      return project.createSourceFile(name, content);
+    };
```
samayer12, line 136: This looks identical to the one in `extractDetails`, thoughts on extracting this to a common generator function?

src/cli/crd/generate.test.ts
```diff
@@ -0,0 +1,202 @@
...
+        expect.stringContaining("missing interface SomethingSpec"),
+      );
+    });
+
+    it("should generate a CRD YAML file for valid input", () => {
+      const writeFileSync = jest.spyOn(fs, "writeFileSync").mockImplementation(() => {});
+      const consoleLog = jest.spyOn(console, "log").mockImplementation(() => {});
+      const file = createProjectWithFile(
+        "valid.ts",
+        `
```
samayer12, line 175: Storing relatively simple files as string in a test is okay, but I wonder if we can store these as _actual_ files and read them in during a test. Maybe not required if we mock file read operations.

The thing I want to avoid is a large, bespoke test string like we've seen elsewhere in Pepr. One example of this anti-pattern is `rbac.test.ts` before we created `defaultTestObjects`. Thoughts?

src/cli/crd/generate.ts
```diff
@@ -0,0 +1,316 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2023-Present The Pepr Authors
+
+import { Command } from "commander";
+import fs from "fs";
+import path from "path";
+import { stringify as toYAML } from "yaml";
```
samayer12, line 7: That seems like an odd alias for a function. I wonder why it wasn't called `toYAML` in the first place...

src/cli/crd/generate.ts
```diff
@@ -0,0 +1,316 @@
...
+  Project,
+  InterfaceDeclaration,
+  TypeAliasDeclaration,
+  SyntaxKind,
+  Node,
+  SourceFile,
+  Type,
+} from "ts-morph";
+import { createDirectoryIfNotExists } from "../../lib/filesystemService";
+import { kind as k } from "kubernetes-fluent-client";
```
samayer12, line 18: Do we gain anything by calling it `k`? Why not just `kind` as presented in the export?

src/cli/crd/generate.ts
```diff
@@ -0,0 +1,316 @@
...
+export function extractCRDDetails(
+  content: string,
+  sourceFile: SourceFile,
+): {
+  kind: string | undefined;
+  fqdn: string;
+  scope: "Cluster" | "Namespaced";
+  plural: string;
+  shortNames?: string[];
+} {
```
samayer12, line 35: Should we create a `CRDdetails` type?

src/cli/crd/generate.ts
```diff
@@ -0,0 +1,316 @@
...
+
+  const fqdn = `${group}.${domain}`;
+
+  const { plural, scope } = details;
+
+  const shortNames = details.shortName ? [details.shortName] : undefined;
+  return { kind, plural, scope, shortNames, fqdn };
+}
+
+export async function generateCRDs(options: { output: string }): Promise<void> {
```
samayer12, line 49: I personally am wary of `Promise<void>` as a return type. Should we return something instead? Perhaps the filepath to where resources were created or some other success indicator?

Maybe something along the lines of `Promise<{ status: 'success', filePath?: string }>`?

src/cli/crd/generate.ts
```diff
@@ -0,0 +1,316 @@
...
+  if (!kind) {
+    console.warn(`⚠️ Skipping ${sourceFile.getBaseName()}: missing '// Kind: <KindName>' comment`);
+    return;
+  }
+
+  const spec = sourceFile.getInterface(`${kind}Spec`);
+  if (!spec) {
+    console.warn(`⚠️ Skipping ${sourceFile.getBaseName()}: missing interface ${kind}Spec`);
+    return;
+  }
```
samayer12:
```suggestion
  if (!kind) {
    console.warn(`Skipping ${sourceFile.getBaseName()}: missing '// Kind: <KindName>' comment`);
    return;
  }

  const spec = sourceFile.getInterface(`${kind}Spec`);
  if (!spec) {
    console.warn(`Skipping ${sourceFile.getBaseName()}: missing interface ${kind}Spec`);
    return;
  }
```

I think the use of emoji in Pepr log messages is inconsistent with the rest of the project's style? I only see emoji in some `console.info()` messages in the CLI and that's it.

src/cli/crd/generate.ts
```diff
@@ -0,0 +1,316 @@
...
+  const outPath = path.join(outputDir, `${kind.toLowerCase()}.yaml`);
+  fs.writeFileSync(outPath, toYAML(crd), "utf8");
+  console.log(`✔ Created ${outPath}`);
+}
+
+// Extracts a comment from the content of a file based on a label.
+export function extractSingleLineComment(content: string, label: string): string | undefined {
+  // https://regex101.com/r/oLFaHP/1
+  const match = content.match(new RegExp(`//\\s+${label}:\\s+(.*)`));
+  return match?.[1].trim();
```
samayer12, line 119: Is the index `1` because `0` is something outside the capture group? At first glance this return statement felt like "return the second match" which made me curious.

src/cli/crd/generate.ts
```diff
@@ -0,0 +1,316 @@
...
+}
+
+export function extractDetails(sourceFile: SourceFile): {
```
samayer12, line 122: Should we call this a `CRDdetails` type?

src/cli/crd/generate.ts
```diff
@@ -0,0 +1,316 @@
...
+  const scope = getStr("scope");
+  if (scope !== "Cluster" && scope !== "Namespaced") {
+    throw new Error(`'scope' must be either "Cluster" or "Namespaced", got "${scope}"`);
+  }
+
+  return {
+    plural: getStr("plural"),
+    scope,
+    shortName: getStr("shortName"),
+  };
```
samayer12:
```suggestion
  if (scope === "Cluster" || scope === "Namespaced") {
    return {
      plural: getStr("plural"),
      scope,
      shortName: getStr("shortName"),
    };
  }

  throw new Error(`'scope' must be either "Cluster" or "Namespaced", got "${scope}"`);
```

This proposal rewrites the conditional in the positive case instead of the negative case.

samayer12 (changes requested):
Left some comments on my first pass, ran out of steam as I got into `generate.ts`. Will finish that review later.

Overall, things look good! My comments generally center around testing & organization versus functionality.

src/cli.ts
cmwylie19: we have to have this since CRD is a command that branches to commands. It is the first command we have had that has children commands

src/cli/crd/generate.ts
cmwylie19: it is for extract certain details, like `kind`, `group`, `plurals`

The comments look like this in the generated file:

```ts
// Kind: Memcache
// Kind: VirtualService
// Kind: Gateway
// Kind: PeprStore
```

src/cli/crd/generate.ts
cmwylie19: i personally don't care if the error or the return gets returned first but i guess so!

src/cli/crd/generate.ts
cmwylie19: In this case it is commander that calls this function through an action. I don't really know if this is going to make a difference? idk though, given that do you still feel that way?

```ts
export default new Command("generate")
  .description("Generate CRD manifests from TypeScript definitions")
  .option("--output <output>", "Output directory for generated CRDs", "./crds")
  .action(generateCRDs);
 ```

Comment (samayer12) [reply src/cli/crd/generate.ts:49]: Yeah, this is fine for now. I think I ought to consider the pros/cons of `Promise<void>` in the context of the whole project instead of this within this one PR. We can leave it as-is for now.

File: src/cli/crd/generate.ts (hunk 18 above)
Comment (cmwylie19) [reply src/cli/crd/generate.ts:122]: we have something really similar in the file called `extractCRDDetails`. This extract details is used specifically for getting the details section that is generated during `npx pepr crd create`. I think it aligns very well to what is does, which is just reading those details. Feel free to push back and we can have a convo.

File: src/cli/crd/create.test.ts (hunk 2 above)
Comment (cmwylie19) [reply src/cli/crd/create.test.ts:29]: I see where you are thinking, this exact suggestion would not work due to those comments which do not actually exist in the string that the function uses to scaffol (would fail on every) for things like `// CRD config fields`.

This is a good idea so here is the issue for it so we can tackle it in the future since we have limited resources this week.

https://github.com/defenseunicorns/pepr/issues/2053

File: src/cli/crd/create.test.ts (hunk 3 above)
Comment (cmwylie19) [reply src/cli/crd/create.test.ts:17]: Yes, we could do that. Here is an issue for it. Due to resource constraints this week. 

https://github.com/defenseunicorns/pepr/issues/2053

Comment (samayer12) [reply src/cli/crd/create.test.ts:17]: Sounds good, happy to take that on as follow-on work

**Review by samayer12 — DISMISSED** [review]
Changes since last review look good. #2053 captures some follow-on work

File: src/cli.ts [hunk 20]
```diff
@@ -47,4 +49,5 @@ format(program);
 monitor(program);
 uuid(program);
 kfc(program);
+crd(program);
```
samayer12, line 52: We considered printing a brief message that this feature is in `alpha` to users. Where would be the best place to log it? Here or within `cli/crd/index.ts`?

cmwylie19: yes sounds good

samayer12 (dismissed):

src/cli/crd/generate.ts
cmwylie19: im not sure, copied it from another section of the code and was looking like that. definitely strange

src/cli/crd/generate.test.ts
samayer12: see #2064

src/cli/crd/generate.test.ts
samayer12: see #2065

src/cli/crd/generate.test.ts
samayer12: see #2065

src/cli/crd/generate.test.ts
samayer12: see #2066

src/cli/crd/generate.test.ts
samayer12: see #2066

src/cli/crd/generate.test.ts
samayer12: see #2067

src/cli/crd/generate.ts
samayer12: see #2068

samayer12 (approved):

PR #2538, opened 2025-08-12
chore: remove wasm tutorial and clean up tables for existing docs

### LINKED ISSUE #2526: Remove documentation around WASM

- Update docs
- Remove CLI Commands

**Background** - In the early days of Pepr there was an idea to use a Pepr module to replace or compliment [Zarf's Admission Controller](https://github.com/zarf-dev/zarf/issues/1632). 

The idea is that you get more expressiveness and flexibility through the Admission Process with Pepr and the controller would be the same size or less. The problem however, is that Zarf's admission controller uses  the Docker's [distribution](https://github.com/zarf-dev/zarf/blob/8067c8b31deea98610917cdcd918ede5ac0d20af/src/pkg/transform/image.go#L12) Go library to parse images. Since that library is battle tested and had been working out great for them we did not want to implement from scratch in TypeScript.

The idea was to use WASM to call Docker's distribution library (or call to WASM Functions that are implemented in Zarf Go libaries) from TypeScript and Unmarshal the strings into Go objects.

This is still possible today, and Pepr has a CLI command that adds `includedFiles` into the Docker build and the user pushes the file to their own registry.

**Problem**
- There are currently not users using this feature
- There is a lot of prep work required since Pepr runs on the server side and the WASM libraries are [DOM libraries](https://docs.pepr.dev/main/user-guide/webassembly/)
- We do not have a great way to test this because we do not include Go code in our Pepr Core repo

Ultimately, if someone wants to do this workflow, they still can because included files is still there.

cmwylie19:
## Description

Remove wasm tutorial and clean up `README.md` in the current directories.

**Note** - [on-demand workflows do not run on changes](https://github.com/defenseunicorns/pepr/blob/41849f6c01b95350461679f8dc845dd449eda3f0/.github/workflows/node.js.yml#L7) to `*.md` files so it is expected that they do not report.

## Related Issue

Fixes #2526 

Relates to #

## Type of change

## Checklist before merging


samayer12 (approved):
Sync discussion around removing WASM docs since it's an unused feature, to the best of our knowledge.

PR #2590, opened 2025-09-02
chore: force user to use an integer for webhook's timeoutSeconds

### LINKED ISSUE #2514: `pepr build --timeout` accepts unexpected float value
### Environment
Device and OS: N/A
App version: 0.52.3
Kubernetes distro being used: N/A
Other: N/A

### Steps to reproduce
1. `npx -y pepr@0.52.3 build --timeout 5.0`

### Expected result

`Error: Value must be an integer.`

### Actual Result

Module builds

```
❯ npx -y pepr@0.52.3 build --timeout 5.0

[...]
[16:13:25.397] INFO (98321): Module asdf has capability: hello-pepr
```

### Severity/Priority

Low

### Additional Context

The pepr docs state:

```
-t, --timeout <seconds> - How long the API server should wait for a webhook to respond before treating the call as a failure.
```

We should force input to be an, even though `5.0` can be parsed as `5` since 5.0 is an invalid timeout value in kubernetes

cmwylie19:
## Description

Currently the user can use a float for the webhook timeout, in Kubernetes it must be an integer so we are going to force the user to use an integer.

```bash
> k explain MutatingWebhookConfiguration.webhooks.timeoutSeconds
GROUP:      admissionregistration.k8s.io
KIND:       MutatingWebhookConfiguration
VERSION:    v1

FIELD: timeoutSeconds <integer>


DESCRIPTION:
    TimeoutSeconds specifies the timeout for this webhook. After the timeout
    passes, the webhook call will be ignored or the API call will fail based on
    the failure policy. The timeout value must be between 1 and 30 seconds.
    Default to 10 seconds.
```

## Related Issue

Fixes #2514 

Relates to #

## Type of change

## Checklist before merging

### DIFF
--- a/src/cli/build/index.test.ts
+++ b/src/cli/build/index.test.ts
@@ -336,7 +336,11 @@ describe("build CLI command", () => {
         error: "Number must be between 1 and 30.",
         description: "reject values above maximum",
       },
-      { value: "not-a-number", error: "Not a number.", description: "reject non-numeric values" },
+      {
+        value: "not-a-number",
+        error: "Value must be an integer.",
+        description: "reject non-numeric values",
+      },
       { value: "5.2", error: "Value must be an integer.", description: "reject float values" },
     ])("should $description: $value", async ({ value, error }) => {
       await runProgramWithError([timeoutFlag, value], error);
--- a/src/cli/build/index.test.ts
+++ b/src/cli/build/index.test.ts
@@ -341,6 +341,7 @@ describe("build CLI command", () => {
         error: "Value must be an integer.",
         description: "reject non-numeric values",
       },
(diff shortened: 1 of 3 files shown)


samayer12 (changes requested):
Check out the commented out test case on line 348-350 of this file for an example of a failing test case that rejects `float` input.

The issue reported in #2514 doesn't match the expected result when I run `npx -y pepr@file://../pepr-0.0.0-development.tgz build --timeout 5.0` against a candidate build off this branch.

This PR still accepts `5.0` as input, and then converts it to an integer. That's fine for the K8s resource... but should we let users think that `--timeout` can be a float?

samayer12 (approved):
Sync review

PR #2707, opened 2025-10-20
fix: new release workflow

cmwylie19:
## Description

Since we cannot call a workflow from a workflow, and we cannot use a PAT to allow to to trigger a workflow, we need to pivot a run all critical jobs in CI before the release so if and only if they pass then the release will occur.

## Related Issue

Fixes #

Relates to #2714 

## Type of change

## Checklist before merging


.github/workflows/commitlint.yml
```diff
@@ -8,6 +8,7 @@ on:
     branches: [main]
     types: [opened, edited, synchronize]
 
+
```
cmwylie19:
```suggestion
```

.github/workflows/dependency-review.yml
```diff
@@ -7,12 +7,13 @@
...
 name: 'Dependency Review'
-on: 
+on:
   pull_request:
     paths-ignore:
     - "**.md"
     - "LICENSE"
-    - "CODEOWNERS" 
+    - "CODEOWNERS"  
+
```
cmwylie19:
```suggestion
```

.github/workflows/dependency-review.yml
```diff
@@ -7,12 +7,12 @@
...
 # Source repository: https://github.com/actions/dependency-review-action
 name: 'Dependency Review'
-on: 
+on:
   pull_request:
     paths-ignore:
     - "**.md"
     - "LICENSE"
-    - "CODEOWNERS" 
+    - "CODEOWNERS"
```
cmwylie19:
```suggestion
    - "CODEOWNERS"
```

samayer12 (dismissed):
Sync discussion

samayer12 (approved):

PR #2712, opened 2025-10-21
chore: eslint warning cleanup

ericwyles:
## Description

Refactors several methods to reduce code complexity scores reported by eslint. This should get the project to a clean state and eslint warnings can be treated as errors.

End to End Test:    

## Type of change

## Checklist before merging


src/lib/assets/webhooks.test.ts
```diff
@@ -376,6 +402,35 @@ describe("configureAdditionalWebhooks", () => {
...
+    };
+    const additionalWebhooks = [
+      { failurePolicy: "Fail", namespace: "additional-namespace-1" },
+      { failurePolicy: "Ignore", namespace: "additional-namespace-2" },
+    ];
+    const result = configureAdditionalWebhooks(webhookConfig, additionalWebhooks);
+    expect(result).toEqual(webhookConfig);
+  });
+
+  it("should ignore additionalWebHooks if the original webhookConfig contains an empty array of webhooks", () => {
```
ericwyles: Flagging this one for attention. I added this test chasing coverage and reverse engineered the implementation to figure out the behavior. The behavior didn't intuitively make sense to me (basically additionalWebHooks get discarded in this case), but I verified this test also passes against `main`

ericwyles: @cmwylie19 / @samayer12 -- Just thought this one looked weird and wanted your sanity check!

src/lib/assets/yaml/overridesFile.ts
```diff
@@ -14,6 +14,33 @@ export type ChartOverrides = {
...
+  runAsNonRoot: true;
+  fsGroup: number;
+};
+export type ContainerSecurityContext = {
+  runAsUser: number;
+  runAsGroup: number;
+  runAsNonRoot: true;
+  allowPrivilegeEscalation: false;
+  capabilities: { drop: ["ALL"] };
+};
```
samayer12:
```suggestion
type Probes = {
  readinessProbe: {
    httpGet: { path: string; port: number; scheme: "HTTPS" };
    initialDelaySeconds: number;
  };
  livenessProbe: {
    httpGet: { path: string; port: number; scheme: "HTTPS" };
    initialDelaySeconds: number;
  };
};
type Resources = {
  requests: { memory: string; cpu: string };
  limits: { memory: string; cpu: string };
};
type PodSecurityContext = {
  runAsUser: number;
  runAsGroup: number;
  runAsNonRoot: true;
  fsGroup: number;
};
type ContainerSecurityContext = {
  runAsUser: number;
  runAsGroup: number;
  runAsNonRoot: true;
  allowPrivilegeEscalation: false;
  capabilities: { drop: ["ALL"] };
};
```

I'm not certain these types need to be exported. If they do, this project has a `types.ts` file in `src/lib/` that traditionally holds exported types (not that that's necessarily the best way to do it?)

src/lib/assets/yaml/overridesFile.ts
```diff
@@ -206,3 +138,75 @@ export async function writeSchemaYamlFromObject(
...
+
+export function namespaceBlock(config: ModuleConfig): {
+  annotations: Record<string, string>;
+  labels: Record<string, string>;
+} {
+  return {
+    annotations: {},
+    labels: config.customLabels?.namespace ?? { "pepr.dev": "" },
+  };
+}
```
samayer12:
```suggestion
function runIdsForImage(image: string): { uid: number; gid: number; fsGroup: number } {
  const id = image.includes("private") ? 1000 : 65532;
  return { uid: id, gid: id, fsGroup: id };
}

function commonProbes(): Probes {
  return {
    readinessProbe: {
      httpGet: { path: "/healthz", port: 3000, scheme: "HTTPS" },
      initialDelaySeconds: 10,
    },
    livenessProbe: {
      httpGet: { path: "/healthz", port: 3000, scheme: "HTTPS" },
      initialDelaySeconds: 10,
    },
  };
}

function commonResources(): Resources {
  return {
    requests: { memory: "256Mi", cpu: "200m" },
    limits: { memory: "512Mi", cpu: "500m" },
  };
}

function podSecurityContext(image: string): PodSecurityContext {
  const ids = runIdsForImage(image);
  return {
    runAsUser: ids.uid,
    runAsGroup: ids.gid,
    runAsNonRoot: true,
    fsGroup: ids.fsGroup,
  };
}

function containerSecurityContext(image: string): ContainerSecurityContext {
  const ids = runIdsForImage(image);
  return {
    runAsUser: ids.uid,
    runAsGroup: ids.gid,
    runAsNonRoot: true,
    allowPrivilegeEscalation: false,
    capabilities: { drop: ["ALL"] },
  };
}

function controllerLabels(
  name: string,
  uuid: string,
  kind: "admission" | "watcher",
): Record<string, string> {
  return {
    app: kind === "admission" ? name : `${name}-watcher`,
    "pepr.dev/controller": kind,
    "pepr.dev/uuid": uuid,
  };
}

function controllerAnnotations(description?: string): Record<string, string> {
  return { "pepr.dev/description": `${description ?? ""}` };
}

function namespaceBlock(config: ModuleConfig): {
  annotations: Record<string, string>;
  labels: Record<string, string>;
} {
  return {
    annotations: {},
    labels: config.customLabels?.namespace ?? { "pepr.dev": "" },
  };
}
```

Similarly, it looks like these helpers are only used internally. So, probably no need to export.

src/lib/assets/yaml/overridesFile.ts
```diff
@@ -183,8 +113,10 @@ export async function overridesFile(
       },
     },
   };
-  /** write values.schema.yaml */
+
+  // write values.schema.yaml
```
samayer12:
```suggestion
```
Function name is clear enough, in my opinion

src/lib/assets/webhooks.ts
```diff
@@ -64,82 +64,19 @@ export async function webhookConfigGenerator(
...
-      namespace: "pepr-system",
-      path: fullApiPath,
-    };
-  }
-
-  const isMutate = mutateOrValidate === WebhookType.MUTATE;
-  const rules = await generateWebhookRules(assets, isMutate);
-
-  // If there are no rules, return null
+  const rules = await generateWebhookRules(assets, mutateOrValidate === WebhookType.MUTATE);
```
samayer12: This is probably the original implementation showing here, but the fact that we need to tell this function what type of webhook it is as a boolean input feels odd. Can that conditional check be moved into the function body instead of determined when the function is called?

src/lib/controller/index.test.ts
```diff
@@ -58,3 +61,64 @@ describe("Controller", () => {
     expect(mockHooks.onReady).toHaveBeenCalled();
   });
 });
+
+describe("getContextAndLog", () => {
```
samayer12: This batch of 5 test cases may be a good opportunity to use a table-driven testing pattern with `it.each()`. I think the "act" step in all of them is the same and we're just adjusting inputs to cover several different cases.

This change could be worth it to achieve a little more brevity while providing the same level of confidence.

src/lib/controller/index.ts
```diff
@@ -322,3 +308,26 @@ export class Controller {
     }
   }
 }
+
+export function getContextAndLog(
+  req: express.Request,
+  admissionKind: "Mutate" | "Validate",
```
samayer12: I think we have a defined type enum that represents this union type of two strings? Maybe we can use that here instead? Check one of the `types.ts` files.

src/lib/core/module.test.ts
```diff
@@ -56,6 +56,23 @@ describe("PeprModule", () => {
...
+    });
+
+    it("should use pepr.alwaysIgnore.namespaces when it contains values", () => {
+      const pkg = clone(mockPackageJSON);
+      pkg.pepr.alwaysIgnore = { namespaces: ["ns1", "ns2"] };
+      Reflect.deleteProperty(pkg.pepr, "watch");
+
+      new PeprModule(pkg);
+      expect(startServerMock).toHaveBeenCalledWith(3000);
+    });
```
samayer12, line 85: It looks like the "assert" portion of these two tests are the same. Is there any observable difference we can detect at this level aside from "startServerMock was called with 3000"? This may also be something appropriate for a higher level of test (or not worth testing directly?).

src/lib/processors/watch-processor.ts
```diff
@@ -89,7 +89,7 @@ const eventToPhaseMap = {
 export function setupWatch(capabilities: Capability[], ignoredNamespaces?: string[]): void {
   for (const capability of capabilities) {
     for (const binding of capability.bindings.filter(b => b.isWatch)) {
-      runBinding(binding, capability.namespaces, ignoredNamespaces);
+      void runBinding(binding, capability.namespaces, ignoredNamespaces);
```
samayer12, line 92: `runBinding` is `async` and it looks like this function may not be treating it as an async function? This may be an implementation detail for the maintainers to handle. If it gets too complex to look at, toss an issue into the backlog.

src/lib/processors/validate-processor.test.ts
samayer12, line 1: See general notes on type-coercion in tests.

src/lib/processors/mutate-processor.test.ts
samayer12, line 1: See general notes on type-coercion.

src/lib/assets/webhooks.test.ts
samayer12: Based off the usage of this function (exported just so that it can be tested), these tests appear to go after implementation-details. I think it would be more appropriate to test this behavior through `webhookConfigGenerator()` since that contains this function and is called by other parts of the project.

src/lib/processors/validate-processor.ts
```diff
@@ -73,45 +73,81 @@ export async function validateProcessor(
...
+  req: AdmissionRequest,
+  namespaces: string[] | undefined,
+  ignoreNamespaces: string[],
+): boolean {
+  const reason = shouldSkipRequest(binding, req, namespaces ?? [], ignoreNamespaces);
+  if (reason !== "") {
+    Log.debug(reason);
+    return true;
+  }
+  return false;
```
samayer12: Be sure to validate that these functions must be exported. Exporting too many functions can lead to a tangled web of dependencies and send us back to fighting circular dependencies.

You can check for circulars with:
```
npx madge --circular --ts-config config/tsconfig.root.json --extensions ts,js src/
```

And get a cool(?) chart with:
```
npx madge --ts-config config/tsconfig.root.json --extensions ts,js src/ -i out.png
```

src/lib/assets/ignoredNamespaces.test.ts
```diff
@@ -44,3 +54,56 @@ describe("when namespaces are set in environment variables", () => {
     });
   });
 });
+
+describe("getIgnoreNamespaces", () => {
```
samayer12: This looks like a good candidate for a table-driven test. It could also be tested indirectly by test cases within `mutate-processor.ts` and `validate-processor.ts`. 

Running `vitest` with code coverage turned on but only checking a subset of the test files (e.g., just the files that call this function) can indicate indirect code coverage of this block of code. It's a little tedious, but is a mechanism I've used to avoid writing _too many_ tests that make future changes very brittle.

src/lib/assets/yaml/overridesFile.test.ts
```diff
@@ -269,4 +280,178 @@ describe("overridesFile", () => {
...
+    const calls = (fs.writeFile as Mock).mock.calls;
+    const yamlCall = calls.find(([path]) => path === mockPath);
+    const [, writtenContent] = yamlCall!;
+    const parsedYaml = loadYaml(writtenContent as string) as OverridesFileSchema;
+
+    expect(parsedYaml.additionalIgnoredNamespaces).toEqual(["nsA", "nsB"]);
+  });
+});
+
+describe("overrides helpers", () => {
```
samayer12: We may not need to test these helpers directly as they could be covered by test cases that call whatever function uses these helpers. If we do need to test these helpers, it _could_ be a good opportunity for some table-driven testing since they're multiple cases focused on relatively simple blocks of code.

integration/cli/crd.test.ts
```diff
@@ -414,3 +408,59 @@ describe("crd", () => {
...
+  workdir: Workdir;
+  testModule: string;
+  group: string;
+  version: string;
+  kindName: string;
+  shortName: string;
+  plural: string;
+  scope: string;
+  domain: string;
+}): Promise<void> {
```
samayer12: Can we extract this to a `type` instead of declaring a somewhat verbose object as part of the function declaration? Something like:

```
type ArtifactParamsOrSomeAppropriateName = {
  workdir: Workdir;
  testModule: string;
  group: string;
  version: string;
  kindName: string;
  shortName: string;
  plural: string;
  scope: string;
  domain: string;
}

[...]

generateCRDArtifacts(artifactsOrSomeName: ArtifactParamsOrSomeAppropriateName): Promise<void>
```

integration/cli/build.configequality.test.ts
```diff
@@ -22,91 +30,25 @@ describe("build", () => {
...
-      await pepr.cli(testModule, { cmd: `npm install` });
+      await setupTestModule(testModule, id, workdir.path());
     }, time.toMs("2m"));
 
     describe("maps config options from package.json into pepr + helm manifests", () => {
       let moduleConfig: Required<ModuleConfig>;
-      let peprResources: any; // eslint-disable-line @typescript-eslint/no-explicit-any
-      let helmResources: any; // eslint-disable-line @typescript-eslint/no-explicit-any
+      let peprResources: KubernetesObject[];
+      let helmResources: KubernetesObject[];
```
samayer12, line 44: Nice! Thank you!!

integration/cli/build.configequality.test.ts
```diff
@@ -258,9 +200,8 @@ describe("build", () => {
...
           expect(deployWatcher.spec!.template!.spec!.serviceAccountName).toBe(peprUuid);
           expect(
-            deployWatcher
-              .spec!.template!.spec!.volumes!.filter(vol => vol.name === "tls-certs")
-              .at(0)!.secret!.secretName,
+            deployWatcher.spec!.template!.spec!.volumes!.find(vol => vol.name === "tls-certs")!
+              .secret!.secretName,
           ).toBe(`${peprUuid}-tls`);
           expect(
             deployWatcher.spec!.template!.spec!.volumes!.filter(vol => vol.name === "module").at(0)!
```
samayer12: Complexity probably triggers here because we have a nested test checking several assertions against a collection. It'll be a common-theme for test organization throughout this PR, but perhaps using an `it.each()` test against `[resource.select(peprResources, kind.ClusterRole, peprUuid)]` would let us avoid a suppression here.

Additionally, say that there were 10 objects in the collection and only one of them was broken. With a table-driven approach the test suite will check all 10 and report 9 ok and 1 bad. With this approach, the test fails as soon as any 1 of the 10 are bad. I believe the former approach is preferable since it shows a maintainer that "9/10 things are fine" rather than saying "at least 1 thing is broken."

samayer12 (changes requested):
Great PR and I appreciate the extra work to add tests in addition to simply resolving the linter warnings! This PR definitely achieves the Dash Days goal of resolving the remaining linter warnings, thank you! In my opinion, even if this PR is still open by the end of Dash Days you should **still be eligible for the prize** since we'll merge this PR once feedback is resolved.

Here's general feedback I'd like to address before we officially merge it in:
* Let's double-check that exported functions actually must be exported.
* I noticed 25 instances of type-coercion (`as unknown as`) in the tests. Some of these may be pre-existing patterns. Usually, we want to make sure that the type-coercion is truly necessary before bringing it in on a PR. Historically, type-coercion has been added as a quick kludge that becomes unintentionally permanent. A good reason to use type-coercion is when using a full type in a test requires excessive amounts of boilerplate.
* Use of table-driven tests in new test cases. For instances where `func(x)` is being tested with several different inputs and consistent mocking, it may be better to use a table-driven approach with `it.each`. There should be a few examples of that sort of testing elsewhere in the project.

src/lib/assets/yaml/overridesFile.ts
ericwyles: This should be addressed now. Thanks so much for the feedback, I simplified a lot by not exporting these and then just hitting some additional edge cases on the tests from the top level function.

src/lib/assets/yaml/overridesFile.ts
ericwyles: This should be addressed now. Thanks so much for the feedback, I simplified a lot by not exporting these and then just hitting some additional edge cases on the tests from the top level function.

src/lib/assets/yaml/overridesFile.test.ts
ericwyles: Addressed this exactly like you suggested! Was able to hit the edges with a combination of table driven tests and one additional test case, and handling it all through the original exported function now. Much better!

src/lib/assets/webhooks.ts
ericwyles: @samayer12 I'm not sure how to handle this one. This ultimately ends up in validateRule which handles it like this:

```
export const validateRule = (
  binding: Binding,
  isMutateWebhook: boolean,
): V1RuleWithOperations | undefined => {
  const { event, kind, isMutate, isValidate } = binding;

  // Skip invalid bindings based on webhook type
  if ((isMutateWebhook && !isMutate) || (!isMutateWebhook && !isValidate)) {
    return undefined;
  }
  ...

  return ruleObject;
};
```

So in the part you commented on there is a boolean `mutateOrValidate` but down in this level we have `isValidate` which is contained in the bindings in the capabilities in the asset. And we still have `isMutate` that's been passed down from above.

I'm not really sure what to do with the fact that `mutateOrValidate` could be set to WebhookType.VALIDATE but then inside the asset you could have a binding with isValidate===false (or vice versa)

Not being super familiar with the data, I don't know if this means these are two different concepts (isValidate in the binding vs WebhookType.VALIDATE) or if there is some reduncancy in the data structures. 

Should we just pass the `WebhookType` all the way down into `validateRule` and keep the same logic and just move it deeper?

samayer12: Check out #2736. That's kinda what I had in mind. Which, as you mentioned, just passes `WebhookType` down a little bit. I believe it's more appropriate for the `generateWebhookRules` function to determine webhook type than the `webhookConfigGenerator`. However, making that change caused some test failures that I haven't looked into in detail.

This is also unrelated to linter cleanup, just a target of opportunity I saw during the review. We can shift this to a new item in the backlog if necessary.

ericwyles: I think those test failures might be a coincidence and something we broke in flight. I understand what you are suggesting now and will take a look once we get other stuff back to green

src/lib/controller/index.ts
ericwyles: The more I worked with this, the less I liked getContextAndLog. I took a fresh look this afternoon and reworked index.ts again and it's all much simpler.

src/lib/controller/index.test.ts
ericwyles: Ended up getting rid of these tests completely and simplified the changes to index.test.ts

src/lib/assets/webhooks.ts
ericwyles: I addressed this one!

integration/cli/build.configequality.test.ts
ericwyles: @josiblair took care of this one!

integration/cli/crd.test.ts
josiblair: extracted that into a type! thanks :)

src/lib/core/module.test.ts
josiblair: Good catch! You're right that both tests had identical assertions. I've updated them to actually verify the different configuration paths are working correctly. Lmk if you see anything else that needs updating!

src/lib/processors/validate-processor.test.ts
josiblair: I eliminated 3 of them (the `Binding` and `ModuleConfig` ones) using helpers. The remaining ones I feel are justifiably coerced—they're intentionally testing edge cases where incomplete Capability objects need to bypass the type system to verify error handling. Creating properly-typed objects for those scenarios would require excessive boilerplate, so the type-coercion serves a legitimate purpose. Lmk if you feel otherwise!

src/lib/assets/ignoredNamespaces.test.ts
ericwyles: I was able to get rid of these, everything is being handled indirectly!

src/lib/processors/watch-processor.ts
ericwyles: Hey @samayer12 yeah that's true. I'm not really sure the intent here, I was thinking the caller here is trying to fire-and-forget but I'm not sure. I can open an issue but how would you like to leave this in the PR for now? A few options:

1. We could leave it as-is and let eslint keep warning about it
2. We could leave it as I have it with `void` to ignore
3. I could add something here to just catch errors from the promise and log them. I was worried about doing this because we're in a nested loop here and I didn't want to introduce something blocking if it should truly be fire and forget.

Thanks!

src/lib/assets/webhooks.test.ts
ericwyles: Overhauled this a lot, should be good for another look!

src/lib/processors/validate-processor.test.ts
ericwyles: I ended up reworking this a bit more today. Please take another look, it should be quite a bit simpler and less changes now.

src/lib/processors/mutate-processor.test.ts
ericwyles: I ended up reworking this a bit more today. Please take another look, it should be quite a bit simpler and less changes now.

src/lib/processors/validate-processor.ts
ericwyles: I simplified the changes in this file a lot, please take another look when you can!

integration/cli/build.configequality.test.ts
```diff
@@ -76,95 +74,104 @@ describe("build", () => {
...
+          { kind: kind.ValidatingWebhookConfiguration, name: (uuid: string): string => uuid },
+        ];
+
+        it.each(
+          sources.flatMap(src =>
+            resourceKinds.map(r => ({
+              source: src.source,
+              getResources: src.getResources,
+              resourceKind: r.kind,
+              name: r.name,
```
samayer12:
```suggestion
            resourceKinds.map(resourceKind => ({
              source: src.source,
              getResources: src.getResources,
              resourceKind: resourceKind.kind,
              name: resourceKind.name,
```

Not enforced by the linter, but I have a general preference against single-character variable names.

integration/cli/build.configequality.test.ts
```diff
@@ -76,95 +74,104 @@ describe("build", () => {
...
+          "$resourceKind in $source resources should be defined",
+          ({ getResources, resourceKind, name }) => {
+            const res = resource.select(getResources(), resourceKind, name(peprUuid));
             expect(res).toBeDefined();
-          }
-        });
+          },
+        );
 
+        // --- ClusterRoleBinding tests ---
```
samayer12:
```suggestion
```

This is already scoped within a `describe` block for `uuid`. The test name makes the purpose of the test clear.

integration/cli/build.configequality.test.ts
```diff
@@ -76,95 +74,104 @@ describe("build", () => {
...
-            expect(clusterRoleBinding.roleRef.name).toBe(peprUuid);
-            expect(clusterRoleBinding.subjects?.at(0)?.name).toBe(peprUuid);
+            const crb = resource.select(getResources(), kind.ClusterRoleBinding, peprUuid);
+            expect(crb).toBeDefined();
+            expect(crb.roleRef.name).toBe(peprUuid);
+            expect(crb.subjects?.[0]?.name).toBe(peprUuid);
           },
         );
 
+        // --- RoleBinding tests ---
```
samayer12:
```suggestion
```

This is already scoped within a `describe` block for `uuid`. The test name makes the purpose of the test clear.

integration/cli/build.configequality.test.ts
```diff
@@ -76,95 +74,104 @@ describe("build", () => {
...
-            expect(roleBindingStore.roleRef.name).toBe(`${peprUuid}-store`);
-            expect(roleBindingStore.subjects?.at(0)?.name).toBe(`${peprUuid}-store`);
+            const rb = resource.select(getResources(), kind.RoleBinding, `${peprUuid}-store`);
+            expect(rb).toBeDefined();
+            expect(rb.roleRef.name).toBe(`${peprUuid}-store`);
+            expect(rb.subjects?.[0]?.name).toBe(`${peprUuid}-store`);
           },
         );
 
+        // --- Deployment (admission) ---
```
samayer12:
```suggestion
```

This is already scoped within a `describe` block for `uuid`. The test name makes the purpose of the test clear.

integration/cli/build.configequality.test.ts
```diff
@@ -76,95 +74,104 @@ describe("build", () => {
...
-            ).toBe(`${peprUuid}-tls`);
+            const d = resource.select(getResources(), kind.Deployment, peprUuid);
+            expect(d).toBeDefined();
+            expect(d.metadata?.labels?.app).toBe(peprUuid);
+            expect(d.metadata?.labels?.["pepr.dev/uuid"]).toBe(moduleConfig.uuid);
+            expect(d.spec?.template?.spec?.serviceAccountName).toBe(peprUuid);
           },
         );
 
+        // --- Deployment (watcher) ---
```
samayer12:
```suggestion
```

This is already scoped within a `describe` block for `uuid`. The test name makes the purpose of the test clear.

integration/cli/build.configequality.test.ts
```diff
@@ -76,95 +74,104 @@ describe("build", () => {
...
-            expect(deploy.spec?.template?.spec?.serviceAccountName).toBe(peprUuid);
+            const d = resource.select(getResources(), kind.Deployment, `${peprUuid}-watcher`);
+            expect(d).toBeDefined();
+            expect(d.metadata?.labels?.app).toBe(`${peprUuid}-watcher`);
+            expect(d.metadata?.labels?.["pepr.dev/uuid"]).toBe(moduleConfig.uuid);
+            expect(d.spec?.template?.spec?.serviceAccountName).toBe(peprUuid);
           },
         );
 
+        // --- Services ---
```
samayer12:
```suggestion
```

This is already scoped within a `describe` block for `uuid`. The test name makes the purpose of the test clear.

integration/cli/build.configequality.test.ts
```diff
@@ -76,95 +74,104 @@ describe("build", () => {
...
-            const svc = resource.select(getResources(), kind.Service, `${peprUuid}-watcher`);
-            expect(svc).toBeDefined();
-            expect(svc.spec?.selector?.app).toBe(`${peprUuid}-watcher`);
+            const s = resource.select(getResources(), kind.Service, `${peprUuid}-watcher`);
+            expect(s).toBeDefined();
+            expect(s.spec?.selector?.app).toBe(`${peprUuid}-watcher`);
           },
         );
 
+        // --- Webhooks ---
```
samayer12:
```suggestion
```

This is already scoped within a `describe` block for `uuid`. The test name makes the purpose of the test clear.

integration/cli/build.configequality.test.ts
```diff
@@ -193,12 +200,13 @@ describe("build", () => {
           },
         );
 
+        // --- Store SA ---
```
samayer12:
```suggestion
```

This is already scoped within a `describe` block for `uuid`. The test name makes the purpose of the test clear.

integration/cli/build.configequality.test.ts
```diff
@@ -76,95 +74,104 @@ describe("build", () => {
...
-            expect(deploy.spec?.template?.spec?.serviceAccountName).toBe(peprUuid);
-            expect(
-              deploy.spec?.template?.spec?.volumes?.find(v => v.name === "tls-certs")?.secret
-                ?.secretName,
-            ).toBe(`${peprUuid}-tls`);
+            const d = resource.select(getResources(), kind.Deployment, peprUuid);
+            expect(d).toBeDefined();
+            expect(d.metadata?.labels?.app).toBe(peprUuid);
+            expect(d.metadata?.labels?.["pepr.dev/uuid"]).toBe(moduleConfig.uuid);
+            expect(d.spec?.template?.spec?.serviceAccountName).toBe(peprUuid);
```
samayer12:
```suggestion
            const deployment = resource.select(getResources(), kind.Deployment, peprUuid);
            expect(deployment).toBeDefined();
            expect(deployment.metadata?.labels?.app).toBe(peprUuid);
            expect(deployment.metadata?.labels?.["pepr.dev/uuid"]).toBe(moduleConfig.uuid);
            expect(deployment.spec?.template?.spec?.serviceAccountName).toBe(peprUuid);
```

Preference for variable names that are not one character long

integration/cli/build.configequality.test.ts
```diff
@@ -76,95 +74,104 @@ describe("build", () => {
...
-            const deploy = resource.select(getResources(), kind.Deployment, `${peprUuid}-watcher`);
-            expect(deploy).toBeDefined();
-            expect(deploy.metadata?.labels?.app).toBe(`${peprUuid}-watcher`);
-            expect(deploy.metadata?.labels?.["pepr.dev/uuid"]).toBe(moduleConfig.uuid);
-            expect(deploy.spec?.template?.spec?.serviceAccountName).toBe(peprUuid);
+            const d = resource.select(getResources(), kind.Deployment, `${peprUuid}-watcher`);
+            expect(d).toBeDefined();
+            expect(d.metadata?.labels?.app).toBe(`${peprUuid}-watcher`);
+            expect(d.metadata?.labels?.["pepr.dev/uuid"]).toBe(moduleConfig.uuid);
+            expect(d.spec?.template?.spec?.serviceAccountName).toBe(peprUuid);
```
samayer12:
```suggestion
            const deployment = resource.select(getResources(), kind.Deployment, `${peprUuid}-watcher`);
            expect(deployment).toBeDefined();
            expect(deployment.metadata?.labels?.app).toBe(`${peprUuid}-watcher`);
            expect(deployment.metadata?.labels?.["pepr.dev/uuid"]).toBe(moduleConfig.uuid);
            expect(deployment.spec?.template?.spec?.serviceAccountName).toBe(peprUuid);
```

Preference for variable names that are not one character long

integration/cli/build.configequality.test.ts
```diff
@@ -76,95 +74,104 @@ describe("build", () => {
...
 
         it.each(sources)(
           "Service (watcher) in $source resources should have correct selector",
           ({ getResources }) => {
-            const svc = resource.select(getResources(), kind.Service, `${peprUuid}-watcher`);
-            expect(svc).toBeDefined();
-            expect(svc.spec?.selector?.app).toBe(`${peprUuid}-watcher`);
+            const s = resource.select(getResources(), kind.Service, `${peprUuid}-watcher`);
+            expect(s).toBeDefined();
+            expect(s.spec?.selector?.app).toBe(`${peprUuid}-watcher`);
```
samayer12:
```suggestion
            const svc = resource.select(getResources(), kind.Service, peprUuid);
            expect(svc).toBeDefined();
            expect(svc.spec?.selector?.app).toBe(peprUuid);
},
);

it.each(sources)(
"Service (watcher) in $source resources should have correct selector",
({ getResources }) => {
            const svc = resource.select(getResources(), kind.Service, `${peprUuid}-watcher`);
            expect(svc).toBeDefined();
            expect(svc.spec?.selector?.app).toBe(`${peprUuid}-watcher`);
```

`s` to `svc`

integration/cli/build.configequality.test.ts
```diff
@@ -303,16 +311,16 @@ describe("build", () => {
...
 
         it.each(deployments.flatMap(deployment => sources.map(s => ({ ...s, deployment }))))(
           "$deployment in $source resources should have LOG_LEVEL env var",
-          ({ getResources, deployment }) => {
-            const d = resource.select(getResources(), kind.Deployment, deployment);
+          ({ getResources }) => {
+            const d = resource.select(getResources(), kind.Deployment, peprUuid);
             expect(d.spec?.template?.spec?.containers?.[0]?.env).toEqual(
               expect.arrayContaining([
                 expect.objectContaining({
```
samayer12:
```suggestion
            const deployment = resource.select(getResources(), kind.Deployment, peprUuid);
expect(deployment.spec?.template?.spec?.containers?.[0]?.env).toEqual(
expect.arrayContaining([
expect.objectContaining({
```

integration/cli/build.configequality.test.ts
```diff
@@ -334,8 +342,8 @@ describe("build", () => {
 
         it.each(deployments.flatMap(deployment => sources.map(s => ({ ...s, deployment }))))(
           "$deployment in $source resources should have TEST env var",
-          ({ getResources, deployment }) => {
-            const d = resource.select(getResources(), kind.Deployment, deployment);
+          ({ getResources }) => {
+            const d = resource.select(getResources(), kind.Deployment, `${peprUuid}-watcher`);
             expect(d.spec?.template?.spec?.containers?.[0]?.env).toEqual(
               expect.arrayContaining([
```
samayer12, line 349:
```suggestion
          ({ getResources }) => {
            const deployment = resource.select(getResources(), kind.Deployment, `${peprUuid}-watcher`);
expect(deployment.spec?.template?.spec?.containers?.[0]?.env).toEqual(
expect.arrayContaining([
```

samayer12 (dismissed):
This is a great improvement (and I think half the size of the original PR?)! Let's resolve the remaining comments and we'll be set.

integration/cli/crd.test.ts
```diff
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: 2023-Present The Pepr Authors
 
+/* eslint-disable max-statements */
```
samayer12: Are we able to remove this suppression? I could also understand if there was some excessively verbose bit of setup that had to happen.

src/lib/processors/watch-processor.ts
samayer12: For this PR, let's cross-reference to an issue in the backlog and leave it as an option to address later. In terms of implementation, I think we're okay calling it `void` for now.

src/lib/assets/webhooks.ts
```diff
@@ -104,8 +98,8 @@ export async function webhookConfigGenerator(
     };
   }
 
-  const isMutate = mutateOrValidate === WebhookType.MUTATE;
-  const rules = await generateWebhookRules(assets, isMutate);
+  //const isMutate = webhookType === WebhookType.MUTATE;
```
samayer12:
```suggestion
```

Dangling comment?

integration/cli/crd.test.ts
ericwyles: I grouped some of the tests that were just checking for specific data in the CRD into the same `describe` block and that reduce the number of statements in the arrow function. I was able to remove the suppression then. It might make the test report output a little bit nicer too by grouping those together.

src/lib/processors/watch-processor.ts
ericwyles: Created #2743 to track this.

samayer12 (approved):
Looks good! Thank you!

PR #2771, opened 2025-11-14
fix: schema post render

### LINKED ISSUE #2766: `values.schema.yaml` file has incorrect schema for nodeSelector
[upstream bug] - This is not isolated to purely nodeSelector but all `map[string]string` type objects in the `values.yaml`. With that said - the first priority is to unblock the nodeSelector

### Steps to reproduce
1. build a pepr module, `npx pepr build`
2. in the `values.yaml` add a nodeSelector for any controller
3. execute `helm template .` to ensure the manifests are generate

### Expected result

Controller is generated with the given nodeSelector in the podSpec

### Actual Result

```bash
> helm template .              
Error: values don't meet the specifications of the schema(s) in the following chart(s):
testing123:
- admission.nodeSelector: Additional property color is not allowed
```

### Severity/Priority

Medium / High, they are unable to upgrade now but they need the scheduling ability to be an easy process

### Additional Context

The reason this bug exists is due to a bug in quicktype https://github.com/defenseunicorns/kubernetes-fluent-client/issues/810

[Function in question](https://github.com/defenseunicorns/pepr/blob/dec78dbb38abebb03e21243a900cce22a65c4583/src/lib/assets/yaml/overridesFile.ts#L120)

cmwylie19:
## Description

The value.yaml are unusable due the the `values.schema.yaml` generating incorrect types. 

This fix sets the affinity type's additionalProperties to true which allows users to use `nodeSelector`.

This is an issue that came up in product suppoort

## Related Issue

Fixes #2766 

Relates to #

## Type of change

## Checklist before merging

### DIFF
--- a/src/lib/assets/yaml/overridesFile.test.ts
+++ b/src/lib/assets/yaml/overridesFile.test.ts
@@ -3,7 +3,7 @@
 
 import { expect, describe, it, vi, type Mock, beforeEach } from "vitest";
 import { promises as fs } from "fs";
-import { overridesFile, writeSchemaYamlFromObject } from "./overridesFile";
+import { overridesFile, writeSchemaYamlFromObject, fixSchemaForFlexibleMaps } from "./overridesFile";
 import type { ChartOverrides } from "./overridesFile";
 import { load as loadYaml } from "js-yaml";
 import { ModuleConfig } from "../../types";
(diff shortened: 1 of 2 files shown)


mjnagel (approved):
As discussed synchronously this will work for now, but ultimately probably want to revisit this to make a better schema (possibly quicktype isn't the right too here?). Would be good to track that in a follow-on issue.

PR #2794, opened 2025-11-26
chore: move @types/ramda to devDependencies

soltysh:
## Description
Move `@types/ramda` to `devDependencies` where it belongs, we already have proper `ramda` dependency in the main deps.

## Related Issue

None

## Type of change

## Checklist before merging


cmwylie19 (approved):

cmwylie19:
Tests that use the unicorn images will fail for this forked PR due to missing registry credentials. All other checks pass, and the PR looks good, so we’ll merge it without waiting for those required tests.

cmwylie19 (approved):
https://github.com/defenseunicorns/pepr/actions/runs/19704456259/job/56448892055?pr=2794 Just noticing that there is another failure in the UDS Smoke test.

cmwylie19 (changes requested):
https://github.com/defenseunicorns/pepr/actions/runs/19704456259/job/56448892055?pr=2794 Just noticing that there is another failure in the UDS Smoke test.

soltysh:
> https://github.com/defenseunicorns/pepr/actions/runs/19704456259/job/56448892055?pr=2794 Just noticing that there is another failure in the UDS Smoke test.

Yeah, the failures from this particular test are real and we can't ignore those. It looks like we are "leaking" ramda types in https://github.com/defenseunicorns/pepr/blob/main/src/lib/assets/webhooks.ts, primarily here:
https://github.com/defenseunicorns/pepr/blob/614fd25e9b1d6de7d60964faeb98732fdd974e54/src/lib/assets/webhooks.ts#L71 although that can be easily replaced with built-in `Array.concat`. But here https://github.com/defenseunicorns/pepr/blob/614fd25e9b1d6de7d60964faeb98732fdd974e54/src/lib/assets/webhooks.ts#L60 it's not that easy to replace both of these functions :sweat:

soltysh:
I'm going to close this based on previous comment. When we have more time we might want to return to it.

PR #2972, opened 2026-02-28
chore: prevent store cache corruption on retry by snapshotting before flush

joonas:
## Description

The retry path in `sendUpdatesAndFlushCache` used `Number(index)` to
look up operations by position in a parallel values array. Cache keys
are non-numeric strings produced by `fillStoreCache` (e.g.
`"add:/data/cap-v2-key:value"`), so `Number(key)` always evaluated to
`NaN`, and `payload[NaN]` produced `undefined`. Every retryable
failure silently replaced all pending operations with `undefined`,
permanently losing them.

The fix snapshots the cache (`{ ...cache }`) before the patch attempt
and restores it with `Object.assign(cache, snapshot)` on retryable
failure, avoiding the key-to-index cross-reference entirely.

The existing retry test was tautological — it compared `result` to
`cache`, but both were the same object reference, so the assertion
always passed regardless of corruption. The updated test snapshots the
expected state into an independent copy and uses realistic keys from
`fillStoreCache` to exercise the actual code path.

End to End Test:    

## Type of change

## Checklist before merging


cmwylie19 (approved):
yep, thanks. Nice catch

samayer12:
@greptileai

PR #2973, opened 2026-02-28
chore: eliminate array coercion in `fillStoreCache` path construction

joonas:
## Description

`Storage.clear()` collects all store keys into a single `string[]` and dispatches one `("remove", ["key1", "key2", ...])` call through to `fillStoreCache`.

The path construction placed `cacheItem.key` as a single element inside another array literal:

```
[`/data/${capabilityName}`, cacheItem.version, cacheItem.key]
```

When `.join("-")` encountered the nested array, it called `.toString()`, coercing the keys into a comma-separated string and producing one malformed JSON Patch path like
`/data/cap-v2-key1,v2-key2,v2-key3` instead of three individual remove operations. The Kubernetes API rejects or misapplies this path, so `Storage.clear()` never actually removed anything from the PeprStore resource.

Move path computation out of the shared preamble and into each branch:

- The `remove` branch now iterates over `cacheItem.key`, computing a separate path and cache entry per key.

- The `add` branch uses `cacheItem.key[0]` (a `string`) instead of `cacheItem.key` (a `string[]`), and guards with an explicit length check to prevent the same class of bug from reappearing silently.

The `migrateStore` test expectations are updated to match the corrected paths — the old trailing dashes (e.g.,
`remove:/data/hello-pepr-example-1-`) were artifacts of the array coercion producing `[""].toString()` → `""` which survived the `.filter()` as a nested array but became an empty segment after `.join("-")`.

End to End Test:    

## Type of change

## Checklist before merging


samayer12:
@greptileai

samayer12 (approved):
Test failures are due to a known issue with `thread-stream` which is a `pino` dependency. Approved.

PR #2974, opened 2026-03-01
chore: convert webhook timeout from seconds to milliseconds before comparison

joonas:
## Description

`MeasureWebhookTimeout.start()` stored `config.webhookTimeout` as-is (seconds, 1-30), but `stop()` computed elapsed time via `performance.now()` (milliseconds). The comparison `elapsedTime > this.timeout` therefore treated a 10-second timeout as 10ms.

Multiply the timeout by 1000 on entry so both sides of the comparison are in milliseconds.

Update existing tests to use realistic units (timeout in seconds, elapsed time in milliseconds) instead of abstract values that accidentally passed despite the bug. Add a boundary test for the exact-equals case.

End to End Test:    

## Type of change

## Checklist before merging


samayer12:
@greptileai

PR #2976, opened 2026-03-01
chore: check `response.result` instead of `response.warnings` for `OnError.REJECT` early-return

joonas:
## Description

The early-return condition in `mutateProcessor`'s binding loop used `response?.warnings!.length > 0` to detect whether a rejection had occurred. Since `response.warnings` accumulates entries for every failed callback regardless of the `onError` mode the pre-existing warnings from a prior iteration could trigger a spurious early-return even when the current binding succeeded and no rejection was issued. 

The `?.` followed by `!` was also contradictory: the optional chain guarded against `response` being nulli-sh while the non-null assertion claimed `warnings` was definitely defined, but `response` is always initialized on entry to `mutateProcessor`, making `?.` unnecessary.

Replace the condition with `response.result !== undefined`. `response.result` is only set inside `processRequest`'s catch block when `config.onError === OnError.REJECT`, making it the precise signal for rejection.

Add 4 tests covering the fix:

* `processRequest` does not set `response.result` when the callback succeeds, even when the response carries pre-existing warnings
* `processRequest` does not signal rejection when warnings pre-exist but the callback succeeds
* `mutateProcessor` skips remaining bindings when the first fails under REJECT mode
* `mutateProcessor` processes all bindings when all succeed under REJECT mode

End to End Test:    

## Type of change

## Checklist before merging


samayer12:
@greptileai

PR #2977, opened 2026-03-01
chore: merge RBAC verbs in `createRBACMap` instead of overwriting

joonas:
## Description

`createRBACMap` in `src/lib/helpers.ts` had two problems that caused RBAC verb loss when multiple bindings targeted the same resource kind:

1. The `isFinalize` branch unconditionally replaced the entire RBAC entry (`acc[key] = { verbs: ["patch"], ... }`), discarding any `"watch"` verb that a prior `isWatch` binding had set.

2. The `isWatch` branch was guarded by `if (!acc[key] && ...)`, so it only registered `"watch"` for the first binding per resource key. If a finalize binding created the entry first, the subsequent watch binding was silently skipped — again losing the `"watch"` verb.

Both branches now check whether the key already exists and merge the new verb into the existing `verbs` array (with an `includes` guard to prevent duplicates) rather than overwriting the entry.

Four test changes cover the fix across two levels:

* **`helpers.test.ts`**: Two unit tests for `createRBACMap` — watch-then-finalize and finalize-then-watch binding order — each using sorted equality (`toSorted().toEqual(["patch", "watch"])`) to assert exact verb content without duplicates.

* **`rbac.test.ts`**: One integration test for `clusterRole` in scoped mode using a new `capabilityWithWatchAndFinalize` fixture from `defaultTestObjects.ts`. Asserts a single configmaps rule with exactly `["patch", "watch"]` after deduplication.

* **`rbac.test.ts`**: Updated the existing finalize-only test expectation — the peprstores entry now correctly retains all four verbs (`["patch", "create", "get", "watch"]`) from the merge of custom RBAC and the hardcoded peprstore entry, rather than the prior `["patch"]` which was an artifact of the overwrite bug.

End to End Test:    

## Type of change

## Checklist before merging


samayer12:
@greptileai

PR #2978, opened 2026-03-01
chore: add missing `adjudicateMisboundNamespace` check to admission webhook filter

joonas:
## Description

`shouldSkipRequest` omitted the `adjudicateMisboundNamespace` adjudicator that `filterNoMatchReason` (the watch path) already included. This meant bindings targeting `kind: Namespace` with a namespace filter (a configuration error since namespaces don't live inside namespaces) were not rejected on the admission side. Instead, the request fell through to
`adjudicateMismatchedNamespace`, which produced a misleading error:
```
"Binding defines namespaces ... but Object carries ''"
```
rather than the correct "Cannot use namespace filter on a namespace object" rejection.

Additionally, `misboundNamespace` only checked
`definesNamespaces` (literal namespace list) but not `definesNamespaceRegexes`. A `kind: Namespace` binding with `regexNamespaces` would slip through in both the admission and watch paths with the same misleading mismatch message.

Two fixes:

1. Add `adjudicateMisboundNamespace(binding)` to the `shouldSkipRequest` adjudicator list in `filter.ts`, placed before the namespace mismatch checks so the configuration error is caught early.

2. Broaden `misboundNamespace` in `postCollection.ts` from `allPass([bindsToNamespace, definesNamespaces])` to `allPass([bindsToNamespace, anyPass([definesNamespaces, definesNamespaceRegexes])])`, covering both filter variants.

Tests added for both admission and watch paths with literal and regex namespace filters, plus unit tests for the broadened predicate.

End to End Test:    

## Type of change

## Checklist before merging


samayer12:
@greptileai

AmberFryar:
Moving to an upstream branch #3104  to fix CI secrets access. Cherry-picked with original authorship preserved.

PR #2980, opened 2026-03-01
chore: reject when `loadCapabilities` child exits without sending capabilities

joonas:
## Description

`loadCapabilities` forks a child process in build mode and waits for it to call `process.send()` with the module's capability list. It only registered `"message"` and `"error"` handlers. If the child crashed or exited before sending a message (for example, due to a syntax error or missing dependency) the returned promise would hang forever with no error surfaced to the caller.

Add a `program.on("close", ...)` handler that rejects the promise when the child's IPC channel closes without a message having been delivered. We use `"close"` rather than `"exit"` because `"close"` fires after all IPC/stdio streams are drained, guaranteeing that any in-flight `process.send()` from the child has already been delivered as a `"message"` event. The `"exit"` event can race ahead of queued IPC messages, which would cause intermittent false rejections. A `settled` flag guards all three handlers so the promise is only resolved or rejected once regardless of event ordering.

The rejection message distinguishes between exit-code and signal-based terminations, meaning `close(1, null)` produces `"exited with code 1"` while `close(null, "SIGTERM")` produces `"exited with signal SIGTERM"`, so that operators can tell whether the child crashed on its own or was killed externally.

Three new tests cover the fix: exit-code rejection, signal-based rejection with exact message assertion, and a settled-flag guard verifying that `"close"` after a successful `"message"` is a no-op.

End to End Test:    

## Type of change

## Checklist before merging


samayer12:
@greptileai

PR #2981, opened 2026-03-01
chore: clone `config.env` in `genEnv` to prevent shared-object mutation

joonas:
## Description

`genEnv` deleted `config.env["PEPR_WATCH_MODE"]` directly from the caller's `ModuleConfig` reference. Since the same config object is passed twice (once for the admission deployment and once for the watcher deployment) the first call permanently removed the key. Any code reading `config.env` after the first `genEnv` call would no longer see `PEPR_WATCH_MODE`.

Shallow-clone `config.env` into a local `cfg` object and delete the key from the clone instead. This preserves the original config while still preventing a user-supplied `PEPR_WATCH_MODE` from overriding the programmatic value controlled by the `watchMode` parameter.

Both regression tests assert `config.env` identity via snapshot comparison rather than checking `genEnv` return values. The `def` object always supplies `PEPR_WATCH_MODE` from the `watchMode` parameter, so output-based assertions pass even with the mutation bug present. Existing tests already cover output correctness for each mode.

End to End Test:    

## Type of change

## Checklist before merging


cmwylie19 (approved):

samayer12:
@greptileai

AmberFryar:
Moving to an upstream branch #3103 to fix CI secrets access. Cherry-picked with original authorship preserved.

AmberFryar (approved):

PR #3049, opened 2026-04-15
chore: added metrics summary

AmberFryar:
## Description
Improves soak test observability by adding structured metric output to the Workdlow Summary and a downloadable CSV artifact.                                                                                    
                                                                                                              
  Previously, metrics were only visible by digging through raw log artifacts.                                                 
                           
  CSV artifact (logs/metrics.csv) records all three metrics once per iteration (~every 5 minutes) with a UTC timestamp.                                                                       
                                                                                                                
  The summary table is generated and displays metric totals.  
- pepr_cache_miss is a gauge per time window — cache misses are split into startup vs mid-run, since startup  
  misses are expected and should not indicate a problem                                                         
  - pepr_resync_failure_count uses a label to encode the failure count ({count="0"}) — the label value is
  extracted rather than the gauge value                                                                         
                                                                                                                
  Example output on a run (either Test Passed or Failure Reason will display):
          
Screenshot 2026-04-15 at 4 45 55 PM

  
...

End to End Test:    

## Related Issue

Fixes #

Relates to #

## Type of change

## Checklist before merging

### DIFF
--- a/scripts/soak-record-metrics.sh
+++ b/scripts/soak-record-metrics.sh
@@ -15,14 +15,16 @@ INFORMER_LOG="${3:?informer log path required}"
 METRICS_CSV="${4:?metrics csv path required}"
 
 # watch_controller_failures_total: simple counter, take the value directly
-cf=$(grep -v "^#" "$AUDITOR_LOG" | grep "watch_controller_failures_total" | awk '{print $NF}' | tail -1) || true
+ctrl_failures=$(grep -v "^#" "$AUDITOR_LOG" | grep "watch_controller_failures_total" | awk '{print $NF}' | tail -1) || true
 
 # pepr_cache_miss: gauge per time window e.g. pepr_cache_miss{window="..."} 16 — sum all windows
-cm=$(grep -v "^#" "$INFORMER_LOG" | grep "pepr_cache_miss" | awk '{sum += $NF} END {print sum+0}') || true
+cache_misses=$(grep -v "^#" "$INFORMER_LOG" | grep "pepr_cache_miss" | awk '{sum += $NF} END {print sum+0}') || true
 
 # pepr_resync_failure_count: label holds failure count e.g. {count="0"} 1 — sum non-zero count labels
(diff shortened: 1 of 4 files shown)


samayer12 (changes requested):
It'll be really nice to have a summary for soak tests! Given that these are somewhat complex shell scripts, I'd prefer that we store them in the `scripts/` directory so that we can execute them locally and pass them through shellcheck. We also have the option to rewrite these scripts in Typescript and call them in the `soak.yaml` file.

Regardless, the implementation outcome we should avoid is the introduction of complex shell scripts stored directly in YAML files.

scripts/soak-record-metrics.sh
```diff
@@ -0,0 +1,28 @@
...
+
+# pepr_cache_miss: gauge per time window e.g. pepr_cache_miss{window="..."} 16 — sum all windows
+cm=$(grep -v "^#" "$INFORMER_LOG" | grep "pepr_cache_miss" | awk '{sum += $NF} END {print sum+0}') || true
+
+# pepr_resync_failure_count: label holds failure count e.g. {count="0"} 1 — sum non-zero count labels
+rf=$(grep -v "^#" "$INFORMER_LOG" | grep "pepr_resync_failure_count" | sed 's/.*count="\([^"]*\)".*/\1/' | awk '$1+0 > 0 {sum += $1+0} END {print sum+0}') || true
+
+cf="${cf:-0}"; cm="${cm:-0}"; rf="${rf:-0}"
+
+printf "%s,%s,%s,%s,%s\n" "$ITERATION" "$(date -u +%Y-%m-%dT%H:%M:%SZ)" "$cf" "$cm" "$rf" >> "$METRICS_CSV"
```
samayer12: This works, but let's not use two-character variable names. In practice, I find that they're not very apparent when one must revisit or maintain a script several months after initial implementation.

scripts/soak-summary.sh
samayer12, line 1: See other comment about short variable names & long-term maintenance.

samayer12 (changes requested):
Great, this can all be tracked under shellcheck and executed locally for troubleshooting! Left some changes for variable names that I think we should apply for the sake of clarity.

samayer12 (approved):

PR #3055, opened 2026-04-18
chore: add workflow to check for stale cve suppressions

AmberFryar:
`.grype.yaml` CVE suppression list requires manual maintenance to remove entries that are no longer needed. This PR adds an automated weekly workflow to audit the suppression list and remove stale entries. The workflow scans `pepr:dev` with grype. Grype's output identifies which suppressed CVEs it actually matched against the image  (ignoredMatches). Any entry in `.grype.yaml` that wasn't matched is stale — the CVE is no longer present in the image and the suppression can be safely removed.                       
                                                                                                           
  This workflow only removes suppressions. It never adds them.                                           
                                                                                                         
  How it works

  1. Builds `pepr:dev` and scans it with grype
  2. Compares `.grype.yaml` entries against the scan's ignoredMatches — any suppression not matched is stale              
  3. Removes stale simple entries from `.grype.yaml`                      
  4. Opens or updates a PR on the grype/suppression-audit branch with the list of removed suppressions     
                                                                                                           
  Files                                                                                                    
                                                                                                         
  - `.github/workflows/grype-suppression-audit.yaml` — runs weekly on Mondays at 8am UTC, also triggerable   
  via workflow_dispatch                                                                                  
  - `.github/workflows/scripts/audit-grype-suppressions.ts` — orchestration: scan, detect, remove, commit, PR
   creation                                                                                                
  - `.github/workflows/scripts/audit-grype-suppressions.lib.ts` — pure functions for suppression analysis and
   YAML manipulation                                                                                       
  - `.github/workflows/scripts/audit-grype-suppressions.test.ts` — 18 table-driven tests covering the core 
  logic                                                                                                    
  - `.github/workflows/scripts/tsconfig.json` — TypeScript config for the CI scripts (separate from the main
  build)                                                                                                   
                                                                                                         
  Other changes                                                                                            
                                                                                                         
  - config/vitest.root.config.ts — added .github/**/*.test.ts to test includes                             
  - package.json — added .github/workflows/scripts/**/*.ts to lint glob                                  
  - docs/contribute/contributor-guide.md — documented the workflow and bot-owned branch     
...

End to End Test:  <!-- if applicable -->  

## Related Issue

Fixes #
<!-- or -->
Relates to #

## Type of change

## Checklist before merging


.github/workflows/grype-suppression-audit.yaml
```diff
@@ -0,0 +1,69 @@
...
+        uses: step-security/harden-runner@6c3c2f2c1c457b00c10c4848d6f5491db3b629df # v2.18.0
+        with:
+          egress-policy: audit
+
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - name: Use Node.js latest
+        uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f # v6.3.0
+        with:
+          node-version: 20
```
samayer12, line 31:
```suggestion
          node-version: 24
```

The current [node.js release](https://nodejs.org/en/about/previous-releases) is `24`, is there a particular reason why we opt for node `20` here?

.github/workflows/scripts/grype-suppression-summary.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: 2023-Present The Pepr Authors
+#
+# Writes a GitHub Actions workflow summary for the grype suppression audit.
+#
+# Expected environment variables:
+#   STALE_COUNT   - number of stale suppressions found
+#   STALE_LIST    - newline-separated list of stale IDs
```
samayer12, line 9: Instead of environment variables, which bake in some assumptions that aren't obvious to a future reader, we should pass this information as parameters to the script. Something like `script.sh --stale-count 2 --stale-list cve-123,cve-456`

samayer12 (changes requested):
Since we're doing somewhat complex data processing here and there's a good bit of string-manipulation and file management I'd like to consider the pros/cons of rewriting this check in Typescript for the sake of maintenance. We can also probably bring in some libraries to handle the string-processing and file I/O for us.

We _could_ accept an additional 200-ish lines of bash to automate away this developer toil, but I'd like to evaluate other ways to manage a net-new automation before resorting to yet another bash script. My litmus test here is that we're doing more than ~50 lines worth of work and the actions are more complex than simple `mv`/`cp` commands that don't require in-depth bash familiarity. If we decide bash makes the most sense, then we can proceed.

.github/workflows/scripts/grype-suppression-summary.sh
AmberFryar: Done!

.github/workflows/scripts/audit-grype-suppressions.ts
```diff
@@ -0,0 +1,327 @@
...
+  ignore?: { vulnerability: string; [key: string]: unknown }[];
+  [key: string]: unknown;
+}
+
+interface GrypeMatch {
+  vulnerability: { id: string };
+  relatedVulnerabilities?: { id: string }[];
+}
+
+interface GrpeScanResult {
```
samayer12:
```suggestion
interface GrypeScanResult {
```

Looks like a typo here, we should update other references on the PR

.github/workflows/scripts/audit-grype-suppressions.ts
```diff
@@ -0,0 +1,327 @@
...
+  );
+  return JSON.parse(json) as GrpeScanResult;
+}
+
+function collectIgnoredIds(scan: GrpeScanResult): Set<string> {
+  const ids = new Set<string>();
+  for (const m of scan.ignoredMatches ?? []) {
+    ids.add(m.vulnerability.id);
+    for (const r of m.relatedVulnerabilities ?? []) {
+      ids.add(r.id);
```
samayer12:
```suggestion
  for (const match of scan.ignoredMatches ?? []) {
    ids.add(match.vulnerability.id);
    for (const relatedVuln of match.relatedVulnerabilities ?? []) {
      ids.add(relatedVuln.id);
```

As a personal preference, I believe we should avoid single-character variable names. When code is revisited months later, adding extra clarity is helpful even in relatively simple cases like this one.

.github/workflows/scripts/audit-grype-suppressions.ts
```diff
@@ -0,0 +1,327 @@
...
+    for (const id of staleIds) {
+      lines.push(`- ${id}`);
+    }
+    lines.push("");
+
+    if (prOutcome === "success") {
+      lines.push(
+        "## PR created or updated",
+        "",
+        "The above suppressions were removed from `.grype.yaml`. A PR has been opened or updated for review.",
```
samayer12: Bonus points if we can link to the PR that was just created

samayer12 (changes requested):
Left some observations here and I'll pass along a more detailed review via slack.

.github/workflows/scripts/tsconfig.json
samayer12, line 1: Let's collapse .github/workflows/scripts/tsconfig.json by extending the root tsconfig.
 
We have six `tsconfig*.json` files right now (config/tsconfig.{root,integration,test}.json, src/templates/tsconfig{,.module}.json, plus this one), and the duplicated module/target/strict/esModuleInterop/moduleResolution keys make it easy for them to drift.

We can shrink this config file from 11 lines to 4 by using extends, which is the idiomatic way TypeScript suggests sharing config across multiple tsconfig.json files in a repo:

```
{
   "extends": "../../../config/tsconfig.root.json",
   "include": ["./*.ts"]
}
```

Further reading:
- https://www.typescriptlang.org/tsconfig#extends
- https://www.typescriptlang.org/tsconfig#include
- https://typescript-eslint.io/getting-started/typed-linting/

samayer12 (changes requested):
Left a comment around a pattern I see in the project where we duplicate tsconfig data in several places. Aside from that, I'm ready to merge this and see what we get!

.github/workflows/scripts/tsconfig.json
AmberFryar: Because the root tsconfig sets "rootDir": "../src", and tsc enforces that all included files live under rootDir- even with --noEmit. Since these scripts are in .github/, not src/, we have to override it.

samayer12 (approved):

PR #3062, opened 2026-04-21
chore: soak test scripts to typescript

### LINKED ISSUE #3061: Soak Test Scripts to Typescript
The Soak Test scripts are long and messy. Rewrite them in TypeScript for better maintainability.

AmberFryar:
## Description
  Converted soak-test.sh, soak-summary.sh, soak-test-metrics.sh,                
  soak-record-metrics.sh to .ts files to reduce complexity and increase
  maintainability.                                                              
                                                                              
  Key changes

  - Shell → TypeScript rewrite — Replaced all four bash soak test scripts with  
  typed TypeScript equivalents (soak-test.ts, soak-summary.ts,
  soak-record-metrics.ts) plus a shared constants module (soak-constants.ts)    
  - Pod stability check — Replaced counter-based Map<string, number> approach 
  with a Set-based snapshot diff (Set<string> baseline vs current), reporting   
  all recreated pods in a single error
  - .csv path validation — Added guard that metricsCsvPath must end in .csv to  
  prevent silent data loss from String.replace no-op; switched to slice(0, -4)  
  for .prev path construction
  - First-iteration delta baseline — On first call (no .prev file), deltas are  
  now 0 instead of treating cumulative counter values as deltas, preventing     
  false failures from pre-existing counter history
  - Testability refactoring — Exported core functions with parameterized logsDir
   defaults, added CLI guards so files can be both imported and run directly    
  - Extracted helpers — computeDelta, readPrevState, assertResyncFailures to
  meet ESLint max-statements and complexity limits                              
                                                                              
  Test coverage                                                                 
                                                                              
  - soak-test-kubectl.test.ts — Mocked execSync tests for collectMetrics,
  failWithReason, fetchPodNames, checkPodStability, assertMetrics,              
  assertCacheMissGrowth
  - soak-summary.test.ts — Tests for buildCacheMissTrend, buildSummaryLines,    
  writeSummary                                                                  
  - soak-record-metrics.test.ts — Tests for metric parsing, delta computation,
  CSV validation, counter resets                                                
  - soak-constants.test.ts — Tests for parseEnvNumber and constant values   

End to End Test:    

## Related Issue

Fixes #

Relates to #

## Type of change

## Checklist before merging


scripts/soak-record-metrics.ts
```diff
@@ -0,0 +1,82 @@
...
+  process.exit(1);
+}
+if (!informerLog) {
+  console.error("informer log path required");
+  process.exit(1);
+}
+if (!metricsCSV) {
+  console.error("metrics csv path required");
+  process.exit(1);
+}
```
samayer12: I'd recommend using [commander](https://github.com/tj/commander.js/) here since hand-rolled CLIs are notoriously brittle and it's already a project dependency.

scripts/soak-summary.ts
```diff
@@ -0,0 +1,98 @@
...
+  "",
+  "| Metric | Value | Status |",
+  "|--------|-------|--------|",
+  `| \`watch_controller_failures_total\` | ${totalCtrlFailures} | ${ctrlFailuresStatus} |`,
+  `| \`pepr_cache_miss\` | ${cacheMissesDisplay} | ${cacheMissesStatus} |`,
+  `| \`pepr_resync_failure_count\` | ${resyncFailuresDisplay} | ${resyncFailuresStatus} |`,
+  "",
+  `**Iterations completed:** ${iters} / 70 | **Duration:** ~${Number(iters) * 5} minutes`,
+  "",
+];
```
samayer12: This is a pretty simple markdown table and our script has a narrow use, I could go either way on adding in a table-generation library such as [markdown-table](https://www.npmjs.com/package/markdown-table). Up to you!

scripts/soak-test.ts
```diff
@@ -0,0 +1,177 @@
...
+
+const sleep = (ms: number): Promise<void> => new Promise(resolve => setTimeout(resolve, ms));
+
+fs.mkdirSync(LOGS_DIR, { recursive: true });
+fs.writeFileSync(`${LOGS_DIR}/auditor-log.txt`, "");
+fs.writeFileSync(`${LOGS_DIR}/informer-log.txt`, "");
+fs.writeFileSync(`${LOGS_DIR}/watch-logs.txt`, "");
+fs.writeFileSync(
+  `${LOGS_DIR}/metrics.csv`,
+  "iteration,timestamp,watch_controller_failures_delta,pepr_cache_miss_delta,pepr_resync_failure_count\n",
```
samayer12: What might the update process look like if we added additional fields to the csv or rearranged values? Would that be an easy change that can leverage some of the typing support we get from typescript? While we anticipate this to be low-churn, the extensiblity of this format is worth considering.

Should we keep reporting as `.csv` data? Maybe reporting as JSON is more useful? Regardless, that would be work for a separate PR. Just wanted to note it here.

scripts/soak-test.ts
```diff
@@ -0,0 +1,177 @@
...
+    );
+    execSync("kubectl logs deploy/pepr-soak-ci-watcher -n pepr-system", { stdio: "inherit" });
+    process.exit(1);
+  }
+}
+
+function failWithReason(reason: string): never {
+  fs.writeFileSync(`${LOGS_DIR}/failure-reason.txt`, reason);
+  collectMetrics();
+  process.exit(1);
```
samayer12: Generally, I'd advise against calling `process.exit(1)` directly as noted [here](https://www.appmarq.com/public/tqi,1020876,Avoid-using-nodejs-processexit-TypeScript).

scripts/soak-test.ts
```diff
@@ -0,0 +1,177 @@
...
+    );
+  }
+}
+
+function checkPodStability(iteration: number): void {
+  updatePodMap();
+
+  execSync("kubectl get pods -n pepr-demo", { stdio: "inherit" });
+  execSync("kubectl top po -n pepr-system", { stdio: "inherit" });
+  execSync("kubectl get po -n pepr-system", { stdio: "inherit" });
```
samayer12:
```suggestion
  execSync("kubectl get pods -n pepr-demo", { stdio: "inherit" });
  execSync("kubectl top pods -n pepr-system", { stdio: "inherit" });
  execSync("kubectl get pods -n pepr-system", { stdio: "inherit" });
```

If we're going to automate the kubectl command, we may as well use the entire command instead of shorthand

scripts/soak-test.ts
```diff
@@ -0,0 +1,177 @@
...
+
+  for (const [pod, count] of podMap) {
+    checkPod(pod, count, iteration);
+  }
+}
+
+async function main(): Promise<void> {
+  updatePodMap();
+
+  for (let i = 1; i <= 70; i++) {
```
samayer12: Should the number of iterations be easily-tunable here instead of a hardcoded 70? If we decide to do so, that should happen in a separate PR since this is just a rewrite from bash to ts

samayer12 (changes requested):
Generally good, I like the re-write. I identified a few questions that may be better-suited as a follow-on PR instead of adding directly to this PR.

scripts/soak-summary.ts
AmberFryar: Since it is such a simple table, I don't know if it is worth another dependency.

scripts/soak-test.ts
AmberFryar: I opened an [Issue](https://github.com/defenseunicorns/pepr/issues/3091) for the column values. As far as the `.csv` format goes, the choice was intentional for human readability. Engineers debugging a soak test failure can scan the raw output directly to identify the iteration where the failure began and help pinpoint the root cause. Open to changing it if the `.csv` was the wrong decision.

scripts/soak-test.ts
AmberFryar: Opened Issue #3092

AmberFryar:
@greptileai review

scripts/soak-test.ts
samayer12: I think `.csv` works here, especially given how we've used soak test results in the past. Perhaps in the future we could also support JSON for machine-readable parsing of data but that's not required right now.

samayer12 (dismissed):

AmberFryar:
@greptileai review

AmberFryar:
@greptileai review

scripts/soak-test.ts
```diff
@@ -30,16 +30,18 @@ fs.writeFileSync(
   "iteration,timestamp,watch_controller_failures_delta,pepr_cache_miss_delta,pepr_resync_failure_count\n",
 );
 
-let previousPods = new Set<string>();
+const podMap = new Map<string, number>();
```
samayer12: Why did this return type change to `void`? I see that `podMap` now lives outside of the `main()` function and is globally-scoped to anything in this file.

scripts/soak-test.ts
samayer12, line 1: The control-structure of `main()` seems like it could be improved in a few ways. Here's what I suggest:

* Reduce the amount of magic numbers floating around in this part of the code (I thought we turned on [`no-magic-numbers`](https://eslint.org/docs/latest/rules/no-magic-numbers) from eslint? Perhaps I'm mistaken).
* The `if(i<70)` check for calls to `sleep()` is the sort of situation that causes the [fencepost error](https://en.wikipedia.org/wiki/Off-by-one_error#Fencepost_error). What do you think of switching from an iteration-driven loop to a time-bound loop? That could look something like `while (Date.now() < endTime)` instead. Since the original implementation used this iterative structure, I'm also okay leaving it as-is for now. Up to you!
* We should move test-setup actions into `main()` if that's the structure we're going to use. For example, the `fs.*` calls at the top of this file are part of the test script setup, but live outside of main. So, we're currently interleaving function definitions and consts with actual test behavior, which is a little tricky to read.

samayer12 (changes requested):
Some more structural feedback around the `main()` function now that I looked at it in the context of the rest of `soak-test.ts`. Once we address those open questions I think we'll be good to merge with this conversion from shell to TS

scripts/soak-test.ts
```diff
@@ -0,0 +1,248 @@
...
+  });
+  execSync("kubectl get pods -n pepr-system", {
+    stdio: "inherit",
+    timeout: KUBECTL_TIMEOUT_MS,
+  });
+
+  for (const [pod, count] of podMap) {
+    checkPod(pod, count, elapsedMinutes);
+  }
+}
```
samayer12: There's potentially an issue here with our pod stability check. Here's a finding from an agent that we can consider:

>   1. checkPodStability guarantees failure on the first pod stability check
> 
>   scripts/soak-test.ts:140-173, 214 — Confidence 95
> 
> ```
>   // main(), line 214:
>   updatePodMap(podMap);  // seeds all pods at count=1
> 
>   // iteration 2 → checkPodStability → line 155:
>   function checkPodStability(podMap, elapsedMinutes): void {
>     updatePodMap(podMap);  // every alive pod now has count=2
>     for (const [pod, count] of podMap) {
>       checkPod(pod, count, elapsedMinutes);  // fails if count > 1
>     }
>   }
> ```
> 
>   Trace: after the seeding call in main(), every running pod has count=1. On iteration 2 (POD_CHECK_INTERVAL = 2), checkPodStability calls updatePodMap
>   again — every pod still alive increments to count=2. checkPod at line 63 throws failWithReason("Pod X was recreated (seen 2 times) ..."). The test cannot
>   pass unless every pepr-demo pod terminates between iterations 0 and 2 — which is the opposite of what the test claims to verify.
> 
>   This was pre-existing in the original code. This PR touched updatePodMap (parameterized away from module-level state) and did not fix it. It must be fixed
>    now.
> 
>   Fix: The semantics should be "fail if a pod name appears that was not present at startup." Replace the counter with a snapshot diff:
> 
> ```
>   // In main(): seed the baseline set
>   const initialPods = new Set(fetchPodNames());
> 
>   // In checkPodStability(): compare current against baseline
>   function checkPodStability(initialPods: Set<string>, elapsedMinutes: number): void {
>     const current = fetchPodNames();
>     const recreated = current.filter(p => !initialPods.has(p));
>     for (const pod of recreated) {
>       failWithReason(`Pod ${pod} is new (possible recreation) ~${elapsedMinutes} minutes in`);
>     }
>   }
> ```
> 
>   A pod that appears for the first time after startup is a recreation. Healthy pods that persist are invisible. Pods that terminate are removed from current
>    and simply absent — not reported as recreated (unless their replacement has a new name, which is the correct Kubernetes behavior for Deployment pod
>   restarts).

scripts/soak-record-metrics.ts
```diff
@@ -0,0 +1,116 @@
...
+}: RecordMetricsArgs): void {
+  const auditorContent = fs.readFileSync(auditorLogPath, "utf-8");
+  const informerContent = fs.readFileSync(informerLogPath, "utf-8");
+
+  const ctrlFailures = parseCtrlFailures(auditorContent);
+  const cacheMisses = parseCacheMisses(informerContent);
+  const resyncFailures = parseResyncFailures(informerContent);
+
+  // Compute deltas using a state file that tracks the previous cumulative values.
+  const prevStatePath = metricsCsvPath.replace(/\.csv$/, ".prev");
```
samayer12: Here's a valid finding that doesn't trigger in our case, but is good to consider for the sake of input-validation. In our case, we can ignore this since the YAML file is correct.

>   2. .prev state path derivation silently overwrites the metrics CSV
> 
>   scripts/soak-record-metrics.ts:78, 94 — Confidence 98
> 
> ```
>   const prevStatePath = metricsCsvPath.replace(/\.csv$/, ".prev");
>   // ...
>   fs.writeFileSync(prevStatePath, `${ctrlFailures},${cacheMisses}\n`);
> 
> ```
>   String.prototype.replace returns the original string unchanged when the pattern does not match. If metricsCsvPath does not end in .csv (typo in CI YAML,
>   alternate test invocation, CLI caller using .txt, any of dozens of plausible scenarios), then prevStatePath === metricsCsvPath. The writeFileSync then
>   clobbers the metrics CSV with "X,Y\n" after potentially hours of accumulated data. All subsequent appendFileSync calls corrupt the file further. Silent,
>   total data loss.
> 
>   The current call site and all tests use .csv, so this branch has never been exercised. It will be discovered at the worst possible time.
> 
>   Fix:
> 
> ```
>   if (!metricsCsvPath.endsWith(".csv")) {
>     throw new Error(`metricsCsvPath must end in .csv, got: ${metricsCsvPath}`);
>   }
>   const prevStatePath = `${metricsCsvPath.slice(0, -4)}.prev`;
> 
> ```
>   Add a test passing a non-.csv path and asserting the CSV is not overwritten.

scripts/soak-record-metrics.ts
```diff
@@ -0,0 +1,116 @@
...
+
+  let prevCtrl = 0;
+  let prevCache = 0;
+
+  if (fs.existsSync(prevStatePath)) {
+    const parts = fs.readFileSync(prevStatePath, "utf-8").trim().split(",");
+    prevCtrl = Number(parts[0]) || 0;
+    prevCache = Number(parts[1]) || 0;
+  }
```
samayer12: There's a finding for an instance where the environment for test setup could have an issue. Given our CI environment (which seems pretty stable?) I think this finding sounds dramatic but is probably not an issue in our environment. We could take it or leave it.


>   3. First-iteration delta attributes all pre-existing counter history to iteration 1, causing assertMetrics to immediately fail
> 
>   scripts/soak-record-metrics.ts:80-86, 130 / scripts/soak-test.ts:120-123 — Confidence 80
> 
>   On the very first call to recordMetrics, no .prev file exists, so prevCtrl = 0 and prevCache = 0. If the pepr-soak-ci-watcher has accumulated any
>   controller failures before the test starts (prior run artifacts, startup noise, test fixtures), then ctrlFailuresDelta = ctrlFailures - 0 = N > 0. Then
>   assertMetrics at line 122 checks ctrlFailures !== 0 and calls failWithReason immediately at iteration 1.
> 
>   The existing test "appends a row to the CSV with correct deltas on first iteration" asserts cols[2]).toBe("5") — it encodes this broken behavior as the
>   expected outcome. A test that certifies the wrong invariant is worse than no test.
> 
>   Fix: On the first iteration (no .prev file), record delta=0 and treat the current values as the baseline, not as the delta:
> 
>   ```
> const isFirstIteration = !fs.existsSync(prevStatePath);
>   const ctrlFailuresDelta = isFirstIteration ? 0
>     : (ctrlFailures >= prevCtrl ? ctrlFailures - prevCtrl : ctrlFailures);
>   const cacheMissesDelta = isFirstIteration ? 0
>   
> ```  : (cacheMisses >= prevCache ? cacheMisses - prevCache : cacheMisses);
> 
>   Update the test to assert cols[2]).toBe("0") for iteration 1.

samayer12 (changes requested):
I highlighted a few findings from an agent-assisted review in addition to what we noticed about test coverage for this rewrite. I'll share out the detailed summary on slack.

scripts/soak-constants.test.ts
```diff
@@ -0,0 +1,36 @@
...
+    expect(parseEnvNumber("0", 10)).toBe(0);
+  });
+
+  it("parses negative numbers", () => {
+    expect(parseEnvNumber("-5", 10)).toBe(-5);
+  });
+
+  it("parses floating point numbers", () => {
+    expect(parseEnvNumber("3.14", 10)).toBe(3.14);
+  });
```
samayer12: This is a good opportunity for a table-driven test here since the function we're checking is `parseEnvNumber()` without any complex mock setup.

scripts/soak-test.ts
```diff
@@ -0,0 +1,255 @@
...
+export function fetchPodNames(): string[] {
+  const output = execSync("kubectl get pods -n pepr-demo -o jsonpath='{.items[*].metadata.name}'", {
+    timeout: KUBECTL_TIMEOUT_MS,
+  })
+    .toString()
+    .trim();
+
+  if (!output) return [];
+
+  return output.split(" ");
```
samayer12: `split(" ")` is fragile: kubectl output with trailing whitespace, tabs, or a future jsonpath format change produces empty strings in `initialPods`, causing `checkPodStability` to fire false positives for the rest of the run. `parseCacheMisses` already uses `/\s+/`; this should too.

```suggestion
  return output.split(/\s+/).filter(Boolean);
```

scripts/soak-record-metrics.ts
```diff
@@ -0,0 +1,134 @@
...
+    }, 0);
+}
+
+/** Parse the informer log and return the sum of non-zero resync failure count labels. */
+export function parseResyncFailures(informerContent: string): number {
+  return informerContent
+    .split("\n")
+    .map(line => line.trim())
+    .filter(line => !line.startsWith("#"))
+    .filter(line => line.includes("pepr_resync_failure_count"))
```
samayer12: Unanchored `.includes()` will match a future `pepr_resync_failure_count_created` or `pepr_resync_failure_count_bucket` Prometheus series (standard suffixes), silently inflating the parsed resync total. `parseCacheMisses` on line 31 already uses the anchored form — this should be consistent.

```suggestion
    .filter(line => /^pepr_resync_failure_count[\s{]/.test(line))
```

scripts/soak-test.ts
```diff
@@ -0,0 +1,255 @@
...
+    this.name = "SoakTestFailure";
+  }
+}
+
+export function failWithReason(reason: string, logsDir: string = LOGS_DIR): never {
+  fs.writeFileSync(`${logsDir}/failure-reason.txt`, reason);
+  try {
+    collectMetrics(logsDir);
+  } catch {
+    // collectMetrics requires kubectl — expected to fail in test environments
```
samayer12: The empty catch silently drops the error even in production. If the cluster is unreachable — the exact scenario this test is meant to catch — `collectMetrics` throws here, leaving `failure-reason.txt` with no supporting logs and no indication that collection itself failed.

```suggestion
  } catch (e) {
    console.error("Failed to collect diagnostic metrics during failure handling:", e);
```

samayer12 (commented):
One-liner suggestions from brutal review. See the full review thread for findings that require larger fixes.

.github/workflows/soak.yaml
```diff
@@ -129,14 +129,21 @@ jobs:
...
       - name: Run the soak test and collect metrics
-        run: bash scripts/soak-test.sh
-        shell: bash
+        run: npx tsx scripts/soak-test.ts
 
       - name: Write job summary
         if: always()
-        run: bash scripts/soak-summary.sh logs/metrics.csv logs/informer-log.txt logs/failure-reason.txt "$CACHE_MISS_GROWTH_THRESHOLD"
-        shell: bash
+        run: npx tsx scripts/soak-summary.ts logs/metrics.csv logs/informer-log.txt logs/failure-reason.txt
```
samayer12, line 146: I see that the re-written version doesn't track the `CACHE_MISS_GROWTH_THRESHOLD` field, was this intentional? If not, we should track it with another `commander` option in the CLI parser.

AmberFryar: I shouldn't need it on the CLI.  It is defined in soak.yaml on line 70.  The typescript files don't need it explicitly set in CLI since it inherently has access to the process.env.

AmberFryar:
@greptileai review

AmberFryar:
@greptileai review

samayer12 (approved):
Great work! Thanks for all the attention to detail in the rewrite from bash to ts

PR #3094, opened 2026-05-01
chore: automate peerDependencies maintenance

samayer12:
## Description

Replaces the manual `scripts/peerDeps.sh` copy-paste workflow with a scheduled GitHub Actions automation that owns the `peerDependencies` block of `package.json`.

**Why a new workflow rather than Dependabot.** Dependabot is already configured for npm in `.github/dependabot.yml`, but its provider does not surface peer-only entries as bumpable targets — across 2+ years of repo history there is **zero** Dependabot PR touching the `peerDependencies` block (verified via `git log --grep=peer -i`). Recent peer-dep bumps have all been hand-authored (e9f4694d, 6ba93b7a, 0496372a, c1cdccc6 / #3093). Result: peer deps drift, exposing consumers of `pepr` to stale ranges (the most recent manual catch-up included a TypeScript 5→6 and a uuid 13→14 jump).

**Risk-tiered PR grouping.** The workflow opens one PR for all minor/patch bumps grouped together, plus one PR per major-version bump. Major bumps deserve their own diff so that a breaking TS or uuid major can be reverted independently of the routine prettier/esbuild minors — exactly the failure mode that bit #3093 (one diff conflated 4 bumps including 2 majors).

## Fan-out execution model

The workflow is a two-job pipeline that uses a GitHub Actions matrix to fan out PR creation:

1. **`discover`** runs once. It checks out `main`, queries `npm view <pkg> version` for every entry in `peerDependencies`, classifies each candidate as minor/patch vs. major using `majorOf(version)` from `update-peer-deps.mjs`, and writes a matrix descriptor to `$GITHUB_OUTPUT`:

   ```
   matrix={"include":[{"kind":"minor","pkg":""},{"kind":"major","pkg":"typescript"},{"kind":"major","pkg":"uuid"}]}
   empty=false
   ```

   If `empty=true` (everything already at latest), the second job is skipped via `if: needs.discover.outputs.empty == 'false'` and no PRs are opened.

2. **`propose`** consumes that matrix via `strategy.matrix: ${{ fromJSON(needs.discover.outputs.matrix) }}` and **fans out one parallel job per matrix cell**. With 1 minor group + 2 majors pending, three propose jobs run simultaneously, each producing its own independent PR (`chore/peer-deps/minor` or `chore/peer-deps/major-<pkg>`).

   `fail-fast: false` is the load-bearing knob here: a broken cell (e.g. a TypeScript major that fails `tsc -p`) aborts only that cell and produces no PR for it, while unrelated minor/patch bumps still ship a green PR. Without this flag, one breakage would block all sibling PRs.

Each propose cell runs sequentially: checkout `main` (with `PEER_DEPS_PR_TOKEN` if provisioned, falling back to `GITHUB_TOKEN`) → `npm ci` → `apply --kind $KIND --pkg $PKG` → `npm install` (refresh lockfile) → `format:check` → `build` → `test:unit` → `open-pr`. The three gates (`format:check`, `build`, `test:unit`) run **before** PR creation so that broken bumps never become PRs at all — the matrix cell fails, the operator sees a clean signal in the workflow run, and the team's PR list stays uncluttered. The `In-workflow verification` block in each bot-rendered PR body is anchored to those gates having actually passed.

## Components

| File | Role |
| --- | --- |
| `.github/workflows/peer-deps-update.yml` | Two-job workflow (`discover` → `propose`) described above. Schedule: `0 6 * * MON` + `workflow_dispatch`. Actions are SHA-pinned (`actions/checkout@v6.0.2`, `actions/setup-node@v6.4.0`); no third-party action dependencies. |
| `scripts/update-peer-deps.mjs` | Canonical Node script with subcommand CLI via `commander`: `report [--matrix]`, `apply --kind <kind> [--pkg <name>]`, `pr-body --kind …`, `open-pr --kind …`. Pure logic (`classifyBumps`, `pickUpdates`, `pickBranchAndTitle`, `diffPeerDeps`, `renderPrBody`, `reportToMatrixInclude`, `majorOf`) is exported and isolated from I/O wrappers (`git`, `gh`, `npmLatest`) for testability. CLI bootstrap is gated on `import.meta.url === pathToFileURL(process.argv[1]).href` so test imports do not fire the parser. `open-pr` configures git identity, force-pushes the bot branch, then `gh pr create` with a `gh pr edit` fallback if a PR for that branch is already open (so consecutive runs update one PR rather than spawning new ones). |
| `docs/contribute/contributor-guide.md` | New `Peer-Dependency Updates` subsection under Coding Guidelines describing the workflow, manual dispatch, and the local subcommand-style commands. |
| `scripts/peerDeps.sh` | **Deleted** — the manual bash copy-paste helper is fully replaced by the new Node script's `report` subcommand. |

`.github/dependabot.yml` is unchanged — the two automations never touch the same region of `package.json`.

## PAT prerequisite

`gh pr create` invoked with the default `GITHUB_TOKEN` produces PRs that **do not trigger** `pull_request`-event workflows (a documented GitHub limitation). The workflow falls back to `GITHUB_TOKEN` if `secrets.PEER_DEPS_PR_TOKEN` is unset, but in that fallback path the auto-opened PRs will not run `node.js.yml` and will land without CI gating beyond the in-workflow gates that just ran. Provisioning a project-scoped PAT or GitHub-App token at `secrets.PEER_DEPS_PR_TOKEN` is recommended before the first scheduled run lands a PR.

## In-PR verification

Locally, against this branch:

- `node scripts/update-peer-deps.mjs report` — correctly identifies the same four bumps that #3093 captured (esbuild + prettier as minor; typescript + uuid as major).
- `node scripts/update-peer-deps.mjs report --matrix` — emits the GitHub-Actions-shaped `matrix=` and `empty=` outputs ready for `>> "$GITHUB_OUTPUT"`.
- `node scripts/update-peer-deps.mjs apply --kind minor` and `apply --kind major --pkg <name>` — apply the expected change to `package.json`, exit 0; exit 2 when there is nothing to do; clean error on missing `--pkg` for major.
- `node scripts/update-peer-deps.mjs pr-body --kind <kind> [--pkg <name>]` — renders the PR body markdown for a dry-run preview.
- `npm run format:check` — passes (markdownlint + prettier on the doc edit).
- `npm run build` — passes.
- `node -e "require('js-yaml').load(...)"` against the workflow YAML — parses cleanly; both `discover` and `propose` jobs visible.

## Related Issue

Relates to #

## Type of change

## Checklist before merging


scripts/update-peer-deps.mjs
```diff
@@ -0,0 +1,262 @@
...
+
+const here = dirname(fileURLToPath(import.meta.url));
+const PACKAGE_JSON = resolve(here, "..", "package.json");
+const NO_BUMPS_AVAILABLE = 2;
+const BOT_NAME = "github-actions[bot]";
+const BOT_EMAIL = "41898282+github-actions[bot]@users.noreply.github.com";
+const PR_LABELS = ["dependencies", "automated"];
+
+// ── Pure logic ──────────────────────────────────────────────────────────────
+
```
samayer12:
```suggestion
```

scripts/update-peer-deps.mjs
```diff
@@ -0,0 +1,262 @@
...
+    title: `chore: bump peerDependency ${opts.pkg} to ${peerVersions[opts.pkg]} (major)`,
+  };
+}
+
+export function parseDiffBumps(diff) {
+  const before = {};
+  const after = {};
+  for (const line of diff.split("\n")) {
+    const m = line.match(/^([-+])\s+"([^"]+)":\s+"([^"]+)",?\s*$/);
+    if (m) (m[1] === "-" ? before : after)[m[2]] = m[3];
```
samayer12:
```suggestion
    const match = line.match(/^([-+])\s+"([^"]+)":\s+"([^"]+)",?\s*$/);
    if (match) (match[1] === "-" ? before : after)[match[2]] = match[3];
```

scripts/update-peer-deps.mjs
```diff
@@ -0,0 +1,262 @@
...
+
+### In-workflow verification
+
+- \`npm run format:check\` — passed
+- \`npm run build\` — passed
+- \`npm run test:unit\` — passed
+
+## Related Issue
+
+Relates to #
```
samayer12:
```suggestion
None, ongoing maintenance
```

scripts/update-peer-deps.mjs
```diff
@@ -0,0 +1,262 @@
...
+  const bumps = diskBumps();
+  const peers = readPackageJson().peerDependencies;
+  const { branch, title } = pickBranchAndTitle(opts, peers);
+
+  git(["config", "user.name", BOT_NAME]);
+  git(["config", "user.email", BOT_EMAIL]);
+  git(["checkout", "-B", branch]);
+  git(["add", "package.json", "package-lock.json"]);
+  git(["commit", "-m", title, "--signoff"]);
+  git(["push", "--force", "origin", branch]);
```
samayer12:
```suggestion
  git(["push", "--force-with-lease", "origin", branch]);
```

samayer12:
@greptile review

samayer12:
@greptile review

AmberFryar (dismissed):

AmberFryar (approved):

PR #3104, opened 2026-05-06
chore: add missing `adjudicateMisboundNamespace` check to admission webhook filter

AmberFryar:
- Cherry-picked from #2978 (fork PR by @joonas) to enable CI secrets access

shouldSkipRequest omitted the adjudicateMisboundNamespace adjudicator that filterNoMatchReason (the watch path) already included. This meant bindings targeting kind: Namespace with a namespace filter (a configuration error since namespaces don't live inside namespaces) were not rejected on the admission side. Instead, the request fell through to
adjudicateMismatchedNamespace, which produced a misleading error:

"Binding defines namespaces ... but Object carries ''"
rather than the correct "Cannot use namespace filter on a namespace object" rejection.

Additionally, misboundNamespace only checked
definesNamespaces (literal namespace list) but not definesNamespaceRegexes. A kind: Namespace binding with regexNamespaces would slip through in both the admission and watch paths with the same misleading mismatch message.

Two fixes:

Add adjudicateMisboundNamespace(binding) to the shouldSkipRequest adjudicator list in filter.ts, placed before the namespace mismatch checks so the configuration error is caught early.

Broaden misboundNamespace in postCollection.ts from allPass([bindsToNamespace, definesNamespaces]) to allPass([bindsToNamespace, anyPass([definesNamespaces, definesNamespaceRegexes])]), covering both filter variants.

Tests added for both admission and watch paths with literal and regex namespace filters, plus unit tests for the broadened predicate.

...

End to End Test:    

## Related Issue

Fixes #

Relates to #

## Type of change

## Checklist before merging

### DIFF
--- a/src/lib/filter/adjudicators/postCollection.test.ts
+++ b/src/lib/filter/adjudicators/postCollection.test.ts
@@ -226,16 +226,32 @@ describe("bindsToNamespace", () => {
 describe("misboundNamespace", () => {
   //[ Binding, result ]
   it.each([
-    [{ kind: { kind: "Kind" }, filters: { namespaces: [] } }, false],
-    [{ kind: { kind: "Kind" }, filters: { namespaces: ["namespace"] } }, false],
-    [{ kind: { kind: "Namespace" }, filters: { namespaces: [] } }, false],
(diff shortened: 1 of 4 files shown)


AmberFryar:
Adding an approval gate to merge forked PRs instead of this route of cherry picking.

PR #3171, opened 2026-06-28
validate registry-info format in pepr build

AruneshDwivedi:
The --registry-info flag accepts any string without validation, then silently produces an invalid image reference. Adds a check that the value contains a / separator (expected: registry/username) before using it in the controller image path.

### DIFF
--- a/src/cli/build/build.helpers.ts
+++ b/src/cli/build/build.helpers.ts
@@ -78,6 +78,14 @@ interface ImageOptions {
  * @param imageOptions CLI options for image
  * @returns image string
  */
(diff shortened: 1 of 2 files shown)


samayer12 (changes requested):
Hi @AruneshDwivedi, thanks for the PR! A bare registry hostname without a namespace is a valid value for `--registry-info`. However, our docs would lead someone to think that's not the case. I've opened up #3195 as an issue with #3196 to make that clarification.

samayer12:
We'll consider this addressed by #3196, which notes the optionality of the field.

PR #3192, opened 2026-07-16
feat: add configurable/overridable startup probes to deployments

Metsyu:
## Description

Generated Pepr deployments currently configure liveness and readiness probes (see related issue), but do not configure a startup probe or render startup probe overrides from the Helm chart.

This change adds default startup probes to the admission and watcher deployments and exposes their configuration through Helm values, matching the existing liveness and readiness probe behavior. This provides more flexibility for slow-starting or resource-constrained environments, rather than solely relying on liveness and readiness.

## Related Issue

Fixes #

Relates to defenseunicorns/pepr#1085

## Type of change

- [X] New feature (non-breaking change which adds functionality)
- [X] Other (security config, docs update, etc)

## Checklist before merging

- [X] Unit, Integration, [docs](<https://github.com/defenseunicorns/pepr/tree/main/docs>), [adr](<https://github.com/defenseunicorns/pepr/tree/main/adr>) added or updated as needed
- [X] [Contributor Guide Steps](<https://docs.pepr.dev/main/contribute/#submitting-a-pull-request>) followed

### DIFF
--- a/.github/workflows/peer-deps-update.yml
+++ b/.github/workflows/peer-deps-update.yml
@@ -79,7 +79,11 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
         with:
+<<<<<<< HEAD
           node-version: "22"
+=======
+          node-version: "22.19"
+>>>>>>> 2c37de94 (fix: align minimum Node version with dependency requirements)
           cache: npm
       - name: Install dependencies
         run: npm ci
--- a/.github/workflows/peer-deps-update.yml
+++ b/.github/workflows/peer-deps-update.yml
@@ -79,11 +79,7 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
         with:
-<<<<<<< HEAD
           node-version: "22"
-=======
-          node-version: "22.19"
->>>>>>> 2c37de94 (fix: align minimum Node version with dependency requirements)
           cache: npm
       - name: Install dependencies
         run: npm ci
--- a/integration/cli/dev.dev-test.test.ts
+++ b/integration/cli/dev.dev-test.test.ts
@@ -8,7 +8,8 @@ import { Workdir } from "../helpers/workdir";
 import * as time from "../helpers/time";
 import * as pepr from "../helpers/pepr";
 import { ChildProcessWithoutNullStreams, spawn } from "child_process";
-import { fetch, RequestInit, Agent } from "undici";
+import { fetch } from "kubernetes-fluent-client";
+import { RequestInit, Agent } from "undici";
 
 const FILE = path.basename(__filename);
 const HERE = __dirname;
@@ -163,8 +164,8 @@ const fetchOpts: RequestInit = {
 
 // Wait for the server to start and report healthy
 async function waitForServer(): Promise<void> {
-  const response = await fetch(`${fetchBaseUrl}/healthz`, fetchOpts).catch(() => undefined);
-  if (!response?.ok) {
+  const resp = await fetch(`${fetchBaseUrl}/healthz`, fetchOpts);
+  if (!resp.ok) {
     await sleep(2);
     return waitForServer();
   }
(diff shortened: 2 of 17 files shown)


src/cli/init/templates.ts
```diff
@@ -42,7 +42,10 @@ export type peprPackageJSON = {
     };
     scripts: { "k3d-setup": string };
     dependencies: { pepr: string; undici: string };
-    devDependencies: { typescript: string };
+    devDependencies: {
+      "@types/node": string;
```
Metsyu, line 46: Let me know if you'd like to not see this merge. 

I made this change trying to fix the many `thread-stream` and `@types/node@26` issues/failures I was seeing.

Seems like, from what I can tell several packages are using `@types/node` wildcards, which resolves to `26`. Here I'm explicitly constraining the generated modules to `node@25.x.x`, matching the version used by Pepr itself.

integration/cli/dev.dev-test.test.ts
```diff
@@ -8,8 +8,7 @@ import { Workdir } from "../helpers/workdir";
 import * as time from "../helpers/time";
 import * as pepr from "../helpers/pepr";
 import { ChildProcessWithoutNullStreams, spawn } from "child_process";
-import { fetch } from "kubernetes-fluent-client";
-import { RequestInit, Agent } from "undici";
+import { fetch, RequestInit, Agent } from "undici";
```
Metsyu: Let me know if you'd like to not see this merge.

Small change here fixing some incompatibility issues I was seeing when running tests locally.

src/cli/init/templates.ts
samayer12: I think this is an okay update to the template, but let's use `@types/node:24.13.3` instead of `25`. Based on what I see about the versioning schema, using a `24.x` version matches up with node 24 being in LTS. See #3194 for my suggestion that you can bring into this PR.

integration/cli/dev.dev-test.test.ts
samayer12: KFC's fetch does more than what `undici`'s fetch does, so we should really publish an update to KFC to bring us up to date instead of swapping out the import (such as https://github.com/defenseunicorns/kubernetes-fluent-client/pull/1138)

samayer12 (changes requested):
This is generally good, thanks for the suggestion! As a KFC maintainer, I should look into some overdue dependency maintenance for that project which can help here. For this PR, we should consider how we might also update some of our tests for templates and overrides. Try taking a look at `k8sObjects.test.ts`, `helm.test.ts`, and `overridesFile.ts` as places to add coverage.

integration/cli/dev.dev-test.test.ts
Metsyu: Reverted those changes and updated undici version to hopefully get rid of the incompatibility issues I was seeing before.

Metsyu:
> This is generally good, thanks for the suggestion! As a KFC maintainer, I should look into some overdue dependency maintenance for that project which can help here. For this PR, we should consider how we might also update some of our tests for templates and overrides. Try taking a look at `k8sObjects.test.ts`, `helm.test.ts`, and `overridesFile.ts` as places to add coverage.

Added changes to test files, that should cover all probes: startup, liveness, readiness.

samayer12 (changes requested):
Let's undo the versioning changes for `undici` since I'm working on a KFC release that will upgrade us to version 8 (see https://github.com/defenseunicorns/kubernetes-fluent-client/pull/1182). Once that's merged then we should be ready to bring this change in.

.github/workflows/peer-deps-update.yml
```diff
@@ -33,7 +33,7 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
         with:
-          node-version: "20"
+          node-version: "22.19"
```
samayer12:
```suggestion
          node-version: "22"
```

Just plain 22 should be fine here

.github/workflows/peer-deps-update.yml
```diff
@@ -79,7 +79,7 @@ jobs:
...
```
samayer12:
```suggestion
          node-version: "22"
```

Just plain 22 should be fine here

samayer12 (approved):
CI failures are for known issues with dependencies related to `pino`.

PR #3208, opened 2026-07-22
ci: add resilient setup-k3d composite and harden transient CI flakes

samayer12:
## Description

Addresses network flakes from a triage of failing CI on `main`. These are non-deterministic network/runner failures that surfaced in scheduled runs.

### Root cause

k3d was installed via an inline one-liner duplicated across 11 workflows:

```
curl -s --retry 5 --retry-all-errors --fail https://raw.githubusercontent.com/k3d-io/k3d/main/install.sh | bash
```

`curl --retry` only covers the transport for the script download. Transient failures inside `install.sh` are not retried, and `pipefail` fails the step.

### Changes

- **New composite action** `.github/actions/setup-k3d` — wraps the install in a **5-attempt outer retry loop** with linear backoff, verifies the binary with `k3d version`, and supports an optional pinned `version` (via k3d's `TAG`).
- **All 11 workflows** converted to the composite. The workspace-relative reference differs by each job's checkout topology:
  - pepr at workspace **root** before k3d → `./.github/actions/setup-k3d` (soak, on-demand-workflows, upgrade-upstream, upgrade-unicorn)
  - pepr checked out to `./pepr` → k3d **moved after** that checkout → `./pepr/.github/actions/setup-k3d` (deploy, load, tutorials-test)
  - pepr **not** checked out in the job → **added** a `path: pepr` checkout before k3d → `./pepr/.github/actions/setup-k3d` (uds, pepr-excellent-examples-upstream/unicorn/ironbank-amd)
- **IronBank e2e** — bumped the `Run e2e tests` retry `timeout_minutes` from `8` → `12` so a slow image-load/deploy doesn't fail the matrix leg.

### DIFF
--- a/.github/actions/setup-k3d/action.yaml
+++ b/.github/actions/setup-k3d/action.yaml
@@ -12,19 +12,15 @@ runs:
   using: composite
   steps:
     - name: install k3d
-      shell: bash
+      uses: nick-fields/retry@ad984534de44a9489a53aefd81eb77f87c70dc60 # v4.0.0
       env:
         TAG: ${{ inputs.version }}
-      run: |
-        set -euo pipefail
-        for attempt in 1 2 3 4 5; do
-          if curl -sfL --retry 5 --retry-all-errors --retry-delay 5 \
-               https://raw.githubusercontent.com/k3d-io/k3d/main/install.sh | bash; then
-            k3d version
-            exit 0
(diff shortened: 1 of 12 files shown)


samayer12:
@greptile review

AmberFryar (approved):

PR #3292, opened 2026-08-26
chore: fix no-binding chart admission deployment

AmberFryar:
## Description

[The UDS Dependency Operator](https://github.com/defenseunicorns/dependency-operator) works when deployed from Pepr’s generated raw manifests, but fails when using `pepr build -z chart`.

The UDS Dependency Operator does not use standard Pepr bindings such as `When(...).Watch(...)`, `Mutate(...)`, or `Validate(...)`. Instead, it starts lower-level Kubernetes watches from the Pepr runtime. The raw manifest generator handles this no-binding case by generating the admission controller deployment, but the Helm chart generator did not.

As a result, the generated chart set `admission.enabled: true` but did not write `templates/admission-deployment.yaml`, so no UDS Dependency Operator controller pods were created.

## Changes

- Align Helm chart admission deployment generation with the existing manifest generation behavior.
- Generate `admission-deployment.yaml` when the module has admission bindings or has neither admission nor watcher bindings.
- Keep webhook templates gated on actual mutating/validating webhook objects.
- Add a regression test for no-binding modules using chart output....

End to End Test:    

## Related Issue

Fixes #

Relates to #

## Type of change

## Checklist before merging


src/lib/assets/assets.ts
```diff
@@ -181,8 +181,9 @@ export class Assets {
     validateWebhook: V1MutatingWebhookConfiguration | V1ValidatingWebhookConfiguration | null,
     mutateWebhook: V1MutatingWebhookConfiguration | V1ValidatingWebhookConfiguration | null,
     helm: Record<string, Record<string, string>>,
+    deployAdmissionController: boolean = Boolean(validateWebhook || mutateWebhook),
   ): Promise<void> => {
-    if (validateWebhook || mutateWebhook) {
+    if (deployAdmissionController) {
```
samayer12: Since we derive this field from `validateWebhook || mutateWebhook`, I don't think we gain much by adding another parameter to this function. It also masks what would cause `deployAdmissionController` to be `true`.

src/lib/assets/assets.ts
```diff
@@ -281,8 +282,10 @@ export class Assets {
         apiPath: this.apiPath,
         capabilities: this.capabilities,
       };
+      const deployAdmissionController =
+        isAdmission(this.capabilities) || norWatchOrAdmission(this.capabilities);
       await overridesFile(overrideData, helm.files.valuesYaml, this.imagePullSecrets, {
-        admission: isAdmission(this.capabilities) || norWatchOrAdmission(this.capabilities),
+        admission: deployAdmissionController,
```
samayer12: See https://github.com/defenseunicorns/pepr/pull/3292/changes#r3960426234

samayer12 (commented):
I'd recommend one more pass to ensure our documentation is consistent and then this looks good.

For example, in the [`README.md`](https://docs.pepr.dev/#module), we say:

> During the Pepr build process, each module produces a unique Kubernetes MutatingWebhookConfiguration and ValidatingWebhookConfiguration, along with a secret containing the transpiled and compressed TypeScript code. 

Since this PR accounts for an example where we don't use the standard Pepr bindings, we should mention that behavior.

src/lib/assets/assets.test.ts
```diff
@@ -299,7 +299,7 @@ describe("Assets", () => {
     );
   });
 
-  it("should call writeWebhookFiles and write admissionController Deployment, ServiceMonitor, and WebhookConfigs", async () => {
+  it("should call writeWebhookFiles and write WebhookConfigs", async () => {
```
samayer12, line 302: The change to this test doesn't alter the test body and just removes the number of expected calls. Should this be a separate test case with a different input? Or do we really not need to ensure that four calls to `writeFile` occur in all use-cases?

samayer12 (changes requested):

samayer12 (approved):

PR #3307, opened 2026-09-03
fix: consume KFC watcher types from package root

samayer12:
## Summary
- replace Pepr's KFC watcher deep imports with the supported package-root API
- upgrade the production dependency from `kubernetes-fluent-client@3.12.0` to `3.12.1`, the published release that provides those root exports
- rely on KFC's packed-artifact consumer test for public API coverage, avoiding a duplicate Pepr-level contract test

## Verification
- `npm exec vitest -- run src/lib/controller/storeCache.test.ts src/lib/core/capability.test.ts src/lib/core/queue.test.ts src/lib/deploymentChecks.test.ts src/lib/finalizer.test.ts src/lib/processors/watch-processor.test.ts`
- `npm exec tsc -- -p config/tsconfig.root.json`
- `npm run format:src`
- `npm run format:tests`
- `npm run build`


samayer12:
The smoke test fails because UDS Core uses an old version of pepr without these changes in place. We should merge this even though the smoke test fails and then create a PR for UDS Core to handle the import path management for them.

AmberFryar (approved):

PR #3326, opened 2026-09-09
ci: add Digestabot digest refresh workflow

samayer12:
## Summary
- Add a daily Digestabot workflow for mutable Chainguard image digests.
- Reuse the existing Chainguard OIDC identity and GitHub App token so generated PRs trigger CI.
- Scan Digestabot's default repository-wide supported file set.

## Validation
- Prettier check for the workflow file
- YAML structure parse
- git diff --check
- npm run check:container-image-equality

## Follow-up
- Manually dispatch from main to confirm the existing Chainguard identity is scoped for this repository, main branch, and registry pull access.

### DIFF
--- a/.github/workflows/digestabot.yml
+++ b/.github/workflows/digestabot.yml
@@ -1,0 +1,56 @@
+name: Update Chainguard Image Digests
+
+on:
+  schedule:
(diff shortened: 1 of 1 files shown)


samayer12:
Probably not needed since we configured dependabot (see #3327). If we have future issues, this PR can be a reference.

